PR #2, opened 2021-04-21
Initial Commit

eldadfux:


src/Database/Database.php
```diff
@@ -0,0 +1,739 @@
...
+    {
+        self::$filters[$name] = [
+            'encode' => $encode,
+            'decode' => $decode,
+        ];
+    }
+
+    public function encode(Document $collection, Document $document):Document
+    {
+        $rules = $collection->getAttribute('rules', []);
```
kodumbeats, line 620: Wondering if I've missed where the `rules` attribute is used. Should this be?:
```suggestion
        $rules = $collection->getAttribute('attributes', []);
```

kodumbeats:
A collection's attributes are stored as a JSON string, and I think I can get it like so:
`$rules = $collection->getAttribute('attributes', []);`

Should I be using the new `encode()`/`decode()` methods somehow to decode this JSON string, or can I run with `json_decode()`

src/Database/Database.php
eldadfux: Nope, the encode/decode is still not done, these are leftovers from the old code.

eldadfux:
> A collection's attributes are stored as a JSON string, and I think I can get it like so:
> `$rules = $collection->getAttribute('attributes', []);`
> 
> Should I be using the new `encode()`/`decode()` methods somehow to decode this JSON string, or can I run with `json_decode()`

This decoding process should be solved at a deeper level, when using the `getAttribute` method you'll get the data as you would expect to get it (structured).

README.md
```diff
@@ -1 +1,172 @@
...
+$database->createAttribute('documents', 'colors', Database::VAR_STRING, 32, true,  true);
+
+// Create an Index
+$database->createIndex('indexes', 'index1', Database::INDEX_KEY, ['string', 'integer'], [128], [Database::ORDER_ASC]);
+```
+
+**Create a document:**
+
+```php
+$document = static::getDatabase()->createDocument('documents', new Document([
```
kodumbeats, line 99:
```suggestion
$document = $database->createDocument('documents', new Document([
```

.phpunit.result.cache
```diff
@@ -0,0 +1 @@
...gresTest::testList";d:0.002;s:53:"Utopia\Tests\Adapter\MariaDBTest::testListCollections";d:0.034;s:53:"Utopia\Tests\Adapter\MongoDBTest::testListCollections";d:0.007;s:51:"Utopia\Tests\Adapter\MySQLTest::testListCollections";d:0.003;s:54:"Utopia\Tests\Adapter\PostgresTest::testListCollections";d:0.005;s:51:"Utopia\Tests\Adapter\MySQLTest::testCreateAttribute";d:0.406;s:59:"Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteAttribute";d:0.182;s:55:"Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteIndex";d:0.099;s:60:"Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteCollection";d:0.047;s:58:"Utopia\Tests\Adapter\MySQLTest::testCreateDeleteCollection";d:0.067;s:57:"Utopia\Tests\Adapter\MySQLTest::testCreateDeleteAttribute";d:0.874;s:53:"Utopia\Tests\Adapter\MySQLTest::testCreateDeleteIndex";d:0.242;s:48:"Utopia\Tests\Adapter\MySQLTest::testCreateDelete";d:1.016;s:50:"Utopia\Tests\Adapter\MySQLTest::testCreateDocument";d:0.126;s:44:"Utopia\Tests\DocumentTest::testGetAttributes";d:0.003;s:50:"Utopia\Tests\Adapter\MariaDBTest::testCreateDelete";d:0.256;s:52:"Utopia\Tests\Adapter\MariaDBTest::testCreateDocument";d:0.074;s:49:"Utopia\Tests\Adapter\MariaDBTest::testGetDocument";d:0.008;}}}
```
TorstenDittmann, line 1: This needs to be `git rm` so its picked up by `.gitignore`.

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+  && rm -rf /var/cache/apk/*
+
+WORKDIR /usr/src/code
+
+RUN echo extension=swoole.so >> /usr/local/etc/php/conf.d/swoole.ini
+
+COPY --from=step0 /usr/local/src/vendor /usr/src/code/vendor
+
+# Add Source Code
+COPY ./ /usr/src/code
```
TorstenDittmann, line 44: Maybe just `.` instead of `./` for better readability.

composer.json
```diff
@@ -0,0 +1,37 @@
...
+        "psr-4": {"Utopia\\Database\\": "src/Database"}
+    },
+    "autoload-dev": {
+        "psr-4": {"Utopia\\Tests\\": "tests/Database"}
+    },
+    "require": {
+        "php": ">=7.1",
+        "ext-pdo": "*",
+        "ext-redis": "*",
+        "ext-mongodb": "*",
```
TorstenDittmann, line 28: Shouldn't this be in the dev-dependencies?

Otherwise it is required in composer when used as a dependency right?

src/Database/Adapter.php
```diff
@@ -0,0 +1,301 @@
...
+    /**
+     * @var array
+     */
+    protected $debug = [];
+
+    /**
+     * @param string $key
+     * @param mixed $value
+     *
+     * @return $this
```
TorstenDittmann, line 23: Is `$this` valid?

I think we should use the actual class here.

src/Database/Adapter.php
```diff
@@ -0,0 +1,301 @@
...
+     */
+    public function setNamespace(string $namespace): bool
+    {
+        if (empty($namespace)) {
+            throw new Exception('Missing namespace');
+        }
+
+        $this->namespace = $namespace;
+
+        return true;
```
TorstenDittmann, line 69: Is there a reason for this returning `true`?

src/Database/Adapter.php
```diff
@@ -0,0 +1,301 @@
...
+
+    /**
+     * Filter Keys
+     * 
+     * @throws Exception
+     * @return string
+     */
+    public function filter(string $value):string
+    {
+        $value = preg_replace("/[^A-Za-z0-9]_/", '', $value);
```
TorstenDittmann, line 293: Can we avoid RegEx somehow?

The readability suffers alot and if there is a native method we could utilize, we might even get a performance boost (even though the impact is not worth mentioning).

src/Database/Adapter/MariaDB.php
```diff
@@ -0,0 +1,790 @@
...
+     * 
+     * @param string $id
+     * @return bool
+     */
+    public function deleteCollection(string $id): bool
+    {
+        $id = $this->filter($id);
+        
+        $this->getPDO()
+            ->prepare("DROP TABLE {$this->getNamespace()}.{$id}_permissions;")
```
TorstenDittmann, line 140: Can we use `bindParam` or `bindValue`for security?

Maybe we should consider it using at every place where we use variables in strings?

src/Database/Database.php
```diff
@@ -0,0 +1,1092 @@
...
+    /**
+     * @param Adapter $adapter
+     * @param Cache $cache
+     */
+    public function __construct(Adapter $adapter, Cache $cache)
+    {
+        $this->adapter = $adapter;
+        $this->cache = $cache;
+
+        self::addFilter('json',
```
TorstenDittmann, line 142: Will `encrypt` and `decrypt` also be part of this library?

composer.json
eldadfux: We actually have to use them in prod not only dev, otherwise the adapters won't work and we have no way to tell which one a user is going to use 🤔

src/Database/Adapter.php
eldadfux: As far as I know it is valid and working, I think you can use both `self` and `$this`, easier to avoid the class name in case we want to change it without working hard.

src/Database/Adapter.php
eldadfux: Just to have a consistent interface across the class.

src/Database/Adapter.php
eldadfux: Yes, I like avoiding regex, let me see if I can come up with something effective.

src/Database/Adapter/MariaDB.php
eldadfux: No. Both work only for prepared statements and not for table or column names, this makes the filter method very important.

src/Database/Database.php
eldadfux: The will be set as part of Appwrite to avoid too many dependencies (all the openssl stuff)

tests/Database/Base.php
```diff
@@ -0,0 +1,949 @@
...
+            'active' => false,
+            'generes' => [],
+        ]));
+
+        /**
+         * Check Basic
+         */
+        $documents = static::getDatabase()->find('movies');
+
+        $this->assertEquals(5, count($documents));
```
kodumbeats, line 394: I think I found a bug while implementing the MongoDB adapter.

**six** movies are added to the `movies` collection here for testing, but only five are found.
Here's the `var_dump($documents)`: https://gist.github.com/kodumbeats/696b41d560dea5a0794378c1a77878be 

`Work in Progress 2` is not found in `static::getDatabase()->find('movies');`

kodumbeats: Answered my own question - the first assertion returns all documents with wildcard read permissions.

src/Database/Adapter/MariaDB.php
```diff
@@ -0,0 +1,825 @@
...
+    /**
+     * Delete Attribute
+     * 
+     * @param string $collection
+     * @param string $id
+     * @param bool $array
+     * 
+     * @return bool
+     */
+    public function deleteAttribute(string $collection, string $id, bool $array = false): bool
```
lohanidamodar, line 184: $array is not used for deleteAttribute, do we need this param?

src/Database/Adapter/MariaDB.php
```diff
@@ -0,0 +1,825 @@
...
+     * @param array $orderTypes
+     * @param bool $count
+     *
+     * @return Document[]
+     */
+    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): array
+    {
+        $name = $this->filter($collection);
+        $roles = Authorization::getRoles();
+        $where = ['1=1'];
```
lohanidamodar, line 491: What does this do?

src/Database/Adapter/MariaDB.php
```diff
@@ -0,0 +1,825 @@
...
+        }
+
+        $stmt->bindValue(':limit', $limit, PDO::PARAM_INT);
+        $stmt->bindValue(':offset', $offset, PDO::PARAM_INT);
+        $stmt->execute();
+
+        $results = $stmt->fetchAll();
+
+        foreach ($results as &$value) {
+            $permissions = (isset($value['_permissions'])) ? json_decode($value['_permissions'], true) : [];
```
lohanidamodar, line 540: `isset($value['permissions'])` if this returns false, means, the permission for that document doesn't exist right?, if the permission doesn't exist aren't we suppose to skip the document and not add it in the final $results?

lohanidamodar (commented):
Hey just few questions.

src/Database/Database.php
```diff
@@ -0,0 +1,1111 @@
...
+    
+    // Index Types
+    const INDEX_KEY = 'text';
+    const INDEX_FULLTEXT = 'fulltext';
+    const INDEX_UNIQUE = 'unique';
+    const INDEX_SPATIAL = 'spatial';
+
+    // Orders
+    const ORDER_ASC = 'ASC';
+    const ORDER_DESC = 'DESC';
```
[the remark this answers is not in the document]
eldadfux: Check this answer: https://stackoverflow.com/a/6920679/2299554

src/Database/Adapter.php
```diff
@@ -0,0 +1,301 @@
...
+     * @param string $collection
+     * @param \Utopia\Database\Query[] $queries
+     * @param int $limit
+     * @param int $offset
+     * @param array $orderAttributes
+     * @param array $orderTypes
+     *
+     * @return Document[]
+     */
+    abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): array;
```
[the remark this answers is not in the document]
eldadfux: Not sure what you mean by `casting`, are you referring to current behavior where we need to indicate the casting type? If yes, this is not required anymore because we store this metadata for each attribute.

src/Database/Adapter/MariaDB.php
```diff
@@ -0,0 +1,825 @@
...
+
+    /**
+     * List Databases
+     * 
+     * @return array
+     */
+    public function list(): array
+    {
+        $list = [];
+        return $list;
```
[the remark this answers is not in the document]
eldadfux: Yes, and not very important at this stage.

src/Database/Adapter/MariaDB.php
eldadfux: We might need it to support `in`\`contains` operators

src/Database/Adapter/MariaDB.php
eldadfux: Helps us avoid from an extra `if` to check if we have any where clauses, just a hacky shortcut

src/Database/Adapter/MariaDB.php
eldadfux: The read permission validation is performed in the query level, if you're using an API key or you're a console user you'll get all docs no matter what permission it has.

src/Database/Database.php
```diff
@@ -0,0 +1,1111 @@
...
+            }
+        }
+
+        $collection->setAttribute('attributes', $attributes);
+    
+        if($collection->getId() !== self::COLLECTIONS) {
+            $this->updateDocument(self::COLLECTIONS, $collection->getId(), $collection);
+        }
+
+        return $this->adapter->deleteAttribute($collection->getId(), $id);
```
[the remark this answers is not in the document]
eldadfux: OK, so every time we update/change a table we update the collection metadata on line 404, We skip updating only the root collection because it's stored on the codebase and not in the adapter.

Here it is:
https://github.com/utopia-php/database/blob/v0/src/Database/Database.php#L72-L126

src/Database/Adapter/MariaDB.php
lohanidamodar: Oh thanks, this is awesome 😄

src/Database/Adapter/MariaDB.php
lohanidamodar: Even then, while deleting attribute do we need array check? (in\contains) is for query and the `array` is the attributes value right?

src/Database/Database.php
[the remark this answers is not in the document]
lohanidamodar: May be we can support it in high level in Appwrite API as this suggests
https://stackoverflow.com/questions/6541644/selecting-random-rows-with-mysql/6542113#6542113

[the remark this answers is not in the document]
eldadfux: `COUNT(*)` can be very very expensive depends on your collection/table size

src/Database/Adapter/MariaDB.php
eldadfux: Yes, because if we're going to have a secondary table for arrays we'll need this flag to know when to delete it.

src/Database/Database.php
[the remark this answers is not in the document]
lohanidamodar: Understood!

src/Database/Database.php
```diff
@@ -0,0 +1,1111 @@
...
+                'size' => 1000000,
+                'required' => false,
+                'signed' => true,
+                'array' => false,
+                'filters' => ['json'],
+            ],
+            [
+                '$id' => 'attributesInQueue',
+                'type' => self::VAR_STRING,
+                'size' => 1000000,
```
TorstenDittmann, line 107: Will the maximum the size defined impact the query times or will it consider only the actual values?

TorstenDittmann (approved):

src/Database/Database.php
eldadfux: Right now we just use it to distinguish between:

```php
            case Database::VAR_STRING:
                // $size = $size * 4; // Convert utf8mb4 size to bytes
                if($size > 16777215) {
                    return 'LONGTEXT';
                }
                
                if($size > 65535) {
                    return 'MEDIUMTEXT';
                }

                if($size > 16383) {
                    return 'TEXT';
                }
                
                return "VARCHAR({$size})";
            break;
```

lohanidamodar (approved):

src/Database/Database.php
TorstenDittmann: Ah okay!

Even better 👍🏻  Maybe we can add constants for this at some point, so its self-explained 🙂

PR #3, opened 2021-04-22
Feat - add custom filter parser

kodumbeats:
## New Feature
This PR introduces a new Query class which will hold all related information for custom document filters.

Note - due to `$filter` being used already, I refer to a custom filter as a `$query` until we can find something more clear. 

## Test plan
Tests included with PR

## Todo

- [x] Implement filter parser with no nesting
- [x] Create new Query class
- [x] Add tests
- [x] Unit test `parseExpression()`

## Should we...

- [ ] Check types against collection schema
- [ ] Expand tests to check all query types


kodumbeats:
Testing locally against MariaDB has 🟢 
```bash
$ docker-compose exec tests vendor/bin/phpunit --debug /usr/src/code/tests/Database/Adapter/MariaDBTest.php

PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDelete' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDelete' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteCollection' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteCollection' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteAttribute' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteAttribute' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteIndex' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDeleteIndex' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDocument' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testCreateDocument' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testGetDocument' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testGetDocument' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testUpdateDocument' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testUpdateDocument' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testDeleteDocument' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testDeleteDocument' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testFindFirst' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testFindFirst' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testFindLast' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testFindLast' ended
Test 'Utopia\Tests\Adapter\MariaDBTest::testParseQueries' started
Test 'Utopia\Tests\Adapter\MariaDBTest::testParseQueries' ended


Time: 00:03.764, Memory: 6.00 MB

OK (11 tests, 101 assertions)
```

eldadfux:
Thinking about this a bit here are some of my ideas:
- We will create a class called `Database\Query` to hold all query related operations/data
- The class constructor will create a query instance with all relevant metadata that will allow us to build a native SQL/NOSQL query (key, operator, value/values), an array of query objects will allow us to build complex queries.
- Create a static method called `Database\Query::parse([STRING])` that will parse a string query (`director.equals('brandon')`) and return a `Database\Query` instance.
- The `Database\Query` class can have multiple getters and setters to make DX nicer.

kodumbeats:
As of e802153
```bash
# kodumbeats @ magician in ~/w/kodumbeats/database on git:v0 x [12:28:00]
$ dc exec tests vendor/bin/phpunit tests/Database/Adapter/MariaDBTest.php 
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

..........                                                        10 / 10 (100%)

Time: 00:00.711, Memory: 6.00 MB

OK (10 tests, 85 assertions)

# kodumbeats @ magician in ~/w/kodumbeats/database on git:v0 x [12:28:17] 
$ dc exec tests vendor/bin/phpunit tests/Database/QueryTest.php          
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

..                                                                  2 / 2 (100%)

Time: 00:00.002, Memory: 6.00 MB

OK (2 tests, 10 assertions)

```

src/Database/Query.php
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Construct.
+     *
+     * Construct a new query object
+     *
+     * @param string $attribute
+     * @param string $operator
+     * @param mixed $operand
+     */
+    public function __construct($attribute, $operator, $operand)
```
eldadfux, line 31: Maybe $value instead of $operand might be a better fit?

eldadfux: Missing types for $attribute and $operator

eldadfux: Can we assume $operand / $value are always arrays? this will make support for `or` queries simpler

src/Database/Query.php
```diff
@@ -0,0 +1,154 @@
...
+        $stanzas = mb_substr_count($filter, ".") + 1;
+
+        // TODO@kodumbeats handle relations between collections, e.g. if($stanzas > 2)
+        switch ($stanzas):
+            case 2:
+                $input = explode('.', $filter);
+                $attribute = $input[0];
+                [$operator, $operand] = Query::parseExpression($input[1]);
+                break;
+        endswitch;
```
eldadfux, line 101: In PHP this `endswitch` syntax is more common for PHTML templates and not standard code, you can use `{}` instead.

src/Database/Query.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Get attribute key-value from query expression
+     * $expression: string with format 'operator(operand)'
+     *
+     * @param string $expression
+     *
+     * @return (string|mixed)[]
+     */
+    public static function parseExpression(string $expression): array
```
eldadfux, line 114: Might make more sense to have this as a protected static method?

src/Database/Query.php
```diff
@@ -0,0 +1,154 @@
...
+        );
+
+        //strip quotes from queries of type string
+        $operand = str_replace('"', "", $operand);
+        $operand = str_replace("'", "", $operand);
+
+        return [$operator, $operand];
+    }
+
+    // /**
```
eldadfux, line 143: Love this idea! we should probably have a validator class for queries syntax...

eldadfux (changes requested):

src/Database/Query.php
kodumbeats: I tried the same at first but hit this error:
```
2) Utopia\Tests\QueryTest::testParseExpression
Error: Call to protected method Utopia\Database\Query::parseExpression() from context 'Utopia\Tests\QueryTest'

/usr/src/code/tests/Database/QueryTest.php:35
```

src/Database/Query.php
kodumbeats: > Can we assume $operand / $value are always arrays? this will make support for `or` queries simpler

Sure! I've refactored it to `$expression: [$operator => $value]`

src/Database/Query.php
eldadfux: You can't test protected methods... which is completely fine if the other tests cover that protected logic...

kodumbeats:
As of 8c8afb3
```bash
# kodumbeats @ magician in ~/w/kodumbeats/database on git:v0 x [15:22:50] 
$ dc exec tests vendor/bin/phpunit tests/Database/Adapter/MariaDBTest.php  
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

................                                                  16 / 16 (100%)

Time: 00:00.721, Memory: 6.00 MB

OK (16 tests, 94 assertions)

# kodumbeats @ magician in ~/w/kodumbeats/database on git:v0 x [15:22:59] 
$ dc exec tests vendor/bin/phpunit tests/Database/QueryTest.php          
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

..                                                                  2 / 2 (100%)

Time: 00:00.002, Memory: 6.00 MB

OK (2 tests, 10 assertions)
```

src/Database/Query.php
kodumbeats: This is my problem then: I call one function in the other:
https://github.com/utopia-php/database/pull/3/files#diff-35b0686b57f21084e677c6f36a52de692612fb03fff51a477b43c6bd3ac7993cR99

Should I just refactor the logic of `parseExpression()` into `parse()`?

src/Database/Query.php
kodumbeats: > > Can we assume $operand / $value are always arrays? this will make support for `or` queries simpler
> 
> Sure! I've refactored it to `$expression: [$operator => $value]`

Didn't have much luck with this approach - curious to know where how you'd like to expose the $expression array:
https://github.com/utopia-php/database/pull/3/commits/1c6e60c20f7afcd8639a55b8fb56c4acfb0cdf7d

src/Database/Query.php
kodumbeats: How's this for a start? I figure this is a good place to check if the operator is supported too:
https://github.com/utopia-php/database/pull/3/commits/179a0d51da3db2313a9c7662ebcc593b68d0c1f0

src/Database/Query.php
eldadfux: use `self::` not `Query::`

src/Database/Query.php
eldadfux: Looks like a good start, we can have it as a standalone PR

src/Database/Query.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        // TODO@kodumbeats handle '.' in expression value 
+        $stanzas = mb_substr_count($filter, ".") + 1;
+
+        // TODO@kodumbeats handle relations between collections, e.g. if($stanzas > 2)
+        switch ($stanzas) {
+            case 2:
+                $input = explode('.', $filter);
+                $attribute = $input[0];
+                [$operator, $value] = Query::parseExpression($input[1]);
```
eldadfux, line 99: Not sure what this change is doing

src/Database/Query.php
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Construct.
+     *
+     * Construct a new query object
+     *
+     * @param string $attribute
+     * @param string $operator
+     * @param mixed $value
+     */
+    public function __construct(string $attribute, string $operator, $value)
```
eldadfux, line 31: My idea for supporting `or` queries was set $value as array and maybe call it `$values`

eldadfux (changes requested):

src/Database/Query.php
kodumbeats: Clarified in 7071923

src/Database/Query.php
kodumbeats: Introduced in e3f1efb, though it means we have to do more work with appropriately typecasting here.

src/Database/Query.php
kodumbeats: Even with `self::`, `phpunit` throws the same error. Since we construct via a static method instead of the constructor, we have additional constraints.
Commit: 7071923

kodumbeats: My misunderstanding! Addressed 👍🏻

eldadfux:
We also need to support multiple params for `or` queries, a few examples:

`director.equals('eldad', 'brandon')`
`director.equals('eldad','brandon')`

kodumbeats:
As of 631ecfd
```
$ docker-compose exec tests vendor/bin/phpunit --debug tests/Database/QueryTest.php
PHPUnit 9.5.4 by Sebastian Bergmann and contributors.

Test 'Utopia\Tests\QueryTest::testParse' started
Test 'Utopia\Tests\QueryTest::testParse' ended
Test 'Utopia\Tests\QueryTest::testGetAttribute' started
Test 'Utopia\Tests\QueryTest::testGetAttribute' ended
Test 'Utopia\Tests\QueryTest::testGetOperator' started
Test 'Utopia\Tests\QueryTest::testGetOperator' ended
Test 'Utopia\Tests\QueryTest::testGetValue' started
Test 'Utopia\Tests\QueryTest::testGetValue' ended
Test 'Utopia\Tests\QueryTest::testGetQuery' started
Test 'Utopia\Tests\QueryTest::testGetQuery' ended


Time: 00:00.002, Memory: 6.00 MB

OK (5 tests, 22 assertions)
```

eldadfux:
Looks good! I would have added more tests to make sure casting works as expected with all types

src/Database/Query.php
```diff
@@ -0,0 +1,168 @@
...
+            } elseif ($value === 'false') {
+                return false;
+            } elseif ($value === '"null"') {
+                // need special case to handle null with or without quotes
+                return 'null';
+            } elseif ($value === 'null') {
+                return null;
+            } else {
+                // strip quotes from queries of type string
+                return str_replace(['"',"'"], "", $value);
```
eldadfux, line 162: We need to remove only the first and last chars if they are `"` or `'`

src/Database/Query.php
```diff
@@ -0,0 +1,168 @@
...
+
+        /** @var mixed */
+        $value = mb_substr($expression, 
+            ($start + 1), /* exclude open paren*/ 
+            ($end - $start - 1) /* exclude closed paren*/
+        );
+
+        // Explode comma-separated values and trim whitespace
+
+        $values = array_map('trim', explode(',', $value));
```
eldadfux, line 142: Instead of using array map twice, we can have this trim as part of casting run - it will be more effective

src/Database/Query.php
eldadfux: Maybe we should also handle escaping of `\'` and `\"`

eldadfux (changes requested):

src/Database/Query.php
```diff
@@ -0,0 +1,168 @@
...
+                return $value + 0;
+            } elseif ($value === 'true') {
+                // since (bool)"false" returns true, check bools manually
+                return true;
+            } elseif ($value === 'false') {
+                return false;
+            } elseif ($value === '"null"') {
+                // need special case to handle null with or without quotes
+                return 'null';
+            } elseif ($value === 'null') {
```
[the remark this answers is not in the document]
kodumbeats: Removed one of these 👍🏻

src/Database/Query.php
kodumbeats: Used `trim()` and `stripslashes()` for this

src/Database/Query.php
kodumbeats: Agreed! 👍🏻

src/Database/Query.php
```diff
@@ -0,0 +1,175 @@
...
+                    return $value + 0;
+
+                // since (bool)"false" returns true, check bools manually
+                case $value === 'true':
+                    return true;
+
+                case $value === 'false':
+                    return false;
+
+                // need special case to handle null with or without quotes
```
TorstenDittmann, line 160: Is there even a need to handle a the case with a quoted `null`? 

imo it should be handled like a string then 🙂

src/Database/Query.php
```diff
@@ -0,0 +1,175 @@
...
+     * @return (string|array)[]
+     */
+    protected static function parseExpression(string $expression): array
+    {
+        //find location of parentheses in expression
+
+        /** @var int */
+        $start = mb_strpos($expression, '(');
+        /** @var int */
+        $end = mb_strpos($expression, ')');
```
TorstenDittmann, line 125: Maybe consider using `mb_strrpos()` - notice the second `r`.
This way we ignore `)` as a part of the value.
https://www.php.net/manual/en/function.mb-strrpos.php

tests/Database/QueryTest.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $this->assertEquals('published', $query->getAttribute());
+        $this->assertEquals('equal', $query->getOperator());
+        $this->assertContains(false, $query->getValues());
+
+        $query = Query::parse('actors.notContains("      Johnny Depp  ")');
+
+        $this->assertEquals('actors', $query->getAttribute());
+        $this->assertEquals('notContains', $query->getOperator());
+        $this->assertContains('Johnny Depp', $query->getValues());
```
eldadfux, line 58: I would expect ` Johnny Depp ` here. (with all the spaces)

eldadfux: we should only trim spaces around the quotes not within.

eldadfux (changes requested):

PR #7, opened 2021-04-30
Feature - Strict Query Validation

kodumbeats:
This PR introduces a new Queries validator to validate an array of Query objects by:
- Individually validating each Query object
- **Strict validation**: validate the query params can all be found in one index
- Checks if index is queued for creation during strict validation

Test plan:
Unit tests included

TODO:

- [ ] Handle index order (ASC, DESC, etc)

### DIFF
--- a/src/Database/Validator/Queries.php
+++ b/src/Database/Validator/Queries.php
@@ -66,26 +66,36 @@ class Queries extends Validator
     public function isValid($value)
     {
         /**
-         * Array of attributes from $value
+         * Array of attributes from Query->getAttribute()
          *
          * @var string[]
          */
-        $found = [];
+        $queries = [];
 
         foreach ($value as $query) {
-            // individually validate $query
+            $queries[] = $query->getAttribute(); 
+
             if (!$this->validator->isValid($query)) {
                 $this->message = 'Query not valid: ' . $this->validator->getDescription();
                 return false;
             }
+        }
 
-            // $attribute = array_search($query->getAttribute(), $indexes);
-            // TODO@kodumbeats add logic to check all attributes can be found in a single index
+        /**
+         * @var string
+         */
+        $indexId = null;
+        
+        // look for strict match among indexes
+        foreach ($this->indexes as $index) {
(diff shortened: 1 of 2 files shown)


eldadfux:
@kodumbeats I might also add a new collection attributed called "indexesInQueue" to hold pending indexes that are still being created, this will allow us to give a better error messages in this validator.

kodumbeats:
@eldadfux I'm not 💯 on what is best to pass to the Queries validator constructor. I landed on this approach in acefb92 because passing in `Utopia\Database\Database` made the code really messy, and I lost the story.

src/Database/Validator/Queries.php
```diff
@@ -0,0 +1,163 @@
...
+        /**
+         * @var bool
+         */
+        $queued = false;
+        
+        // Return false if attributes do not exactly match an index
+        if ($this->strict) {
+            // look for strict match among indexes
+            foreach ($this->indexes as $index) {
+                if ($index['attributes'] === $queries) {
```
eldadfux, line 103: I'm not sure this array comparison in PHP actually work

src/Database/Validator/Queries.php
```diff
@@ -0,0 +1,163 @@
...
+                    $indexId = $index['$id']; 
+                }
+            }
+
+            if (!$indexId) {
+                // check against the indexesInQueue
+                foreach ($this->indexesInQueue as $index) {
+                    if ($index['attributes'] === $queries) {
+                        $queued = true;
```
eldadfux, line 112: If an index was found in the queue, we can stop the iteration by returning false. It will be faster.

eldadfux (changes requested):
Looks very good, left some minor comments. When ready I can merge it and start playing around with it :)

src/Database/Validator/Queries.php
kodumbeats: Moved array comparison logic to its own method. 👍🏻

eldadfux (approved):

PR #8, opened 2021-05-06
Feat: MongoDB Adapter

kodumbeats:
This PR adds a MongoDB database adapter

### DIFF
--- a/src/Database/Adapter/MongoDB.php
+++ b/src/Database/Adapter/MongoDB.php
@@ -273,7 +273,31 @@ class MongoDB extends Adapter
      * @return Document
      */
     public function updateDocument(string $collection, Document $document): Document
-    {}
+    {
+        $name = $this->filter($collection);
+        $collection = $this->getDatabase()->$name;
+
+        $result = $collection->findOneAndUpdate(
+            ['+id' => $document->getId()],
+            ['$set' => $this->replaceChars('$', '+', $document->getArrayCopy())],
+            [
+                'typemap' => [
+                    'root' => 'array',
+                    'document' => 'array',
+                    'array' => 'array'
+                ],
+                'returnDocument' => \MongoDB\Operation\FindOneAndUpdate::RETURN_DOCUMENT_AFTER,
+            ]
+        );
+
+        // Remove internal Mongo ID
+        unset($result['_id']);
+
+        // Change back to $
+        $result = $this->replaceChars('+', '$', $result);
+
+        return new Document($result);
+    }
 
     /**
      * Delete Document
--- a/src/Database/Adapter/MongoDB.php
+++ b/src/Database/Adapter/MongoDB.php
@@ -1,134 +1,185 @@
 <?php
 
-// namespace Utopia\Database\Adapter;
+namespace Utopia\Database\Adapter;
 
-// use Exception;
-// use Utopia\Database\Adapter;
-// use MongoDB\Client;
-// use MongoDB\Database;
+use Exception;
+use Utopia\Database\Adapter;
+use MongoDB\Client;
+use MongoDB\Database;
 
-// class MongoDB extends Adapter
-// {
-//     /**
-//      * @var Client
-//      */
-//     protected $client;
+class MongoDB extends Adapter
+{
+    /**
+     * @var Client
+     */
+    protected $client;
 
-//     /**
-//      * @var Database|null
-//      */
-//     protected $database;
+    /**
+     * @var Database|null
+     */
+    protected $database;
 
-//     /**
-//      * Constructor.
-//      *
-//      * Set connection and settings
-//      *
-//      * @param Client $client
-//      */
-//     public function __construct(Client $client)
-//     {
-//         $this->client = $client;
-//     }
+    /**
+     * Constructor.
+     *
+     * Set connection and settings
+     *
+     * @param Client $client
+     */
+    public function __construct(Client $client)
+    {
+        $this->client = $client;
+    }
 
-//     /**
-//      * Create Database
-//      * 
-//      * @return bool
-//      */
-//     public function create(): bool
-//     {
-//         $namespace = $this->getNamespace();
-//         return (!!$this->client->$namespace);
-//     }
+    /**
+     * Create Database
+     * 
+     * @return bool
+     */
+    public function create(): bool
+    {
+        $namespace = $this->getNamespace();
+        return (!!$this->client->$namespace);
+    }
 
-//     /**
-//      * List Databases
-//      * 
-//      * @return array
-//      */
-//     public function list(): array
-//     {
-//         $list = [];
+    /**
+     * List Databases
+     * 
+     * @return array
+     */
+    public function list(): array
+    {
+        $list = [];
 
-//         foreach ($this->client->listDatabaseNames() as $key => $value) {
-//             $list[] = $value;
-//         }
+        foreach ($this->client->listDatabaseNames() as $key => $value) {
+            $list[] = $value;
+        }
         
-//         return $list;
-//     }
+        return $list;
+    }
 
-//     /**
-//      * Delete Database
-//      * 
-//      * @return bool
-//      */
-//     public function delete(): bool
-//     {
-//         return (!!$this->getDatabase()->dropCollection($this->getNamespace()));
-//     }
+    /**
+     * Delete Database
+     * 
+     * @return bool
+     */
+    public function delete(): bool
+    {
+        return (!!$this->getDatabase()->drop($this->getNamespace()));
+    }
 
-//     /**
-//      * Create Collection
-//      * 
-//      * @param string $id
-//      * @return bool
-//      */
-//     public function createCollection(string $id): bool
-//     {
-//         return (!!$this->getDatabase()->createCollection($id));
-//     }
+    /**
+     * Create Collection
+     * 
+     * @param string $id
+     * @return bool
+     */
+    public function createCollection(string $id): bool
+    {
+        return (!!$this->getDatabase()->createCollection($id));
+    }
 
-//     /**
-//      * List Collections
-//      * 
-//      * @return array
-//      */
-//     public function listCollections(): array
-//     {
-//         $list = [];
+    /**
+     * List Collections
+     * 
+     * @return array
+     */
+    public function listCollections(): array
+    {
+        $list = [];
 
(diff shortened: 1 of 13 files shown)


kodumbeats:
@eldadfux  I could use your advice going forward:

When fetching documents, Mongo client SDK returns everything in a `MongoDB\Model\BSONDocument` object:
https://docs.mongodb.com/php-library/v1.2/reference/bson/#phpclass.MongoDB\Model\BSONDocument

What'd be the best approach to parse these objects and get back the PHP array? I've found I can go one level deep with `iterator_to_array`, but the array contains BSON models as well (`MongoDB\Model\BSONArray` and `MongoDB\BSON\ObjectId`) - looks like this:

```php
array(10) {
  ["_id"]=>
  object(MongoDB\BSON\ObjectId)#167 (1) {
    ["oid"]=>
    string(24) "60945e59acf8b169f83ce597"
  }
  ["+id"]=>
  string(9) "documents"
  ["+read"]=>
  object(MongoDB\Model\BSONArray)#164 (1) {
    ["storage":"ArrayObject":private]=>
    array(1) {
      [0]=>
      string(1) "*"
    }
  }
  ["+write"]=>
  object(MongoDB\Model\BSONArray)#169 (1) {
    ["storage":"ArrayObject":private]=>
    array(1) {
      [0]=>
      string(1) "*"
    }
  }
  ["name"]=>
  string(9) "documents"
  ["attributes"]=>
  string(2) "[]"
  ["indexes"]=>
  string(2) "[]"
  ["attributesInQueue"]=>
  string(2) "[]"
  ["indexesInQueue"]=>
  string(2) "[]"
  ["+collection"]=>
  string(11) "collections"
}

```

TorstenDittmann:
> @eldadfux  I could use your advice going forward:
> 
> When fetching documents, Mongo client SDK returns everything in a `MongoDB\Model\BSONDocument` object:
> https://docs.mongodb.com/php-library/v1.2/reference/bson/#phpclass.MongoDB\Model\BSONDocument
> 
> What'd be the best approach to parse these objects and get back the PHP array? I've found I can go one level deep with `iterator_to_array`, but the array contains BSON models as well (`MongoDB\Model\BSONArray` and `MongoDB\BSON\ObjectId`) - looks like this:
> 
> ```php
> array(10) {
>   ["_id"]=>
>   object(MongoDB\BSON\ObjectId)#167 (1) {
>     ["oid"]=>
>     string(24) "60945e59acf8b169f83ce597"
>   }
>   ["+id"]=>
>   string(9) "documents"
>   ["+read"]=>
>   object(MongoDB\Model\BSONArray)#164 (1) {
>     ["storage":"ArrayObject":private]=>
>     array(1) {
>       [0]=>
>       string(1) "*"
>     }
>   }
>   ["+write"]=>
>   object(MongoDB\Model\BSONArray)#169 (1) {
>     ["storage":"ArrayObject":private]=>
>     array(1) {
>       [0]=>
>       string(1) "*"
>     }
>   }
>   ["name"]=>
>   string(9) "documents"
>   ["attributes"]=>
>   string(2) "[]"
>   ["indexes"]=>
>   string(2) "[]"
>   ["attributesInQueue"]=>
>   string(2) "[]"
>   ["indexesInQueue"]=>
>   string(2) "[]"
>   ["+collection"]=>
>   string(11) "collections"
> }
> 
> ```

https://stackoverflow.com/a/61972048

Did you try this?

src/Database/Database.php
```diff
@@ -1014,40 +1022,42 @@ public function decode(Document $collection, Document $document):Document
      */
     public function casting(Document $collection, Document $document):Document
     {
```
eldadfux, line 1024: A more readable way might be:

```php
if ($this->adapter->getSupportForCasting()) {
    return $document;
}
```

eldadfux (changes requested):

src/Database/Adapter/MongoDB.php
```diff
@@ -1,134 +1,592 @@
...
+            if (is_array($value) && $operator = '$eq') {
+                $filters[$attribute]['$in'] = $value;
+            } else {
+                $filters[$attribute][$operator] = $value;
+            }
+        }
+
+        // permissions
+        if (Authorization::$status) { // skip if authorization is disabled
+            $filters['+read']['$in'] = Authorization::getRoles();
```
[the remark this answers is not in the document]
kodumbeats: Mongo creates a unique index on `_id` when creating a collection, so we have that by default. For `$read` and `$write`, should we create these indexes at the database level when creating a collection?
For example, [`Utopia\Database\Database->createCollection()`](https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L259):

```php
if ($this->adapter->getSupportForIndex) {
    $this->adapter->createIndex(...);
}
```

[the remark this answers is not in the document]
eldadfux: I would keep them in the adapter for now, just because nosql vs sql behavior is very different for the way we handle permissions.

src/Database/Adapter/MongoDB.php
```diff
@@ -1,134 +1,613 @@
...
+
+        /**
+         * @var Document[]
+         */
+        $found = [];
+
+        $results = $collection->find($filters, $options);
+
+        foreach($results as $i => $result) {
+            $found[] = new Document($this->replaceChars('_', '$', $result));
```
eldadfux, line 381: I feel like `replaceChars` is very expensive. We can probably just replace `$id`, `$read`, and `$write`.

src/Database/Adapter/MongoDB.php
```diff
@@ -1,134 +1,613 @@
...
+     */
+    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): array
+    {
+        $name = $this->filter($collection);
+        $collection = $this->getDatabase()->$name;
+        $roles = Authorization::getRoles();
+
+        $filters = [];
+
+        $options = ['sort', 'limit' => $limit, 'skip' => $offset];
```
eldadfux, line 356:
```suggestion
        $options = ['sort' => [], 'limit' => $limit, 'skip' => $offset];
```

src/Database/Adapter/MongoDB.php
```diff
@@ -1,134 +1,613 @@
...
+        $name = $this->filter($collection);
+        $collection = $this->getDatabase()->$name;
+        $roles = Authorization::getRoles();
+
+        $filters = [];
+
+        $options = [];
+
+        // set max limit
+        $options['limit'] = $max;
```
eldadfux, line 407: We need to make sure this limits the number of documents the DB scans and not just the number of results returned (which is always 1 anyway).

src/Database/Adapter/MongoDB.php
```diff
@@ -1,134 +1,613 @@
...
+     * Is fulltext index supported?
+     * 
+     * @return bool
+     */
+    public function getSupportForFulltextIndex(): bool
+    {
+        return true;
+    }
+
+    public function getSupportForCasting(): bool
```
eldadfux, line 609:
```suggestion
    /**
     * Is casting supported?
     * 
     * @return bool
     */
    public function getSupportForCasting(): bool
```

src/Database/Adapter/MongoDB.php
```diff
@@ -1,134 +1,613 @@
...
+        $id = $this->filter($id);
+        $collection = $this->getDatabase()->$name;
+
+        $indexes = [];
+
+        foreach($attributes as $i => $attribute) {
+            $attribute = $this->filter($attribute);
+
+            switch ($type) {
+                case Database::INDEX_KEY:
```
eldadfux, line 200: Missing support for a unique index? don't we have a test for this?

eldadfux (changes requested):

src/Database/Adapter/MongoDB.php
kodumbeats: Happy to add unique indexes and tests, we don't have any right now:

https://github.com/utopia-php/database/blob/main/tests/Database/Base.php#L151-L177

src/Database/Adapter/MongoDB.php
kodumbeats: I'm not sure on this one:
According to the [mongo shell docs](https://docs.mongodb.com/manual/reference/method/db.collection.count/):

Field | Type | Description
-- | -- |  --
limit | integer | Optional. The maximum number of documents to count.

But for the [PHP lib](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-countDocuments/) implementation of the same method:
Field | Type | Description
-- | -- |  --
limit | integer | Optional. The maximum number of matching documents to return.

I think all implementations just limit the number of documents passed to the next step in the pipeline:
https://docs.mongodb.com/manual/reference/operator/aggregation/limit/

eldadfux: Maybe we can raise an issue on the Mongo SDK repo and ask. I feel like this is just a doc mistake

src/Database/Adapter/MongoDB.php
eldadfux: 👌🏻

src/Database/Adapter/MongoDB.php
kodumbeats: Had to replace `$collection` as well:  ebf9c4c

src/Database/Adapter/MongoDB.php
kodumbeats: bbf2f3f

src/Database/Adapter/MongoDB.php
kodumbeats: Yep, `limit` was the wrong setting. After more research, I have good news and bad news 😐 

Good news: I found the exact setting we need for our desired `count()` and `find()` methods: `maxScan`:
> maxScan Optional. Maximum number of documents or index keys to scan when executing the query.

Bad news: it was deprecated in Mongo 4.0 in favor of time-based limits:
> MongoDB deprecates the option maxScan for the find command and the mongo shell helper cursor.maxScan(). Instead, use maxTimeMS option or the helper cursor.maxTimeMS().

https://docs.mongodb.com/manual/release-notes/4.0-compatibility/#maxscan-option

Since the behavior was specifically deprecated by the Mongo team, I'm not sure we'll be able to hack our way around it.

PR #11, opened 2021-05-15
Feat new doc methods

eldadfux:
Added new methods to easily manipulate the Document object.

### DIFF
--- a/src/Database/Document.php
+++ b/src/Database/Document.php
@@ -177,45 +177,99 @@ class Document extends ArrayObject
     }
 
     /**
-     * Search.
+     * Find.
+     *
+     * @param string $key
+     * @param mixed $find
+     * @param string $subject
+     *
+     * @return mixed
+     */
+    public function find(string $key, $find, string $subject = '')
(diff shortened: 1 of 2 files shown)


kodumbeats (approved):
Code looks good, two questions:

1. Should we use strict matching in these find methods? e.g.
```diff
- if(isset($value[$key]) && $value[$key] == $find)
+ if(isset($value[$key]) && $value[$key] === $find)
```
2. Do these changes indicate we'll be supporting nested documents?

eldadfux:
> Code looks good, two questions:
> 
> 1. Should we use strict matching in these find methods? e.g.
> 
> ```diff
> - if(isset($value[$key]) && $value[$key] == $find)
> + if(isset($value[$key]) && $value[$key] === $find)
> ```
> 
> 1. Do these changes indicate we'll be supporting nested documents?

1. Yes, I think the strict approach is better, I have applied the changes.
2. Currently we only do this to handle JSON string parsed with the `JSON` filter internally

PR #14, opened 2021-05-25
Feat - Script to fill databases

kodumbeats:
Three commands have been added to `bin/` to fill, index, and query the DB to test changes:

- `bin/load` invokes `bin/tasks/load.php`
- `bin/index` invokes `bin/tasks/index.php`
- `bin/query` invokes `bin/tasks/query.php`

To test your DB changes under load:

**=> Load the database**

```bash
docker-compose exec tests bin/load --adapter=[adapter] --limit=[limit] [--name=[name]]

# [adapter]: either 'mongodb' or 'mariadb', no quotes
# [limit]: integer of total documents to generate
# [name]: (optional) name for new database
```

**=> Create indexes**

```bash
docker-compose exec tests bin/index --adapter=[adapter] --name=[name]

# [adapter]: either 'mongodb' or 'mariadb', no quotes
# [name]: name of filled database by bin/load
```

**=> Run Query Suite**

```bash
docker-compose exec tests bin/query --adapter=[adapter] --limit=[limit] --name=[name]

# [adapter]: either 'mongodb' or 'mariadb', no quotes
# [limit]: integer of query limit (default 25)
# [name]: name of filled database by bin/load
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ mock.json
 data-tests.php
 loader.php
 .phpunit.result.cache
+/bin/view/results/
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,12 @@ WORKDIR /usr/src/code
 
 RUN echo extension=swoole.so >> /usr/local/etc/php/conf.d/swoole.ini
 
+RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"
+
+RUN echo "opcache.enable_cli=1" >> $PHP_INI_DIR/php.ini
+
+RUN echo "memory_limit=1024M" >> $PHP_INI_DIR/php.ini
+
 COPY --from=step0 /usr/local/src/vendor /usr/src/code/vendor
 
 # Add Source Code
--- a/README.md
+++ b/README.md
@@ -191,28 +191,28 @@ To test your DB changes under load:
 #### Load the database
 
 ```bash
-docker-compose exec tests bin/load [dbms] [limit]
+docker-compose exec tests bin/load [adapter] [limit]
 
-# [dbms]: either 'mongodb' or 'mariadb', no quotes
+# [adapter]: either 'mongodb' or 'mariadb', no quotes
 # [limit]: integer of total documents to generate
 ```
 
 #### Create indexes
 
 ```bash
-docker-compose exec tests bin/index [dbms] [filledDB]
+docker-compose exec tests bin/index [adapter] [collection]
 
-# [dbms]: either 'mongodb' or 'mariadb', no quotes
-# [filledDB]: name of filled database by bin/load
+# [adapter]: either 'mongodb' or 'mariadb', no quotes
+# [collection]: name of filled database by bin/load
 ```
 
 #### Run Query Suite
 
 ```bash
-docker-compose exec tests bin/query [dbms] [totalDocuments] [filledDB]
+docker-compose exec tests bin/query [adapter] [collection] [limit]
 
-# [dbms]: either 'mongodb' or 'mariadb', no quotes
-# [filledDB]: name of filled database by bin/load
+# [adapter]: either 'mongodb' or 'mariadb', no quotes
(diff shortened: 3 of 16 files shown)


kodumbeats:
Not sure why my first execution was so slow, but while testing on 1million records with a fulltext index, search queries start clocking at < 3 seconds:
```bash
$ dc exec tests bin/query
Creating index
Completed in 0.018397092819214s
Querying
Changing role to 'user4569'
Running query: text.search('Queen')
Found 416 results
Completed in 2.8586227893829s
```

kodumbeats:
@eldadfux Using Swoole to speed up DB loading has a side-effect, and I could use your advice to make this script usable.

We need to use Swoole PDO pools to enable coroutines, but this breaks our current typing:

```bash
> docker-compose exec tests bin/load
Filling database with 80000 documents
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Utopia\Database\Adapter\MariaDB::__construct() must be an instance of PDO, instance of Swoole\Database\PDOProxy given, called in /usr/src/code/bin/tasks/load.php on line 98 and defined in /usr/src/code/src/Database/Adapter/MariaDB.php:29
Stack trace:
#0 /usr/src/code/bin/tasks/load.php(98): Utopia\Database\Adapter\MariaDB->__construct()
```

While working locally, I remove the param type `PDO` here:
https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L29

Do you have any advice on how we could resolve this? I'd rather keep the type definitions if possible.

eldadfux:
A few comments from working with this branch:
1. Use an init file for common behavior across tasks, especially initing the db adapters
2. Use a switch case to init the adapters, instead of multiple `if`s
3. Instead of `dbms` I would use `adapter` to be more consistent
4. Instead of `filledDB` I would use `collection` to be more consistent
5. Its very hard to compare mongo and mariadb write speed when mariadb is leveraging multiple coroutines

eldadfux:
You can use `"swoole/ide-helper": "4.6.6"` dependency for code autocompletion of swoole functions and classes

eldadfux:
It will be more comfortable and make more sense to have `totalDocuments` as the last param here: `docker-compose exec tests bin/query [dbms] [totalDocuments] [filledDB]`

eldadfux:
> It will be more comfortable and make more sense to have `totalDocuments` as the last param here: `docker-compose exec tests bin/query [dbms] [totalDocuments] [filledDB]`

OK, so its just the docs that are wrong :)

eldadfux:
It would be nice to pass a custom name for a collection instead of generating a random one

kodumbeats:
> A few comments from working with this branch:
> 
>     1. Use an init file for common behavior across tasks, especially initing the db adapters
> 
>     2. Use a switch case to init the adapters, instead of multiple `if`s
> 
>     3. Instead of `dbms` I would use `adapter` to be more consistent
> 
>     4. Instead of `filledDB` I would use `collection` to be more consistent
> 
>     5. Its very hard to compare mongo and mariadb write speed when mariadb is leveraging multiple coroutines

1. TODO
2. ✅ 
3. ✅ 
4. I don't think `collection` works here because the database name doesn't match the collection (`articles`)
5. Moved MongoDB loading into similar coroutines as MariaDB

> You can use "swoole/ide-helper": "4.6.6" dependency for code autocompletion of swoole functions and classes

TODO

> OK, so its just the docs that are wrong :)

Docs updated with Utopia\CLI instructions

> It would be nice to pass a custom name for a collection instead of generating a random one

`bin/load` now takes an optional `name` param 👍🏻

kodumbeats:
Current status:
- Dashboard dynamically adds chart and table info based on reports generated in `bin/view/results`
- Graph colors are hard-coded for consistency, but only seven are currently implemented
- Graph colors and table colors match to easily identify results
- Graph is interactive - toggle individual benchmarks by clicking the filename at the top
querydashboard

eldadfux (approved):

PR #15, opened 2021-05-26
Feat MariaDB transactions

eldadfux:
Added transactions support for create, update, and delete document method


kodumbeats (approved):
Impressed by how elegantly PDO handles this ❤️

PR #26, opened 2021-06-18
Feat - Attribute default values

kodumbeats:
This PR introduces a default value when creating an attribute. An exception is thrown if the type of default value does not match the type of attribute, as the database would be unaware of the mismatched type until a document is created.

**Testing**

Unit tests have been added to test for valid and invalid default values.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -735,19 +735,6 @@ class Database
 
         $collection = $this->getCollection($collection);
 
-        /**
-         * Find all attributes with default values
-         *
-         * @var Document[]
-         */
-        $defaults = array_filter($collection->getAttributes()['attributes'], function ($attribute) {
-            return (!\is_null($attribute->getAttribute('default')));
-        });
-
-        foreach ($defaults as $default) {
-            $document->setAttribute($default->getId(), $default->getAttribute('default'));
-        }
-
         $document
             ->setAttribute('$id', empty($document->getId()) ? $this->getId(): $document->getId())
             ->setAttribute('$collection', $collection->getId())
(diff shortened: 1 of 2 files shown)


eldadfux:
Where / How do we assign the default values when no value is given?

kodumbeats:
I've added code to include default values for attributes of both primitives and array of primitives, though I'm not happy with how I wrote this section:

https://github.com/utopia-php/database/pull/26/files#diff-269fc35a258fbfc31c559ec9e4bbda254659f9007d4af68474f5663bbd9eda11R751-R778

We need to run the structure validator once before adding default values because required parameters take precedence. Though, I wasn't sure how to write default values for arrays without decoding, adding defaults, and revalidating the structure. I could use some advice on how to do this more elegantly.

kodumbeats:
I thought through my dilemma and realized that this edge case doesn't make sense. 

The problem arose because I couldn't find a good way to insert default values _and_ check for required attributes without duplicating work. However, I realized that having a default value is mutually exclusive of requiring an attribute, so I think it's appropriate to throw an exception in case that happens:

https://github.com/utopia-php/database/blob/feat-attribute-default-values/src/Database/Database.php#L382-L385

eldadfux:
Solution looks solid, I have a hunch that we can improve performance by using the filter loop to apply the default values, it will be great if we can have a benchmark for it.

kodumbeats:
Wasn't sure how to benchmark this, but I agree that this should be more performant (reducing the number of loops over `$collection->getAttribute('attributes');`. Implemented in df9adf6

PR #27, opened 2021-06-21
Feat - Expand Permissions validator

kodumbeats:
This PR expands the permissions validator to check the format of the permission strings. Since the removal of the wildcard char `*`, a "role" satisfies the following simple schema:
- Can be expressed in the format `{$type}:{$value}` with only a single colon separating the permission type from its value
- `$type` must be one of `['member', 'role', 'user', 'team']`
- `$value` must not be empty
- `$value` must satisfy the Key validator for `user` and `member`
- `$value` must be one of `['all', 'guest', 'member']` when `$type = 'role'`
- Since `team:[teamId]/[role]` can have any string for role, no validation needed

### Test plan
Unit tests included

### DIFF
--- a/src/Database/Validator/Permissions.php
+++ b/src/Database/Validator/Permissions.php
@@ -88,8 +88,8 @@ class Permissions extends Validator
              * Substring after ":" $value must not be empty and satisty requirements per $type
              */
 
-            $type = \substr($role, 0, \strpos($role, ':'));
-            $value = \substr($role, \strpos($role, ':') + 1);
+            $type = \substr($role, 0, $pos);
+            $value = \substr($role, $pos + 1);
 
             if (strlen($value) === 0) {
                 $this->message = 'Permission role value must not be empty';
--- a/src/Database/Validator/Permissions.php
+++ b/src/Database/Validator/Permissions.php
@@ -110,7 +110,7 @@ class Permissions extends Validator
                     // user:$id and member:$id must be valid Keys
                     $key = new Key();
                     if (!$key->isValid($value)) {
-                        $this->message = 'Permission role must contain a valid $id: ' . $key->getDescription();
+                        $this->message = '[role:$id] $id must be a valid key: ' . $key->getDescription();
 
                         return false;
                     }
@@ -124,7 +124,7 @@ class Permissions extends Validator
 
                     // if no team role is given and and $id is not valid
                     if ($pos === false && !$key->isValid($value)) {
-                        $this->message = 'Permission role must contain a valid $id: ' . $key->getDescription();
+                        $this->message = '[role:$id] $id must be a valid key: ' . $key->getDescription();
 
                         return false;
                     }
--- a/src/Database/Validator/Permissions.php
+++ b/src/Database/Validator/Permissions.php
@@ -159,7 +159,7 @@ class Permissions extends Validator
 
                         // $teamId and $teamRole must both be valid Keys
                         if (!$key->isValid($teamId) || !$key->isValid($teamRole)) {
-                            $this->message = 'Permission role for team must contain valid teamID and role: ' . $key->getDescription();
+                            $this->message = '[team:$teamId/$role] $teamID and $role must be valid keys: ' . $key->getDescription();
 
                             return false;
                         }
--- a/src/Database/Validator/Permissions.php
+++ b/src/Database/Validator/Permissions.php
@@ -3,6 +3,7 @@
 namespace Utopia\Database\Validator;
 
 use Utopia\Validator;
+use Utopia\Database\Validator\Key;
 
 class Permissions extends Validator
 {
@@ -19,6 +20,15 @@ class Permissions extends Validator
         'role',
         'team',
         'user',
+    ];
+
+    /**
+     * @var string[]
+     */
+    protected $roles = [
+        'all',
+        'guest',
+        'member',
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


eldadfux:
@kodumbeats Maybe we should introduce some rules for the `role` part? maybe same rules as for a key?

src/Database/Validator/Permissions.php
```diff
@@ -98,19 +98,17 @@ public function isValid($roles)
...
                     }
                     break;
+                case 'team':
                 case 'user':
                 case 'member':
-                    // member:[memberId] and user:[userId] must be valid keys
+                    // every valid $value must be a valid Key
                     $key = new Key();
                     if (!$key->isValid($value)) {
                         $this->message = $key->getDescription();
```
eldadfux, line 114: Not sure the message here is verbose enough, maybe we should have a custom one.

tests/Database/Validator/PermissionsTest.php
```diff
@@ -106,7 +106,7 @@ public function testInvalidPermissions()
         $this->assertEquals($object->getDescription(), 'Parameter must contain only letters with no spaces or special chars and be shorter than 32 chars');
         $this->assertEquals($object->isValid(['user:12&4']), false);
         $this->assertEquals($object->getDescription(), 'Parameter must contain only letters with no spaces or special chars and be shorter than 32 chars');
-        $this->assertEquals($object->isValid(['user:ab(124']), false);
+        $this->assertEquals($object->isValid(['team:ab(124']), false);
```
eldadfux, line 109: Maybe we should add more tests for the new role validation, for example: `team:xyz/hello world` -> false

eldadfux (changes requested):

tests/Database/Validator/PermissionsTest.php
kodumbeats: Of course 🤯 I incorrectly assumed the form of the teams permission as `'team:teamID' OR 'team:role'`. Will improve this testing.

src/Database/Validator/Permissions.php
```diff
@@ -45,6 +65,62 @@ public function isValid($roles)
 
                 return false;
             }
+
+            if ($role === '*') {
```
christyjacob4, line 69: This is a nice touch :wink:

src/Database/Validator/Permissions.php
```diff
@@ -45,6 +65,62 @@ public function isValid($roles)
...
+                        $this->message = 'Permission roles must be one of: ' . \implode(", ", $this->roles);
+                        return false;
+                    }
+                    break;
+                case 'team':
+                case 'user':
+                case 'member':
+                    // every valid $value must be a valid Key
+                    $key = new Key();
+                    if (!$key->isValid($value)) {
```
christyjacob4, line 113: Is the Key() validator able to check if the `id` exists in the database? 
Do we plan to have support for that ?

christyjacob4 (approved):
Looks really clean :superhero: . Nice one!

src/Database/Validator/Permissions.php
kodumbeats: I'm not sure we want to do that in this library, with separation of concerns from Appwrite and such. @eldadfux what do you think?

src/Database/Validator/Permissions.php
kodumbeats: Is this appropriately verbose? 951ea03

tests/Database/Validator/PermissionsTest.php
kodumbeats: Added more validation+testing in 2756069

src/Database/Validator/Permissions.php
```diff
@@ -45,6 +65,111 @@ public function isValid($roles)
...
+                    // if "/" is at index zero
+                    if ($pos === 0) {
+                        $this->message = 'Team ID must not be empty.';
+
+                        return false;
+                    }
+
+                    // if a "/" is found, ensure is unique and both substrings are valid
+                    if ($pos > 0) {
+                        // Split into format {$teamId}:{$teamRole}
```
lohanidamodar, line 141:
```suggestion
                        // Split into format {$teamId}/{$teamRole}
```

lohanidamodar (commented):
LGTM

src/Database/Validator/Permissions.php
eldadfux: Not sure I feel like its clear enough that we're talking about the role part: team:[teamid]/**role** Maybe emphasizing: "The role section has..."

tests/Database/Validator/PermissionsTest.php
eldadfux: Do we allow `_` if not in the first character?

src/Database/Validator/Permissions.php
eldadfux: I agree, and I would also be concerned about the performance cost. + we might want to validate the ID with no regard to the database (team role for example)

tests/Database/Validator/PermissionsTest.php
kodumbeats: If I'm reading this regex correctly, yes:
https://github.com/utopia-php/database/blob/feat-permission-validator-expanded/src/Database/Validator/Key.php#L45-L46

src/Database/Validator/Permissions.php
kodumbeats: Tried again in f349a13 🤔

PR #28, opened 2021-06-23
Feat find and delete

lohanidamodar:
First working draft of Find and Delete method.
 - I had to simplify and hack my way but it's working, mariadb DELETE with WHERE/ORDER and LIMIT not working with table alias
- I'm handling limit internally in mysql and mariadb.
 - Also removed orders for now for the shake of simplicity also, if we are not exposing limit, that means we are deleting everything query finds, so orders doesn't really matter?

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -250,7 +250,7 @@ abstract class Adapter
      *
      * @return Document[]
      */
-    abstract public function findAndDelete(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): bool;
+    abstract public function findAndDelete(string $collection, array $queries = [], int $limit = 25, array $orderAttributes = [], array $orderTypes = []): bool;
     
     /**
      * Count Documents
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -237,6 +237,22 @@ abstract class Adapter
     abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): array;
     
     /**
+     * Find and Delete Documents
+     *
+     * Find and delete data sets using chosen queries
+     *
+     * @param string $collection
+     * @param \Utopia\Database\Query[] $queries
+     * @param int $limit
+     * @param int $offset
+     * @param array $orderAttributes
+     * @param array $orderTypes
+     *
+     * @return Document[]
+     */
+    abstract public function findAndDelete(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): bool;
+    
+    /**
      * Count Documents
      * 
      * @param string $collection
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -250,7 +250,7 @@ abstract class Adapter
      *
      * @return Document[]
      */
-    abstract public function findAndDelete(string $collection, array $queries = [], int $limit = 25, array $orderAttributes = [], array $orderTypes = []): bool;
+    abstract public function findAndDelete(string $collection, array $queries = []): bool;
     
     /**
      * Count Documents
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -497,6 +497,77 @@ class MariaDB extends Adapter
     }
 
     /**
+     * Find and Delete Documents
+     *
+     * Find and delete data sets using chosen queries
+     *
+     * @param string $collection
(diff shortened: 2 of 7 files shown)


src/Database/Adapter.php
```diff
@@ -236,6 +236,22 @@ abstract public function deleteDocument(string $collection, string $id): bool;
...
+     * Find and Delete Documents
+     *
+     * Find and delete data sets using chosen queries
+     *
+     * @param string $collection
+     * @param \Utopia\Database\Query[] $queries
+     * @param int $limit
+     * @param int $offset
+     * @param array $orderAttributes
+     * @param array $orderTypes
```
kodumbeats, line 249: Let's make sure the docblocks match the method params
```suggestion
```

src/Database/Adapter/MariaDB.php
```diff
@@ -496,6 +496,75 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
...
+     *
+     * Find and delete data sets using chosen queries
+     *
+     * @param string $collection
+     * @param \Utopia\Database\Query[] $queries
+     * @param int $limit
+     * @param int $offset
+     * @param array $orderAttributes
+     * @param array $orderTypes
+     * @param bool $count
```
kodumbeats, line 510: Let's make sure the docblocks match the method params
```suggestion
```

src/Database/Adapter/MariaDB.php
```diff
@@ -767,13 +836,13 @@ protected function getSQLIndex(string $collection, string $id,  string $type, ar
      * 
      * @return string
      */
-    protected function getSQLPermissions(array $roles): string
+    protected function getSQLPermissions(array $roles, $table = 'table_main'): string
```
kodumbeats, line 839: Method params do not match the docblock.

kodumbeats, line 839: Also, why did we add a `$table` param here? I think it confuses the story of the code, and it isn't used at all in the MySQL method of the same name: 
https://github.com/utopia-php/database/pull/28/files#diff-261256e367d93c8cd8c41e042394eec200852acb2737c27f798a97b87e414c0fR89-R96

src/Database/Adapter/MongoDB.php
```diff
@@ -392,6 +392,40 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
...
+
+        // queries
+        $filters = $this->buildFilters($queries);
+
+        // permissions
+        if (Authorization::$status) { // skip if authorization is disabled
+            $filters['_read']['$in'] = Authorization::getRoles();
+        }
+
+        $collection->deleteMany($filters);
```
kodumbeats, line 425: Should we find a way to delete in batches? This method will attempt to remove all matching documents:
https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-deleteMany/

src/Database/Adapter/MySQL.php
```diff
@@ -79,10 +82,11 @@ protected function getSQLIndex(string $collection, string $id, string $type, arr
      * @param string $operator
      * @param string $placeholder
      * @param mixed $value
+     * @param string $table
      * 
      * @return string
      */
-    protected function getSQLPermissions(array $roles): string
+    protected function getSQLPermissions(array $roles, $table = "table_main"): string
```
kodumbeats, line 89: `$table` param is unused in this method.
```suggestion
    protected function getSQLPermissions(array $roles): string
```

src/Database/Adapter/MySQL.php
```diff
@@ -79,10 +82,11 @@ protected function getSQLIndex(string $collection, string $id, string $type, arr
      * @param string $operator
      * @param string $placeholder
      * @param mixed $value
+     * @param string $table
```
kodumbeats, line 85: $table param is unused in this method.
```suggestion
```

kodumbeats (changes requested):
:)

src/Database/Adapter/MySQL.php
lohanidamodar: Yes but used in Mariadb ,ad not having it here produces error as the method is override

src/Database/Adapter/MariaDB.php
lohanidamodar: This was because the mariadb in `DELETE` query with ` LIMIT` did not support table alias, using table alias if I add ORDER or LIMIT I get error. I had to use the actual table name and roles query needed the table name as well

src/Database/Adapter/MongoDB.php
lohanidamodar: I'm open to suggestions?
(We do want to delete all the matching documents right? At least that is what we are doing with MariaDB and MySQL with internal loop with limit)

PR #30, opened 2021-06-28
Allow for creation of collection with complete schema

kodumbeats:
Creating collections with many attributes (for example, on startup of Appwrite, for testing) can be slow, so this PR adds additional params to `createCollection` to generate a complete collection with one call:
```php
abstract public function createCollection(string $name, array $attributes = [], array $indexes = []): bool;
```

**Testing**
Additional unit tests will be added to validate functionality.
<details>
<summary>Screenshots of index creation from tests</summary>

mariadbcollectionwithschema

defaultschemainmongo

</details>

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -105,16 +105,65 @@ class MariaDB extends Adapter
     {
         $id = $this->filter($id);
 
-        $this->getPDO()
-            ->prepare("CREATE TABLE IF NOT EXISTS {$this->getNamespace()}.{$id} (
-                `_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
-                `_uid` CHAR(255) NOT NULL,
-                `_read` TEXT NOT NULL,
-                `_write` TEXT NOT NULL,
-                PRIMARY KEY (`_id`),
-                UNIQUE KEY `_index1` (`_uid`)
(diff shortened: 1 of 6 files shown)


kodumbeats:
I'm not 💯 convinced that Mongo is working properly, still need some more testing around it.

src/Database/Adapter/MariaDB.php
```diff
@@ -608,6 +658,26 @@ public function getSupportForCasting(): bool
         return false;
     }
 
+    /**
+     * Returns the attribute type for read permissions
+     *
+     * @return string
+     */
+    public function getTypeForReadPermission(): string
```
eldadfux, line 666:
```suggestion
    public function getTypeForReadPermission(): string
```

This should be a protected method used internally only by this class.

src/Database/Adapter/MariaDB.php
```diff
@@ -608,6 +658,26 @@ public function getSupportForCasting(): bool
...
+    {
+        return "TEXT";
+    }
+
+    /**
+     * Returns the index type for read permissions
+     *
+     * @return string
+     */
+    public function getIndexTypeForReadPermission(): string
```
eldadfux, line 676: This should be a protected method used internally only by this class.

eldadfux (changes requested):

PR #33, opened 2021-07-05
Feat - Index limits

kodumbeats:
Each database has a maximum number of indexes per collection/table. This PR adds a new exception IndexLimit (not sure on the name), implemented per adapter, so these cases can be appropriately handled.

**Testing**
Additional tests have been added for these conditions.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -260,22 +260,9 @@ class MariaDB extends Adapter
             $attribute = "`{$attribute}`{$length} {$order}";
         }
 
-        try {
-            $success = $this->getPDO()
-                ->prepare($this->getSQLIndex($name, $id, $type, $attributes))
-                ->execute();
-        } catch (PDOException $e) {
-            switch ($e->getCode()) {
-                case 42000:
-                    throw new IndexLimit('Index limit reached. Cannot create new index.');
-                    break;
-                default:
-                    throw $e;
-                    break;
-            }
-        }
-
-        return $success;
+        return $this->getPDO()
+            ->prepare($this->getSQLIndex($name, $id, $type, $attributes))
+            ->execute();
     }
 
     /**
@@ -662,6 +649,33 @@ class MariaDB extends Adapter
     public function getSupportForFulltextIndex(): bool
     {
         return true;
+    }
+
+    /**
+     * Get current index count.
+     * 
+     * @param string $collection collectionID
+     * @return int
+     */
+    public function getIndexCount(string $collection): int
+    {
+        $name = $this->filter($collection);
+        
+        $stmt = $this->getPDO()->prepare("SHOW KEYS FROM {$this->getNamespace()}.{$name}");
+
+        $stmt->execute();
+
+        return \count($stmt->fetchAll(PDO::FETCH_ASSOC));
+    }
+
+    /**
+     * Get maximum index limit.
+     * 
+     * @return int
+     */
+    public function getIndexLimit(): int
+    {
+        return 64;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


eldadfux:
I think this check should only be implemented on the database level and not in the adapter level. We can check the number of indexes already exists in the collection object and throw the relevant exception.

This implementation also assumes all adapters have a limit of 64 indexes which might not always be the case (SQLite has no limit). Each adapter should implement `getIntLimit():int` method to return the real max indexes allowed as an integer.

We also must remember that we have two doorways to create a new index, directly and using the `createCollection` method. This check should apply to both.

src/Database/Adapter/MongoDB.php
```diff
@@ -277,8 +278,13 @@ public function createIndex(string $collection, string $id, string $type, array
                     return false;
             }
         }
+        try{
+            $success = $collection->createIndex($indexes, $options);
+        } catch (\MongoDB\Driver\Exception\CommandException $e) {
```
eldadfux, line 283: ... and we can't assume the exception here is specifically because of the index limit reached, this might be very misleading if we try to debug an index creation error.

kodumbeats:
> I think this check should only be implemented on the database level and not in the adapter level. We can check the number of indexes already exists in the collection object and throw the relevant exception.
> 
> This implementation also assumes all adapters have a limit of 64 indexes which might not always be the case (SQLite has no limit). Each adapter should implement `getIntLimit():int` method to return the real max indexes allowed as an integer.
> 
> We also must remember that we have two doorways to create a new index, directly and using the `createCollection` method. This check should apply to both.

I tried implementing this at first, but the problem is that we don't store the default indexes in the `$collection` document, so we don't have an accurate count:
- MariaDB/MySQL creates indexes on `_id`, `_uid`, and `_read`
- MongoDB creates indexes on `_id` (by default), `_read`, and `_write`

We can either:
- store the default indexes in the `$collection` document or
- stay at the Adapter level and get those from the database itself (`MongoDB\Collection::listIndexes()` as an example)

eldadfux:
I think we can just have a method on each adapter to let us know how many indexes we're already using, and calculate how many remain available.

src/Database/Adapter/MongoDB.php
```diff
@@ -680,6 +680,30 @@ public function getSupportForFulltextIndex(): bool
...
+        $name = $this->filter($collection);
+        $indexes = $this->getDatabase()->$name->listIndexes();
+
+        return \iterator_count($indexes);
+    }
+
+    /**
+     * Get maximum index limit.
+     * 
+     * @return int
```
eldadfux, line 700: Would be a very cool addition to have a link here with reference to official docs.

src/Database/Adapter/MariaDB.php
```diff
@@ -648,6 +650,33 @@ public function getSupportForFulltextIndex(): bool
...
+
+        return \count($stmt->fetchAll(PDO::FETCH_ASSOC));
+    }
+
+    /**
+     * Get maximum index limit.
+     * 
+     * @return int
+     */
+    public function getIndexLimit(): int
```
eldadfux, line 675: Would be a very cool addition to have a link here with reference to official docs.

src/Database/Adapter/MariaDB.php
```diff
@@ -648,6 +650,33 @@ public function getSupportForFulltextIndex(): bool
...
     }
 
+    /**
+     * Get current index count.
+     * 
+     * @param string $collection collectionID
+     * @return int
+     */
+    public function getIndexCount(string $collection): int
+    {
```
eldadfux, line 660: I think we can calculate this with a hardcoded number + the number of indexes we have in $collection object (+ the ones pending in the queue). It will help us reduce an extra (not cached) call to the DB.

eldadfux: Same for the Mongo adapter

eldadfux (changes requested):

eldadfux (approved):

PR #35, opened 2021-07-07
Feat - Attribute/column limits

kodumbeats:
This PR introduces limits on attribute creation:

- [x] Enforces InnoDB's limit of 1017 columns per table
- [x] Enforces MySQL/MariaDB limit of 65535 bytes of storage per row
    We provide ~300bytes for buffer to throw LimitException from memory before hitting the database

Goals:
- Validate attribute and index creation without a second query to the database
- Check limits when adding to queue for error handling in Appwrite

Tests:
- [x] Columns per table
- [x] Bytes per row

References:
- https://mariadb.com/kb/en/innodb-limitations/
- https://mariadb.com/kb/en/data-type-storage-requirements/
- https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html


kodumbeats:
I've got tests in place to catch the exact byte that trips over the 65535 limit, but my math isn't quite right yet to calculate it without a PDOException.

> I'm avoiding the PDOException because I need to validate the column width in-memory in Appwrite from the main container, not just the Database worker.

kodumbeats:
None of the math adds up, so we'll need to decide on an acceptable buffer to keep the code simple. Otherwise we'll be implementing something like this: https://dba.stackexchange.com/questions/114471/calculate-row-size-and-max-row-size-for-a-table

src/Database/Database.php
```diff
@@ -345,6 +345,18 @@ public function createAttribute(string $collection, string $id, string $type, in
     {
         $collection = $this->getCollection($collection);
 
+        if ($this->adapter->getAttributeLimit() > 0 &&
```
eldadfux, line 348: We have 3 entry points for creating attributes, this method, the `addAttributeInQueue` method and the `createCollection`. We should make sure validation happens on all.

src/Database/Database.php
```diff
@@ -1117,6 +1141,85 @@ public function casting(Document $collection, Document $document):Document
...
 
+    /**
+     * Estimate maximum number of bytes required to store a document in $collection.
+     * Byte requirement varies based on column type and size.
+     * Needed to satisfy MariaDB/MySQL row width limit.
+     * 
+     * @param Document $collection
+     * @return int
+     */
+    protected function getAttributeWidth(Document $collection): int
```
eldadfux, line 1152: Are we sure this method should be here? This logic seems very coupled with MySQL/MariaDB.

tests/Database/Base.php
```diff
@@ -1145,44 +1150,169 @@ public function testWritePermissionsUpdateFailure(Document $document)
         return $document;
     }
 
+    public function testExceptionAttributeLimit()
+    {
+        if (static::getAdapterName() === 'mariadb' || static::getAdapterName() === 'mysql') {
```
eldadfux, line 1155: We should probably just check if this limit apply on current adapter rather than checking specific adapters, this tests will be easily skipped on any new future adapter.

tests/Database/Base.php
```diff
@@ -1145,44 +1150,169 @@ public function testWritePermissionsUpdateFailure(Document $document)
...
+        ];
+    }
+
+    /**
+     * @dataProvider rowWidthExceedsMaximum
+     * @expectedException LimitException
+     */
+    public function testExceptionWidthLimit($key, $stringSize, $stringCount, $intCount, $floatCount, $boolCount)
+    {
+        if (static::getAdapterName() === 'mariadb' || static::getAdapterName() === 'mysql') {
```
eldadfux, line 1208: We should probably just check if this limit apply on current adapter rather than checking specific adapters, this tests will be easily skipped on any new future adapter.

src/Database/Database.php
kodumbeats: I wasn't happy with it here either, but I'd like to discuss why I chose this path.

In Appwrite, the `createAttribute` route in the Database controller calls `addAttributeInQueue()`, which is a method that exists in the Database class and not the Adapters:

https://github.com/appwrite/appwrite/blob/feat-db-refactor-routes/app/controllers/api/database.php#L266

After adding the attribute in queue, the Database worker is triggered, which handles `adapter->createAttribute()` and `database->removeAttributeInQueue()` when successful. If this method `getAttributeWidth()` was in the adapter level, I don't know how to to throw an error from the Appwrite request, instead it would be thrown in the Database worker and not communicated back to the end user.

With this method in the Database class, I can throw an error immediately from the Appwrite route, providing better feedback and DX.

Do you have ideas on how this can be more elegantly handled?

eldadfux: Will it be possible to have as a public method in the adapter (also part of the interface) and call it from `addAttributeInQueue` using the adapter?

kodumbeats: I had avoided that approach because of the increasing complexity of adding a new adapter. I've refactored this logic to the adapter layer in 6bf458f

eldadfux (approved):

PR #36, opened 2021-07-13
Throw exception on `create` when database already exists

abnegate:
If database already exists but the utopia required collections, attributes and indices do not; continue to create them anyway

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -205,8 +205,8 @@ class Database
      */
     public function create(): bool
     {
-        if (!$this->adapter->exists()) {
-            $this->adapter->create();
+        if ($this->adapter->exists()) {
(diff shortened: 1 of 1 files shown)


kodumbeats (approved):

eldadfux:
Would it make sense if we just throw an exception when this conflict happens?

abnegate:
I think either way makes sense, but from my perspective continuing with creation lowers the barrier of entry

eldadfux:
@abnegate after chatting about this with @kodumbeats, I think we should go with an exception for better consistency.

abnegate:
@eldadfux No worries have updated to throw now

kodumbeats:
Force-pushed to rebase `fix-db-exists` on `main` to check tests.

PR #37, opened 2021-07-17
Fix key name

eldadfux:
Changed generic key name to `key` instead of `text`

### DIFF
--- a/tests/Database/Validator/QueriesTest.php
+++ b/tests/Database/Validator/QueriesTest.php
@@ -119,7 +119,7 @@ class QueriesTest extends TestCase
         'indexesInQueue' => [
             [
                 '$id' => 'testindex4',
-                'type' => 'tekeyxt',
+                'type' => 'key',
                 'attributes' => [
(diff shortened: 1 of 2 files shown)


tests/Database/Validator/QueriesTest.php
```diff
@@ -119,7 +119,7 @@ class QueriesTest extends TestCase
         'indexesInQueue' => [
             [
                 '$id' => 'testindex4',
-                'type' => 'text',
+                'type' => 'tekeyxt',
```
kodumbeats, line 122: Is `tekeyxt` intentional for testing?

kodumbeats (changes requested):
One question :)

tests/Database/Validator/QueriesTest.php
eldadfux: No, and yet the tests are green 😬

eldadfux: fixed

PR #39, opened 2021-07-20
Refactor findFirst to findOne

kodumbeats:
`Utopia\Database\Database->findFirst()` is primarily used to find a single document, and `findLast()` isn't used at all. 

This PR renames `findFirst` to `findOne` (similar to [mongo](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-findOne/)) and removes `findLast`.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -873,32 +873,24 @@ class Database
     /**
      * @param string $collection
      * @param array $queries
-     * @param int $limit
-     * @param int $offset
-     * @param array $orderAttributes
-     * @param array $orderTypes
      *
      * @return Document|bool
      */
-    public function findFirst(string $collection, array $queries = [], int $limit = 1, int $offset = 0, array $orderAttributes = [], array $orderTypes = [])
+    public function findFirst(string $collection, array $queries = [])
     {
-        $results = $this->find($collection, $queries, $limit, $offset, $orderAttributes, $orderTypes);
+        $results = $this->find($collection, $queries, /*limit*/ 1);
         return \reset($results);
     }
 
     /**
      * @param string $collection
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -880,7 +880,7 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
      *
      * @return Document|bool
      */
-    public function findFirst(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [])
+    public function findFirst(string $collection, array $queries = [], int $limit = 1, int $offset = 0, array $orderAttributes = [], array $orderTypes = [])
```
eldadfux, line 883: Actually I think we can completely remove it and just hardcode the number one when we call the `find` method.
Same should probably apply for `findLast`

eldadfux (changes requested):

src/Database/Database.php
kodumbeats: I've updated this in 92107b8, though as written now, `findFirst()` and `findLast()` are functionally identical. I couldn't find any instances of calling `findLast()` - should we remove it?

eldadfux:
Maybe we can follow this nice convention and have one method named `findOne`?
https://docs.mongodb.com/manual/reference/method/db.collection.findOne/

src/Database/Database.php
eldadfux: Maybe we can follow this nice convention and have one method named `findOne`?
https://docs.mongodb.com/manual/reference/method/db.collection.findOne/

eldadfux (approved):

PR #40, opened 2021-07-21
feat(find): after document pagination

TorstenDittmann:
- allow queries to paginate after a specific document
- for MariaDB, we are using the internal `_id` as a subsequent value to query after a certain document while allowing to sort
- since we need the internal `_id` and every attribute and its proper datatype, a document need to be passed instead of only its Document ID

References can be found here: https://shopify.engineering/pagination-relative-cursors

For MongoDB we are using the same approach, except that we are using the native `_id` as a subsequent value.

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -233,6 +233,7 @@ abstract class Adapter
      * @param int $offset
      * @param array $orderAttributes
      * @param array $orderTypes
+     * @param Document|null $orderAfter
      *
      * @return Document[]
      */
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -236,8 +236,8 @@ abstract class Adapter
      *
      * @return Document[]
      */
-    abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): array;
-    
+    abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], Document $orderAfter = null): array;
+
     /**
      * Count Documents
      * 
@@ -319,7 +319,7 @@ abstract class Adapter
         if(\is_null($value)) {
             throw new Exception('Failed to filter key');
         }
-        
+
         return $value;
     }
 }
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -251,14 +251,6 @@ abstract class Adapter
     abstract public function count(string $collection, array $queries = [], int $max = 0): int;
 
     /**
-     * Get internal ID of a document.
-     * 
-     * @param Document $document 
-     * @return mixed 
-     */
-    abstract public function getInternalId(Document $document);
-
-    /**
      * Get max STRING limit
      * 
      * @return int
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -253,9 +253,9 @@ abstract class Adapter
      * Get internal ID of a document.
      * 
      * @param Document $document 
-     * @return int 
+     * @return mixed 
      */
-    abstract public function getInternalId(Document $document): int;
+    abstract public function getInternalId(Document $document);
 
     /**
      * Get max STRING limit
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -233,11 +233,11 @@ abstract class Adapter
      * @param int $offset
      * @param array $orderAttributes
      * @param array $orderTypes
-     * @param Document|null $orderAfter
+     * @param array $orderAfter
      *
      * @return Document[]
      */
-    abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], Document $orderAfter = null): array;
+    abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], array $orderAfter = []): array;
 
     /**
      * Count Documents
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -250,6 +250,14 @@ abstract class Adapter
     abstract public function count(string $collection, array $queries = [], int $max = 0): int;
 
     /**
+     * Get internal ID of a document.
+     * 
+     * @param Document $document 
+     * @return int 
+     */
+    abstract public function getInternalId(Document $document): int;
+
+    /**
      * Get max STRING limit
      * 
      * @return int
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -480,16 +480,18 @@ class MariaDB extends Adapter
      * Find data sets using chosen queries
      *
      * @param string $collection
-     * @param \Utopia\Database\Query[] $queries
+     * @param array $queries
      * @param int $limit
      * @param int $offset
      * @param array $orderAttributes
      * @param array $orderTypes
-     * @param bool $count
+     * @param array $orderAfter
      *
-     * @return Document[]
+     * @return array 
+     * @throws Exception 
+     * @throws PDOException 
      */
-    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], Document $orderAfter = null): array
+    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], array $orderAfter = []): array
     {
         $name = $this->filter($collection);
         $roles = Authorization::getRoles();
@@ -510,7 +512,7 @@ class MariaDB extends Adapter
                         OR (
                             {$attribute} = :after 
                             AND
-                            _id > {$orderAfter->getInternalId()}
+                            _id > {$orderAfter['$internalId']}
                         )
                     )";
             }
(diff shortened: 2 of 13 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -488,17 +492,40 @@ public function deleteDocument(string $collection, string $id): bool
...
+    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], Document $orderAfter = null): array
     {
         $name = $this->filter($collection);
         $roles = Authorization::getRoles();
         $where = ['1=1'];
         $orders = [];
-        
+        $applyOrderAfter = !empty($orderAfter) && array_key_exists(0, $orderAttributes);
+
+        if ($applyOrderAfter && is_null($orderAfter->getAttribute('_id', null))) {
```
eldadfux, line 503: With the new method this check might be redundant

src/Database/Adapter/MariaDB.php
```diff
@@ -527,6 +554,14 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
             }
         }
 
+        if ($applyOrderAfter) {
+            $attribute = $orderAttributes[0];
```
eldadfux, line 558: Better use null safety operator to avoid any warnings

eldadfux (changes requested):

src/Database/Adapter/MariaDB.php
TorstenDittmann: Its null safe due to the check here https://github.com/utopia-php/database/blob/be48f14e413e6f3a807e0d5189e2a6cd9cbbda6e/src/Database/Adapter/MariaDB.php#L501

src/Database/Adapter/MariaDB.php
```diff
@@ -292,10 +292,10 @@ public function deleteIndex(string $collection, string $id): bool
      *
      * @return Document
      */
-    public function getDocument(string $collection, string $id): Document
+    public function getDocument(string $collection, string $id, bool $includeInternals = false): Document
```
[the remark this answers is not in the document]
TorstenDittmann: Okay 👍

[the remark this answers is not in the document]
TorstenDittmann: Done.

TorstenDittmann:
According to the reviews I fixed the workflow. Fetching the `orderAfter` document will now be happening automatically, if the param is passed a document id.

That way its cleaner in the controllers.

src/Database/Adapter.php
```diff
@@ -249,6 +250,14 @@ abstract public function find(string $collection, array $queries = [], int $limi
      */
     abstract public function count(string $collection, array $queries = [], int $max = 0): int;
 
+    /**
+     * Get internal ID of a document.
+     * 
+     * @param Document $document 
+     * @return mixed 
+     */
+    abstract public function getInternalId(Document $document);
```
eldadfux, line 259: I think we should completely avoid this method because it forces us to send another query to the database, and also move some business logic to the adapter where we would like to have on the main DB class. This design allow us to make the creation of new adapter as simple as possible if possible.

Instead I would just make sure that the original getDocument methods from the adapter will map `_id` to `$internalId` and return it with document. No need to expose `$internalId` in any response.

src/Database/Adapter/MongoDB.php
```diff
@@ -107,14 +107,15 @@ public function createCollection(string $name, array $attributes = [], array $in
 
         $collection = $database->$id;
 
-        // Mongo creates an index for _id; index _read,_write by default
+        // Mongo creates an index for _id; _read and _write by default
         // Returns the name of the created index as a string.
         // Update $this->getIndexCount when adding another default index
+        $uid = $collection->createIndex(['_uid' => $this->getOrder(Database::ORDER_DESC)], ['name' => '_uid', 'unique' => true]);
```
eldadfux, line 113: Isn't _uid our primary key already? @kodumbeats

src/Database/Adapter/MongoDB.php
```diff
@@ -428,8 +428,43 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
             $options['sort'][$attribute] = $orderType;
         }
 
+        $options['sort']['_id'] = $this->getOrder(Database::ORDER_ASC);
+
         // queries
         $filters = $this->buildFilters($queries);
+        if (empty($orderAttributes) && !empty($orderAfter)) {
```
eldadfux, line 435: I would love to discuss this section more in detail.

src/Database/Database.php
```diff
@@ -921,14 +921,25 @@ public function deleteDocument(string $collection, string $id): bool
      * @param int $offset
      * @param array $orderAttributes
      * @param array $orderTypes
+     * @param string $orderAfter
      *
      * @return Document[]
      */
-    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = []): array
+    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], string $orderAfter = null): array
```
eldadfux, line 928: It would be hard to return a proper error to the HTTP response. We can either pass a document instance and validate it in the controller before we pass it here, or we can throw a custom exception and make sure we catch it in the relevant route and delegate a proper HTTP error.

src/Database/Document.php
```diff
@@ -57,6 +57,23 @@ public function getId(): string
...
+    }
+
+    /**
+     * @param mixed $internalId 
+     * @return void 
+     */
+    public function setInternalId($internalId): void
+    {
+        $this->setAttribute('$internalId', $internalId);
+    }
```
eldadfux, line 75: If we end up using this as a first class citizen attribute maybe we can drop the dedicated `set` method for it.

src/Database/Document.php
```diff
@@ -91,7 +108,7 @@ public function getAttributes(): array
         $attributes = [];
 
         foreach ($this as $attribute => $value) {
-            if(array_key_exists($attribute, ['$id' => true, '$collection' => true, '$read' => true, '$write' => []])) {
+            if(array_key_exists($attribute, ['$id' => true, '$internalId' => true, '$collection' => true, '$read' => true, '$write' => []])) {
```
eldadfux, line 111: 👌

tests/Database/Adapter/MongoDBTest.php
```diff
@@ -39,6 +39,15 @@ static function getAdapterRowLimit(): int
         return MongoDB::getRowLimit();
     }
 
+    /**
+     * 
+     * @return int 
+     */
+    static function getUsedIndexes(): int
```
eldadfux, line 46: Not sure how we use this methods in the test suite is this part of another PR?

eldadfux (changes requested):

src/Database/Adapter/MongoDB.php
TorstenDittmann: The query for after pagination depends on what the first order attribute is. If there is no order attribute - we need to add a special case to make after pagination working 👍🏻

src/Database/Adapter/MongoDB.php
TorstenDittmann: We have overwritten the automatic `_id` from mongo with ours. But we need to take advantage of the auto generated `_id` by MongoDB now for pagination 👍🏻 

Thats why I added the `_uid` attribute.

tests/Database/Adapter/MongoDBTest.php
TorstenDittmann: We are testing the maximum number of indexes in the tests. Previously all Adapters had the same amount of used indexes by default. Now with MongoDB we added another index and we need to tell the tests how many indexes are used by default 👍🏻

src/Database/Database.php
TorstenDittmann: Fixed 👍🏻

src/Database/Document.php
TorstenDittmann: Fixed 👍🏻

src/Database/Adapter.php
TorstenDittmann: Fixed 👍🏻

kodumbeats (changes requested):
Do we have any performance data behind this new pagination? I see OR statements in both the SQL and Mongo adapters, so I just want to understand the tradeoff we're making (since `after` pagination scales much better)

src/Database/Adapter/MongoDB.php
```diff
@@ -107,14 +107,15 @@ public function createCollection(string $name, array $attributes = [], array $in
...
+        // Mongo creates an index for _id; _uid, _read and _write index by default
         // Returns the name of the created index as a string.
         // Update $this->getIndexCount when adding another default index
+        $uid = $collection->createIndex(['_uid' => $this->getOrder(Database::ORDER_DESC)], ['name' => '_uid', 'unique' => true]);
         $read = $collection->createIndex(['_read' => $this->getOrder(Database::ORDER_DESC)], ['name' => '_read_permissions']);
         $write = $collection->createIndex(['_write' => $this->getOrder(Database::ORDER_DESC)], ['name' => '_write_permissions']);
 
 
-        if (!$read || !$write) {
+        if (!$read || !$write || $uid) {
```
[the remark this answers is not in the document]
TorstenDittmann: Good catch!

TorstenDittmann:
> Do we have any performance data behind this new pagination? I see OR statements in both the SQL and Mongo adapters, so I just want to understand the tradeoff we're making (since `after` pagination scales much better)

The initial benchmark I did was very much what you can find here: https://shopify.engineering/pagination-relative-cursors

kodumbeats (approved):

PR #41, opened 2021-07-21
Accept array of formats for custom structure validation

kodumbeats:
Most of the required logic for custom validation was [already in place](https://github.com/utopia-php/database/blob/main/src/Database/Validator/Structure.php#L258-L280) in the structure validator, but it was not being used in the Database class.

This PR:
1. Adds a param when creating attributes to store any custom validation on the attribute:
    ```php
    /**
    @param string $format JSON-encoded string of format metadata. Key 'name' required.
    */
    ```
2. Updates the logic for `Structure::addFormat` to accept a callback which creates a validator based on attribute structure. 
3. Throws an exception on `createAttribute` methods when a format is unknown.

The custom validator name is stored as the $format param for attribute creation and is used to use the correct validator for document structure. This way, validators external to this library can be passed down to structure validation.

**Testing**
- [x] TODO

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -397,8 +397,8 @@ class Database
             throw new LimitException('Column limit reached. Cannot create new attribute.');
         }
 
-        if ($format && !Structure::hasFormat($format, $type)) {
-            throw new Exception('Format ("'.$format.'") not available for this attribute type ("'.$type.'")');
+        if ($format && !Structure::hasFormat(json_decode($format, true)['name'], $type)) {
+            throw new Exception('Format ("'.json_decode($format, true)['name'].'") not available for this attribute type ("'.$type.'")');
         }
 
         $collection->setAttribute('attributes', new Document([
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -806,14 +806,14 @@ class Database
      * 
      * @param string $collection
      * @param Document $document
-     * @param array{name: string, validator: Validator, type: string} $formats
+     * @param array{name: string, validator: Validator, type: string} $formats (optional) Array of format validators
      *
      * @return Document
      *
      * @throws AuthorizationException
      * @throws StructureException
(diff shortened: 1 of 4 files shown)


src/Database/Database.php
```diff
@@ -801,14 +805,16 @@ public function getDocument(string $collection, string $id): Document
...
+     * @param array $formats (optional) Array of format validators
+     * {name: string, validator: \Utopia\Validator, type: string}
      *
      * @return Document
      *
      * @throws AuthorizationException
      * @throws StructureException
      */
-    public function createDocument(string $collection, Document $document): Document
+    public function createDocument(string $collection, Document $document, array $formats = []): Document
```
eldadfux, line 817: I don't this this is the best way to add this, having to pass a new list of formats at every operations like can be very very complex. Think of a scenario where we wish to use this formats internally in Appwrite and we have to pass them to route using the dependency injector.

I think we should init the `Structure` validator with this new formats statically once when the application start. Something like: 
```php
Structure::addFormat('email', new Email());
```

Then we can also add a new method to check if a format is know, like:
```php
Structure::hasFormat('email'):bool
```

Using this method we can verify a format is known when setting it as part of a new attribute.

eldadfux (changes requested):

eldadfux:
A few more comments:
1. We can use `hasFormat` to validate the format when creating an attribute
2. I think it might be better to pass a callback to create a validator rather than an instance. This will allow us to pass arguments from the attribute for validators that have some init options. For example, the `Range` validator should get `min` and `max` values from the attribute metadata. We can execute the callback while the `Structure` validator is performing the validation.

For example:
```php
Structure::addFormat('int-range', function($min, $max) {
    return new Range($min, $max, Range::TYPE_INTEGER);
}, Database::VAR_STRING);
```
 $min and $max can be pulled from the attribute object.

PR #42, opened 2021-07-25
Accept 36 chars as keys

kodumbeats:
This PR increases the supported length for Keys and UIDs from 32 chars to 36 chars. With this change, we can support `uuidv4` as valid UIDs.

**Testing**
Tests have been updated to reflect this change

### DIFF
--- a/src/Database/Validator/Key.php
+++ b/src/Database/Validator/Key.php
@@ -9,7 +9,7 @@ class Key extends Validator
     /**
      * @var string
      */
-    protected $message = 'Parameter must contain only letters with no spaces or special chars and be shorter than 32 chars';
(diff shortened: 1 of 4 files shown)


eldadfux (changes requested):
I think the validators description should be a bit more consistent with the one we  use in Appwrite: https://github.com/appwrite/appwrite/pull/1350/files#diff-3c8abdcca7e143e051f729255409460b7fffbb41a3cf8dcc11c57651243662ecR42

christyjacob4 (approved):
Looks good @kodumbeats  :+1:

TorstenDittmann (approved):

PR #44, opened 2021-07-30
Check index limit after appending to index queue

kodumbeats:
The previous logic properly throws `LimitException` on both `createIndex` and `addIndexInQueue` methods, when handled as separate methods. However, when implementing in Appwrite, the error must be properly thrown from `addIndexInQueue` since the Database worker, not `appwrite` itself, calls `createIndex`. 

This PR: 
1. changes the location in code where we check the limit, but no functionality has changed. Note the sign change from `>=` to `>` with the refactored logic.
2. Limit checking methods now accept a `$strict` parameter to only count columns/indexes on table and not in queue. 

**Testing**
- Existing unit tests cover proper exception handling.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -682,17 +682,17 @@ class MariaDB extends Adapter
      * Get current attribute count from collection document
      * 
      * @param Document $collection
-     * @param bool $strict (optional) Only count indexes in collection, ignoring queue count
+     * @param bool $strict (optional) Only count attributes in collection, ignoring queue count
      * @return int
      */
     public function getAttributeCount(Document $collection, bool $strict = false): int
     {
-        $attributes = $collection->getAttribute('attributes') ?? [];
-        $attributesInQueue = $collection->getAttribute('attributesInQueue') ?? [];
+        $attributes = \count($collection->getAttribute('attributes') ?? []);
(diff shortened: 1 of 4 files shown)


TorstenDittmann (approved):

src/Database/Adapter/MariaDB.php
```diff
@@ -655,15 +655,16 @@ public function getSupportForFulltextIndex(): bool
...
+    public function getIndexCount(Document $collection, bool $strict = false): int
     {
-        $indexes = $collection->getAttribute('indexes') ?? [];
-        $indexesInQueue = $collection->getAttribute('indexesInQueue') ?? [];
+        $indexes = \count($collection->getAttribute('indexes') ?? []);
+        $indexesInQueue = ($strict) ? 0 : \count($collection->getAttribute('indexesInQueue') ?? []);
 
         // +3 ==> hardcoded number of default indexes from createCollection
-        return \count($indexes) + \count($indexesInQueue) + 3;
+        return $indexes + $indexesInQueue + 3;
```
TorstenDittmann, line 667: Can we add the hardcoded amount as a static variable? 

So we can re-use it in the tests 👍🏻

TorstenDittmann (approved):

src/Database/Adapter/MariaDB.php
kodumbeats: Totally - though let's do that in a separate PR for ease of review?

PR #45, opened 2021-08-03
fix(database): casting array attribute that is null

TorstenDittmann:


kodumbeats (approved):

TorstenDittmann:
@kodumbeats now that I think about it, this PR would make `null` values and empty array. But they shouldn't actually be the same right?

kodumbeats:
@TorstenDittmann For this logic to trigger, the attribute is already known to be an array of some sort, since `$array === true`. This is a tricky edge case, but I think the behaviour makes sense: the default of an array attribute is an empty array.

@eldadfux What do you think?

eldadfux:
I think this fix should be fine.

PR #48, opened 2021-08-05
Fix: remove unique index from collection name

kodumbeats:
The `Utopia\Database\Database->create()` method created a unique index on collection names. This index has been removed, as the `uniqid` should be the only unique column.

Unanswered questions:
- [x] How to properly unit test this?


eldadfux:
> How to properly unit test this?

You can try to update the collection document and see if it works similar to the error that we have now on the create collection endpoint on Appwrite

PR #49, opened 2021-08-05
Throw exception on case-insensitive duplicate IDs

kodumbeats:
MongoDB document keys are [case sensitive](https://docs.mongodb.com/manual/core/document/#field-names), but MariaDB column names [are not](https://mariadb.com/kb/en/identifier-case-sensitivity/).

This PR adds a case-insensitive check when creating attributes/indexes or adding one to the queue, and will throw `Utopia\Database\Exception\Duplicate` if a match is found. 

This PR also changes the default index on `_uid` for the MongoDB adapter to be case-insensitive.

**Testing**
Tests have been added to catch these exceptions.

**Unanswered questions**
- [x] ~~When adding attributes to a collection, the attribute document is appended to `$attribute->getAttribute('attributes')`, and when fetched, returns `Document[] $attributes`. However, when adding all attributes at once via `createCollection`, I receive just a plain array, not a Document. I had to hack around this problem, and I would like to not have to do this:~~

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -6,6 +6,7 @@ use Exception;
 use Utopia\Database\Validator\Authorization;
 use Utopia\Database\Validator\Structure;
 use Utopia\Database\Exception\Authorization as AuthorizationException;
+use Utopia\Database\Exception\Duplicate as DuplicateException;
 use Utopia\Database\Exception\Limit as LimitException;
 use Utopia\Database\Exception\Structure as StructureException;
 use Utopia\Cache\Cache;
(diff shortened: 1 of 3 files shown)


src/Database/Database.php
```diff
@@ -382,6 +383,20 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+            if ($attribute instanceof Document) {
+                if (\strtolower($attribute->getId()) === \strtolower($id)) {
+                    throw new DuplicateException('Attribute already exists', 400);
+                }
+            } else {
+                if (\strtolower($attribute['$id']) === \strtolower($id)) {
+                    throw new DuplicateException('Attribute already exists', 400);
+                }
+            }
+        });
```
eldadfux, line 398:
```suggestion
        foreach($attributes as $attribute) {
                if (\strtolower($attribute->getId()) === \strtolower($id)) {
                    throw new DuplicateException('Attribute already exists');
                }
            }
        }
```

1. I think the behavior should be consistent now with this fix: https://github.com/utopia-php/database/pull/53/files and we can drop the instance check - please confirm.
2. Using HTTP status codes for errors in this context is useless.
3. Foreach syntax is easier to read, and has better performance: http://www.ktorides.com/2015/02/php-array_walk-vs-foreach/

src/Database/Database.php
```diff
@@ -519,6 +534,23 @@ public function addAttributeInQueue(string $collection, string $id, string $type
     {
         $collection = $this->getCollection($collection);
 
+        /** @var Document[] $attributes */
+        /** @var Document[] $attributesInQueue */
+        $attributes = $collection->getAttribute('attributes');
+        $attributesInQueue = $collection->getAttribute('attributesInQueue');
+
+        \array_walk($attributes, function ($attribute) use ($id) {
```
eldadfux, line 542: Same comments as above.

eldadfux (changes requested):

eldadfux:
I think we need the same implementation for indexes as well if we don't have it yet.

PR #53, opened 2021-08-09
fix(cache): don't save document into cache on update

TorstenDittmann:
Problem:
- when a collection get created, the cache only had the version without an internal ID present

Solution:
- only save a document into cache on `updateDocument` when an internal ID is present

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1122,7 +1122,7 @@ class Database
                     foreach (array_reverse($filters) as $filter) {
                         $node = $this->decodeAttribute($filter, $node);
 
-                        if (array_key_exists('$id', $node)) { // if `$id` exists, create a Document instance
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -1115,17 +1114,21 @@ public function decode(Document $collection, Document $document):Document
...
-            
+
             $value = ($array) ? $value : [$value];
 
             foreach ($value as &$node) {
                 if (($node !== null)) {
                     foreach (array_reverse($filters) as $filter) {
                         $node = $this->decodeAttribute($filter, $node);
+
+                        if (array_key_exists('$id', $node)) { // if `$id` exist, create a Document instance
```
TorstenDittmann, line 1125:
```suggestion
                        if (array_key_exists('$id', $node)) { // if `$id` exists, create a Document instance
```

PR #56, opened 2021-08-11
Fix default attribute index count

kodumbeats:
In https://github.com/utopia-php/database/pull/46, we refactored the default index count into a public static method for clarity and maintainability. However, I overlooked in my review that we don't have a 1-1 for default attributes&indexes - we don't index **$write** permissions since they aren't used in any of our queries. 

This PR:
- Removes the default index on `_write` in the MongoDB adapter
- Adds an additional method `getNumberOfDefaultAttributes()` to adapters.

**Testing**
Current test suite is sufficient to cover this change.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -31,11 +31,6 @@ class MariaDB extends Adapter
         $this->pdo = $pdo;
     }
 
-    public static function getNumberOfDefaultIndexes(): int
-    {
-        return 4;
-    }
-
     /**
      * Create Database
      * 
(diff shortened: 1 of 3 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -31,11 +31,6 @@ public function __construct($pdo)
         $this->pdo = $pdo;
     }
 
-    public static function getNumberOfDefaultIndexes(): int
```
TorstenDittmann, line 34: don't we need this?

TorstenDittmann (approved):

eldadfux (approved):

PR #64, opened 2021-08-20
Allow filters on null values

eldadfux:


src/Database/Database.php
```diff
@@ -1276,17 +1282,19 @@ protected function encodeAttribute(string $name, $value)
      * 
      * @param string $name
      * @param mixed $value
+     * @param Document $document
+     * @param Document $collection
      * 
      * @return mixed
      */
-    protected function decodeAttribute(string $name, $value)
+    protected function decodeAttribute(string $name, $value, Document $document, Document $collection)
```
kodumbeats, line 1290: Now that we've extended this method to allow us to aggregate&cache the `$collection` document, I think this method could use some additional documentation. Without the Appwrite context, I wouldn't know why we're passing the collection and document.

src/Database/Database.php
```diff
@@ -862,7 +858,7 @@ public function getDocument(string $collection, string $id): Document
 
         $document = $this->casting($collection, $document);
         $document = $this->decode($collection, $document);
-
+
```
kodumbeats, line 861:
```suggestion
```

kodumbeats (changes requested):

src/Database/Database.php
eldadfux: Added a description in the doc-block. Removed the collection for now as we don't really use it at this stage.

kodumbeats (approved):
👍🏻

PR #65, opened 2021-08-21
Updated cache keys for better Redis UI expirience

eldadfux:
The new format allows a better UI presentation and better cleaning capabilities using specific patterns. For example:

image


kodumbeats (approved):
👍🏻

PR #73, opened 2021-09-17
feat: get limit methods on database

kodumbeats:
This PR introduces two methods for easy checks of database limits to keep to a DRY approach:
- `getAttributeLimit()`
- `getIndexLimit()`

**Testing**
Tests have been added for the new methods, and the test suite has been updated to utilize them.

**Dependencies**
Forked from https://github.com/utopia-php/database/pull/72. Will convert from draft when parent branch merged.

### DIFF
--- a/tests/Database/Base.php
+++ b/tests/Database/Base.php
@@ -1295,7 +1295,7 @@ abstract class Base extends TestCase
      */
     public function testCheckAttributeCountLimit()
     {
-        if (static::getAdapterName() === 'mariadb' || static::getAdapterName() === 'mysql') {
+        if ($this->getDatabase()->getAttributeLimit() > 0) {
             $collection = static::getDatabase()->getCollection('attributeLimit');
 
             // create same attribute in testExceptionAttributeLimit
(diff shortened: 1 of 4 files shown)


tests/Database/Base.php
```diff
@@ -1273,7 +1268,7 @@ public function testExceptionAttributeLimit()
         if (static::getAdapterName() === 'mariadb' || static::getAdapterName() === 'mysql') {
```
eldadfux, line 1268: Why is this test limited for SQL only? This is a general bad practice it would be better to check for a feature and not for a specific adapter.

kodumbeats: My bad - addressed in fa73afa

tests/Database/Base.php
```diff
@@ -1295,6 +1290,35 @@ public function testExceptionAttributeLimit()
         $this->assertEquals(1,1);
     }
 
+    /**
+     * @depends testExceptionAttributeLimit
+     */
+    public function testCheckAttributeCountLimit()
+    {
+        if (static::getAdapterName() === 'mariadb' || static::getAdapterName() === 'mysql') {
```
kodumbeats, line 1298:
```suggestion
        if ($this->getDatabase()->getAttributeLimit() > 0) {
```

TorstenDittmann (approved):

PR #81, opened 2021-11-02
Construct query validator from array of attribute documents

kodumbeats:
This PR updates the constructor of the Query Validator to accept an array of Utopia\Database\Document objects. This change was made to harmonize with the Queries validator, which depends on this class.

**Testing**
Affected tests have been updated

### DIFF
--- a/src/Database/Validator/QueryValidator.php
+++ b/src/Database/Validator/QueryValidator.php
@@ -3,6 +3,7 @@
 namespace Utopia\Database\Validator;
 
 use Utopia\Validator;
+use Utopia\Database\Document;
 use Utopia\Database\Query;
 
 class QueryValidator extends Validator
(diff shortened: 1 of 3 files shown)


src/Database/Validator/QueryValidator.php
```diff
@@ -34,11 +35,13 @@ class QueryValidator extends Validator
     /**
      * Expression constructor
      *
-     * @param array $schema
+     * @param array $attributes
      */
-    public function __construct($schema)
+    public function __construct($attributes)
```
TorstenDittmann, line 40: Can we add the `array` type here?

kodumbeats: 👍🏻 dd3d5fa

TorstenDittmann (approved):

PR #85, opened 2021-11-17
Feat-auth-improvement

lohanidamodar:
Imrpving authorization to support database cloning and disabling auth only in particular instance

## Changelog
1. Added instance status property
2. Removed static enable, disable and reset method
3. Added instance enable, disable and reset method that modifies the instance status property
4. Removed all use of disable,enable,reset using either instance enable/disable or Authorization skip where applicable

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -121,27 +121,31 @@ class Database
     static protected $filters = [];
 
     /**
-     * @var \Utopia\Database\Validator\Authorization
+     * @var Authorization
      */
-    protected $read;
+    protected Authorization $read;
 
     /**
-     * @var \Utopia\Database\Validator\Authorization
+     * @var Authorization
      */
-    protected $write;
+    protected Authorization $write;
 
     /**
      * Get read authorization
+     * 
+     * @return Authorization
      */
-    public function getRead()
+    public function getRead(): Authorization
     {
         return $this->read;
     }
 
     /**
      * Get write authorization
+     * 
+     * @return Authorization
      */
-    public function getWrite()
+    public function getWrite(): Authorization
     {
         return $this->write;
     }
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -121,6 +121,32 @@ class Database
     static protected $filters = [];
 
     /**
+     * @var Utopia\Database\Validator\Authorization
+     */
+    protected $read;
+
+    /**
+     * @var Utopia\Database\Validator\Authorization
+     */
+    protected $write;
+
+    /**
+     * Get read authorization
+     */
+    public function getRead()
+    {
+        return $this->read;
+    }
+
+    /**
+     * Get write authorization
+     */
+    public function getWrite()
+    {
+        return $this->write;
+    }
+
+    /**
      * @param Adapter $adapter
      * @param Cache $cache
      */
@@ -128,6 +154,8 @@ class Database
     {
         $this->adapter = $adapter;
         $this->cache = $cache;
+        $this->read = new Authorization(self::PERMISSION_READ);
+        $this->write = new Authorization(self::PERMISSION_WRITE);
 
         self::addFilter('json',
             /**
@@ -665,7 +693,7 @@ class Database
         // TODO@kodumbeats Check if returned cache id matches request
         if ($cache = $this->cache->load('cache-'.$this->getNamespace().':'.$collection->getId().':'.$id, self::TTL)) {
             $document = new Document($cache);
-            $validator = new Authorization(self::PERMISSION_READ);
+            $validator = $this->read;
 
             if (!$validator->isValid($document->getRead()) && $collection->getId() !== self::METADATA) { // Check if user has read access to this document
                 return new Document();
@@ -678,7 +706,7 @@ class Database
 
         $document->setAttribute('$collection', $collection->getId());
 
-        $validator = new Authorization(self::PERMISSION_READ);
+        $validator = $this->read;
 
         if (!$validator->isValid($document->getRead()) && $collection->getId() !== self::METADATA) { // Check if user has read access to this document
             return new Document();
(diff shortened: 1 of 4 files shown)


src/Database/Database.php
```diff
@@ -170,6 +198,12 @@ function($value) {
         );
     }
 
+    public function __clone()
```
eldadfux, line 201: I think this may overwrite the original clone behavior. Maybe we should add a call for __parent::_clone(); or something like that at the bottom?

eldadfux:
> Maybe I missed a conversation, but I thought we discussed cloning the entire Database object Appwrite-side and overwriting the Authorization validator so we can customize the Auth object for just a single request lifecycle. How does this refactor help us overwrite Auth behavior?

We also mentioned we can access the read and write auth object and just set them with `->disable()` instead of overwriting them as long as we don't affect the original instances, but only the cloned objects.

src/Database/Database.php
lohanidamodar: https://www.php.net/manual/en/language.oop5.cloning.php
From here, the __clone method is called on newly cloned object once the clone is complete

TorstenDittmann:
Looking at this current implementation right now - I feel like we can achieve a more cleaner solution with not cloning the object at all.

What if we allow the constructor to be passed a Validator class - that we can then extend in Appwrite itself?

lohanidamodar:
> > Maybe I missed a conversation, but I thought we discussed cloning the entire Database object Appwrite-side and overwriting the Authorization validator so we can customize the Auth object for just a single request lifecycle. How does this refactor help us overwrite Auth behavior?
> 
> We also mentioned we can access the read and write auth object and just set them with `->disable()` instead of overwriting them as long as we don't affect the original instances, but only the cloned objects.

Based on your feedback here
https://discord.com/channels/564160730845151244/808657214708908032/910453704560345109
I went ahead to remove the static methods completely, in favor of using Skip where applicable and cloning and using the disable as I'm doing 
https://github.com/utopia-php/database/blob/e319d1d65ee9ed31a5e457ae14124a089be06278/tests/Database/Base.php#L1294

src/Database/Validator/Authorization.php
```diff
@@ -156,41 +165,41 @@ public static function setDefaultStatus(bool $status): void
      */
     public static function skip(callable $callback)
```
TorstenDittmann, line 166: Can we add the `mixed` return type here?

src/Database/Database.php
```diff
@@ -120,6 +120,32 @@ class Database
...
+        return $this->read;
+    }
+
+    /**
+     * Get write authorization
+     */
+    public function getWrite()
+    {
+        return $this->write;
+    }
```
TorstenDittmann, line 147: Please add PHPDocs and return types 🙏🏻

src/Database/Database.php
```diff
@@ -120,6 +120,32 @@ class Database
...
 
+    /**
+     * @var \Utopia\Database\Validator\Authorization
+     */
+    protected $read;
+
+    /**
+     * @var \Utopia\Database\Validator\Authorization
+     */
+    protected $write;
```
TorstenDittmann, line 131: Is it possible to add the Class type to these variables?

src/Database/Validator/Authorization.php
lohanidamodar: adding mixed as return type for function, psalm throws error
image

src/Database/Database.php
```diff
@@ -120,6 +120,36 @@ class Database
...
+     * @var Authorization
+     */
+    protected Authorization $write;
+
+    /**
+     * Get read authorization
+     * 
+     * @return Authorization
+     */
+    public function getRead(): Authorization
```
kodumbeats, line 138: I don't think `$database->getRead()` tells the right story. Maybe:
```suggestion
    public function getReadValidator(): Authorization
```
I think we should be more verbose with these variables/methods, else the context gets lost.

src/Database/Validator/Authorization.php
```diff
@@ -124,6 +128,11 @@ public static function isRole(string $role): bool
      * @var bool
      */
     public static $status = true;
+
+    /**
+     * @var bool
+     */
+    protected $authStatus = true;
```
kodumbeats, line 135: Can we add a todo to replace this with an enum with PHP8.1? Without explanation, I couldn't tell you the difference between `$status`, `$authStatus`, or `$statusDefault`, and enums would help a _lot_.

src/Database/Validator/Authorization.php
kodumbeats: Maybe because the `mixed` type is a PHP8 feature? In PHP7, they'd be in [reserved status](https://wiki.php.net/rfc/reserve_even_more_types_in_php_7#result)

kodumbeats (changes requested):
Small merge conflict too, it seems.

src/Database/Validator/Authorization.php
lohanidamodar: I'm not happy with this name either, but I don't have other better name in mind. How will we name this even with enums any ideas what you are thinking ?

src/Database/Validator/Authorization.php
lohanidamodar: It should be, however we are in PHP 8 right, even then Pslam throws error.

eldadfux: Let's keep a TODO in the code and tackle this later.

src/Database/Validator/Authorization.php
eldadfux: @lohanidamodar lets start by adding doc blocks descriptions, and than re-review the names of the vars. I totally agree at this state they make almost 0 sense.

eldadfux:
> Looking at this current implementation right now - I feel like we can achieve a more cleaner solution with not cloning the object at all.
> 
> What if we allow the constructor to be passed a Validator class - that we can then extend in Appwrite itself?

This solution is almost ready, and is pretty solid. I would suggest completing it for the upcoming release, and continue the discussion on other alternatives based on our availability after the scheduled release.

src/Database/Validator/Authorization.php
TorstenDittmann: Updating Psalm should fix this.

kodumbeats (approved):

eldadfux:
Closing for now due to too many changes done in Database permissions recently.

PR #88, opened 2021-11-23
Feat:  Adds the start of code coverage.

wess:
Beginnings of adding code coverage to projects using PHPUnit. This uses PCOV as the driver and updates the phpunit.xml to include filters for coverage. 

required PCOV:
`$ pecl install pcov`

Then run:
`$ phpunit --coverage-text=tests/coverage/report.txt`

Adapters, in tests, are updated to allow for environment variables for connections, as well as the original as fallback.

### DIFF
--- a/phpunit.xml.bak
+++ b/phpunit.xml.bak
@@ -1,0 +1,21 @@
+<phpunit backupGlobals="false"
+         backupStaticAttributes="false"
+         bootstrap="vendor/autoload.php"
+         colors="true"
+         convertErrorsToExceptions="true"
+         convertNoticesToExceptions="true"
+         convertWarningsToExceptions="true"
+         processIsolation="false"
+         stopOnFailure="false"
+        >
+    <testsuites>
+        <testsuite name="Application Test Suite">
+            <directory>./tests/</directory>
+        </testsuite>
+    </testsuites>
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="true">
+            <directory suffix=".php">./tests/</directory>
+        </whitelist>
+    </filter>
+</phpunit>
--- a/tests/Database/Adapter/MariaDBTest.php
+++ b/tests/Database/Adapter/MariaDBTest.php
@@ -49,7 +49,7 @@ class MariaDBTest extends Base
         }
 
         $dbHost = getenv('MARIA_HOST') ?: 'mariadb';
-        $dbPort = getenv('MARIA_PORT') ?: '3307';
+        $dbPort = getenv('MARIA_PORT') ?: '3306';
         $dbUser = getenv('MARIA_USER') ?: 'root';
         $dbPass = getenv('MARIA_PASS') ?: 'password';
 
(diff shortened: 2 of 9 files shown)


tests/Database/Adapter/MariaDBTest.php
```diff
@@ -48,10 +48,10 @@ static function getDatabase(): Database
             return self::$database;
         }
 
-        $dbHost = 'mariadb';
-        $dbPort = '3306';
-        $dbUser = 'root';
-        $dbPass = 'password';
+        $dbHost = getenv('MARIA_HOST') ?: 'mariadb';
+        $dbPort = getenv('MARIA_PORT') ?: '3307';
```
kodumbeats, line 52: Ah, here it is. Tests are failing for MariaDB because it's available on port **3306** in the `docker-compose.yml` stack. I patched this locally, and then tests run 🟢 
https://github.com/utopia-php/database/blob/75f206d796dc1b839441fceca2d3dd0186a627a4/docker-compose.yml#L37-L45
```suggestion
        $dbPort = getenv('MARIA_PORT') ?: '3306';
```

phpunit.xml.bak
```diff
@@ -0,0 +1,21 @@
+<phpunit backupGlobals="false"
```
eldadfux, line 1: No need for the old version when this PR is good to go.

eldadfux (commented):
Is there a way to generate a code coverage badge for the repo with the total/average score?

christyjacob4 (changes requested):
Is there a reason we are using a dependency for generating a coverage report ? How would it be different from simply using the coverage report that `phpunit` provides out of the box ?

IMO it would be much cleaner to use 
`vendor/bin/phpunit --coverage-clover coverage.xml` and use the [Codecov service](https://github.com/codecov/example-php) that will take care of generating dynamic code coverage README badges and even provides some other features. 

cc: @eldadfux

TorstenDittmann:
> Is there a reason we are using a dependency for generating a coverage report ? How would it be different from simply using the coverage report that `phpunit` provides out of the box ?
> 
> IMO it would be much cleaner to use `vendor/bin/phpunit --coverage-clover coverage.xml` and use the [Codecov service](https://github.com/codecov/example-php) that will take care of generating dynamic code coverage README badges and even provides some other features.
> 
> cc: @eldadfux

Not sure I understand? That is the Code coverage that PHP offers out of the box 😄 

You need some kind of driver that can actually check what code is covered or not => https://phpunit.readthedocs.io/en/9.5/code-coverage-analysis.html

Even the official code cov example does not work because there is no driver for this installed:

Bildschirmfoto 2021-12-10 um 12 10 28

eldadfux:
I think the dependency should be fine and as far as I can recall is required.

tests/coverage/report.txt
```diff
@@ -0,0 +1,48 @@
+
```
eldadfux, line 1: Is this a file of cached results? should we add it to gitignore or we need to have it for the results badge? Can we auto-generate it?

abnegate (commented):
@wess I'd love to get this up to date! If you could address the conflicts and pending feedback we can push it along 😁

PR #91, opened 2021-12-02
skip decoding for null values

lohanidamodar:
When an encrypted field has NULL value in the db, during decoding it fails. This will resolve that by skipping decoding process if the value is null

Details to error that arises without this fix
https://github.com/appwrite/appwrite/runs/4393263762?check_suite_focus=true

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1051,8 +1051,11 @@ class Database
             $filters = $attribute['filters'] ?? [];
             $value = $document->getAttribute($key, null);
             
+            if(is_null($value)) {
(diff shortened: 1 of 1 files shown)


kodumbeats (commented):
Do we have test coverage to confirm this change works the way we expect?

lohanidamodar:
> Do we have test coverage to confirm this change works the way we expect?

Not sure we do, can you help me with what tests we might need?

lohanidamodar:
This seems like should be handled by filter itself, closing this PR

PR #92, opened 2021-12-08
[WIP] feat: abstract SQL with builder pattern

kodumbeats:
Due to inconsistencies in the SQL queries between methods, bugs began to pop up. Since most of our intended adapters support SQL, this PR uses a Builder pattern to reduce the maintenance costs and handle all SQL abstractions in one place across adapters.

TODO:
- [ ] Testing
- [ ] Filter all user input
- [ ] Bind all params
- [ ] Field class to increase DX in lib


TorstenDittmann:
This looks really good already!

PR #96, opened 2021-12-21
Feat: namespace as table prefixes

kodumbeats:
In order to more easily integrate with existing production setups, Appwrite is moving to contain everything under one database. To do this, the database library required two major changes:

- Supporting a default database where everything will be created
- Prepend namespace to table/collection names

This PR introduces the concept of a "default database" that many DBMS support, synonymous to the `USE` SQL instruction.

**HIDDEN GAINS**
I feel more confident that database/collection names are properly escaped in queries, so we may not need to be afraid to support special ASCII characters as names. These changes were made in both the SQL and Mongo adapters.

**TESTING**
Only small changes to `createExistsDelete` tests were required, rest of suite remains unchanged.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -340,16 +340,16 @@
         },
         {
             "name": "utopia-php/framework",
-            "version": "0.19.2",
+            "version": "0.19.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/utopia-php/framework.git",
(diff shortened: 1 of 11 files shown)


TorstenDittmann (approved):
LGTM!

Awesome work, I have created a PR to add some additional tests regarding special chars with collections, attributes and indexes => https://github.com/utopia-php/database/pull/97

TorstenDittmann (approved):

PR #98, opened 2021-12-29
fix: json filter empty json object

TorstenDittmann:


src/Database/Database.php
```diff
@@ -152,6 +152,10 @@ function($value) {
                     return $value;
                 }
 
+                if($value === '{}') {
+                    return new \stdClass();
+                }
+
                 $value = json_decode($value, true);
```
eldadfux, line 159:
```suggestion
                $value = json_decode($value);
```

This might be the issue

PR #100, opened 2022-01-03
fix: metadata integrity on attributes and indexes

TorstenDittmann:
This fixes leftover Metadata when an attribute or index wasn't successfully created.

This change basically creates the attribute/index before creating the metadata.


lohanidamodar (approved):
LGTM!

PR #101, opened 2022-01-12
Feat: throw custom exception when index key is too large

kodumbeats:
This PR exposes the index key limit of each database adapter and throws a Size exception when necessary.

**Testing**
Unit test has been added to catch this exception

**Related PRs:**
https://github.com/appwrite/appwrite/issues/2596

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -412,6 +412,13 @@ abstract class Adapter
     abstract public function getIndexLimit(): int;
 
     /**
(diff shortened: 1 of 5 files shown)


kodumbeats:
**Blocked**: Not sure what to do on this one - InnoDB happily accepts an index of length 3073 when the limit is supposed to be 3072. 

We'll either need to estimate this limit or figure out exactly how it works. I don't think we want to convert strings to bytestreams just for this.

PR #102, opened 2022-01-19
Feat: check if collection exists in database

kodumbeats:
Since moving every collection to a single default database, we lose the ability to know if a collection has been properly configured (e.g. does the metadata table exist). 

This PR extends `Utopia\Database\Database::exists()` to optionally accept a collection name and test for its existence. 

**Notes**
I'm not sold on this, but the only other nice name I could come up with was `collectionExists()` which doesn't fit the pattern. Alternative suggestions welcome.

**Testing**
Unit tests have been added for both success and failure.

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -142,12 +142,14 @@ abstract class Adapter
 
     /**
      * Check if database exists
+     * Optionally check if collection exists in database
      *
-     * @param string $name
+     * @param string $database database name
+     * @param string $collection (optional) collection name
      *
      * @return bool
      */
-    abstract public function exists(string $name): bool;
+    abstract public function exists(string $database, string $collection): bool;
 
     /**
      * List Databases
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -149,7 +149,7 @@ abstract class Adapter
      *
      * @return bool
      */
-    abstract public function exists(string $database, string $collection): bool;
+    abstract public function exists(string $database, string $collection = null): bool;
 
     /**
      * List Databases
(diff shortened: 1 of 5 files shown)


src/Database/Adapter.php
```diff
@@ -142,12 +142,14 @@ abstract public function create(string $name): bool;
...
+     * Optionally check if collection exists in database
      *
-     * @param string $name
+     * @param string $database database name
+     * @param string $collection (optional) collection name
      *
      * @return bool
      */
-    abstract public function exists(string $name): bool;
+    abstract public function exists(string $database, string $collection): bool;
```
eldadfux, line 152:
```suggestion
    abstract public function exists(string $database, string $collection = ''): bool;
```
If this is optional, let's give a default value to `$collection`

src/Database/Adapter/MariaDB.php
```diff
@@ -49,27 +49,46 @@ public function create(string $name): bool
...
+     * Optionally check if collection exists in database
      *
-     * @param string $name
+     * @param string $database database name
+     * @param string $collection (optional) collection name
      *
      * @return bool
      */
-    public function exists(string $name): bool
+    public function exists(string $database, $collection = ''): bool
```
eldadfux, line 59:
```suggestion
    public function exists(string $database, string $collection = ''): bool
```
Missing type

eldadfux (commented):
Very clean. I think the name is OK unless we think of something better. Left a few comments

src/Database/Adapter.php
TorstenDittmann: We should rather default to `null`, that way we can differentiate from an actual empty string and is nullable.

TorstenDittmann (approved):

PR #103, opened 2022-01-25
Escape index id properly

Meldiron:
Before, attempt to create index with id `order` would fail:

```
CREATE INDEX order ON `appwrite`.`_project_dev_collection_posts` ( `title`(255) ASC );
```

```
appwrite-worker-database        | SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'order ON `appwrite`.`_project_dev_collection_posts` ( `title`(255) ASC )' at line 1
```

With this PR, `order` is a valid ID of the key.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -1191,7 +1191,7 @@ class MariaDB extends Adapter
             break;
         }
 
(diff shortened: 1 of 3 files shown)


Meldiron:
The attribute key seems to be escaped properly here: https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L145

eldadfux (approved):

PR #104, opened 2022-01-31
feat: mariadb wildcard search

TorstenDittmann:


eldadfux (commented):
Not sure if we want the API to behave completely different between MongoDB and MariaDB. Would it make more sense if we always add the `*` and abstract that behavior away from developers.

TorstenDittmann:
> Not sure if we want the API to behave completely different between MongoDB and MariaDB. Would it make more sense if we always add the `*` and abstract that behavior away from developers.

MongoDB might be difficult to implement in the first way.

PR #105, opened 2022-01-31
Added $id support for sortAttribute

Meldiron:
Before, people were not able to sort by `$id`, saying `attribute 'id' not found`. Now, people can use `$id` and it is translated to `_uid` under the hood, similar to `queries` with `$id`.

CleanShot 2022-01-31 at 18 41 19

CleanShot 2022-01-31 at 18 41 14


eldadfux (commented):
Changes looks good, but are we blocking attributes with `_` in the Query validators? We need to make sure the same validation we make for attribute name is applied for queries.

eldadfux:
We also have a few failing tests.

PR #106, opened 2022-01-31
Add order attributes validator

Meldiron:
New validator for order queries to force indexes, similar to queries. This also prevents usage of `_uid`.


eldadfux (commented):
We should probably add unit tests

TorstenDittmann (approved):

PR #108, opened 2022-02-07
feat: dedicated permissions table

TorstenDittmann:
With this PR, every Collection will have a dedicated table with all its row level permissions.

The table has 4 columns:
- `_id` internal id (auto incremented)
- `_type` either `read` or `write` (possibility to extend in the future)
- `_permission` the actual permission value
- `_document` reference to the `_uid` for the document

Now that the Permissions are not saved within the actual row of a document anymore, we need to aggregate those informations. For this I am utilizing the [JSON_ARRAYAGG](https://mariadb.com/kb/en/json_arrayagg/) function, which will automatically get all permissions for a Document and outputs it as a JSON array string.

In terms of performance there is very little to no impact and it also allows us to have data integrity over saving the permissions additionally with the document. Also the queries are written in a way to utilize Sub Query cache - which is enabled by default on MariaDB - which optimizes correlated subqueries by storing results together with correlation parameters in a cache and avoiding re-execution of the subquery in cases where the result is already in the cache.

The permissions table is appended with `_perms`, not `_permissions` to save space (we only have 64 chars of length for a table name).

Query for getDocument:
```sql
SELECT
  table_main.*,
  (
    SELECT
      JSON_ARRAYAGG(DISTINCT _permission)
    FROM
      `utopiaTests`.`myapp_620118ce6192e_movies_perms`
    WHERE
      _document = table_main._uid
      AND _type = 'read'
  ) AS _read,
  (
    SELECT
      JSON_ARRAYAGG(DISTINCT _permission)
    FROM
      `utopiaTests`.`myapp_620118ce6192e_movies_perms`
    WHERE
      _document = table_main._uid
      AND _type = 'write'
  ) AS _write
FROM
  `utopiaTests`.`myapp_620118ce6192e_movies` AS table_main
WHERE
  _uid = 'frozen'
LIMIT 1
;
```

Query for find:
```sql
SELECT
  table_main.*,
  (
    SELECT
      JSON_ARRAYAGG(DISTINCT _permission)
    FROM
      `utopiaTests`.`myapp_620118ce6192e_movies_perms`
    WHERE
      _document = table_main._uid
      AND _type = 'read'
  ) AS _read,
  (
    SELECT
      JSON_ARRAYAGG(DISTINCT _permission)
    FROM
      `utopiaTests`.`myapp_620118ce6192e_movies_perms`
    WHERE
      _document = table_main._uid
      AND _type = 'write'
  ) AS _write
FROM
  `utopiaTests`.`myapp_620118ce6192e_movies` AS table_main
WHERE
  table_main._uid IN (
    SELECT
      *
    FROM
      (
        SELECT
          _document
        FROM
          `utopiaTests`.`myapp_620118ce6192e_movies_perms`
        WHERE
          _permission IN ('role:all', 'user2')
          AND _type = 'read'
      ) AS subquery
  );
```

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -18,7 +18,7 @@ class MariaDB extends Adapter
     /**
      * @var PDO
      */
-    protected PDO $pdo;
+    protected $pdo;
 
     /**
      * Constructor.
@@ -27,7 +27,7 @@ class MariaDB extends Adapter
      *
      * @param PDO $pdo
      */
-    public function __construct(PDO $pdo)
+    public function __construct($pdo)
     {
         $this->pdo = $pdo;
     }
@@ -1413,7 +1413,7 @@ class MariaDB extends Adapter
      * Returns the current PDO object
      * @return PDO 
      */
-    protected function getPDO(): PDO
+    protected function getPDO()
     {
         return $this->pdo;
     }
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -247,6 +247,27 @@ class MariaDB extends Adapter
     }
 
     /**
+     * Rename Attribute
+     *
+     * @param string $collection
+     * @param string $id
+     * @param string $name
+     * @return bool
+     * @throws Exception
+     * @throws PDOException
+     */
+    public function renameAttribute(string $collection, string $id, string $name): bool
+    {
+        $collection = $this->filter($collection);
+        $id = $this->filter($id);
+        $name = $this->filter($name);
+
+        return $this->getPDO()
+            ->prepare("ALTER TABLE `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$collection}` RENAME COLUMN `{$id}` TO `{$name}`;")
+            ->execute();
+    }
+
+    /**
      * Delete Attribute
      *
      * @param string $collection
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -34,10 +34,11 @@ class MariaDB extends Adapter
 
     /**
      * Create Database
-     * @param string $name 
-     * @return bool 
-     * @throws Exception 
-     * @throws PDOException 
+     *
+     * @param string $name
+     * @return bool
+     * @throws Exception
+     * @throws PDOException
      */
     public function create(string $name): bool
     {
@@ -151,12 +152,9 @@ class MariaDB extends Adapter
 
             $indexAttributes = $index->getAttribute('attributes');
             foreach ($indexAttributes as $nested => $attribute) {
-                $indexLength = $index->getAttribute('lengths')[$nested] ?? '';
-                if (!empty($indexLength)) {
-                    $indexLength =  '(' . (int) $indexLength . ')';
-                }
-
-                $indexOrder = $index->getAttribute('orders')[$nested] ?? '';
+                $indexLength = $index->getAttribute('lengths')[$key] ?? '';
+                $indexLength = (empty($indexLength)) ? '' : '('.(int)$indexLength.')';
+                $indexOrder = $index->getAttribute('orders')[$key] ?? '';
                 $indexAttribute = $this->filter($attribute);
 
                 if ($indexType === Database::INDEX_FULLTEXT) {
@@ -182,7 +180,7 @@ class MariaDB extends Adapter
                 ->execute();
 
             $this->getPDO()
-                ->prepare("CREATE TABLE IF NOT EXISTS `{$database}`.`{$namespace}_{$id}+permissions` (
+                ->prepare("CREATE TABLE IF NOT EXISTS `{$database}`.`{$namespace}_{$id}\$permissions` (
                     `_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
                     `_type` VARCHAR(12) NOT NULL,
                     `_permission` VARCHAR(255) NOT NULL,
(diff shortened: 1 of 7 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -397,14 +433,31 @@ public function createDocument(string $collection, Document $document): Document
             $stmt->bindValue(':' . $attribute, $value, $this->getPDOType($value));
         }
 
+        $permissions = [];
+        foreach ($document->getRead() as $permission) {
+            $permissions[] = "('read', '{$permission}', '{$document->getId()}')";
```
eldadfux, line 438: Do we want to filter the id we add here for consistency?

src/Database/Adapter/MariaDB.php
```diff
@@ -397,14 +433,31 @@ public function createDocument(string $collection, Document $document): Document
...
+        }
+
+        foreach ($document->getWrite() as $permission) {
+            $permissions[] = "('write', '{$permission}', '{$document->getId()}')";
```
eldadfux, line 442: Do we want to filter the id we add here for consistency?

src/Database/Adapter/MariaDB.php
```diff
@@ -425,35 +478,117 @@ public function createDocument(string $collection, Document $document): Document
...
         $name = $this->filter($collection);
         $columns = '';
 
+        /**
+         * Get current permissions from the database
+         */
+        $permissionsStmt = $this->getPDO()->prepare("
+                SELECT _type, _permission
+                FROM `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}_perms` p
+                WHERE p._document = {$this->getPDO()->quote($document->getId())}
```
eldadfux, line 498: Why don't we use the prepared statement to bind this value? `->quote()` should be avoided when possible.

src/Database/Adapter/MariaDB.php
```diff
@@ -425,35 +478,117 @@ public function createDocument(string $collection, Document $document): Document
...
+
+        /**
+         * Query to remove permissions
+         */
+        if (!empty($readRemoved) || !empty($writeRemoved)) {
+            $removeQuery = 'AND (';
+            if ($readRemoved) {
+                $removeQuery .= "(
+                    _type = 'read'
+                    AND _permission IN (" . implode(', ', array_map(fn (string $role) => $this->getPDO()->quote($role), $readRemoved)) . ")
```
eldadfux, line 532: Would it make sense to use the prepared values instead of concatenating the query string? Might be safer and more consistent.

eldadfux: Probably same comment to all uses of `quote()`.

src/Database/Adapter/MariaDB.php
```diff
@@ -492,27 +633,31 @@ public function updateDocument(string $collection, Document $document): Document
...
-
+        $stmt = $this->getPDO()->prepare("DELETE FROM `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}` WHERE _uid = :_uid LIMIT 1");
         $stmt->bindValue(':_uid', $id, PDO::PARAM_STR);
 
-        if(!$stmt->execute()) {
+        $stmtPermissions = $this->getPDO()->prepare("DELETE FROM `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}_perms` WHERE _document = :_uid");
+        $stmtPermissions->bindValue(':_uid', $id, PDO::PARAM_STR);
+
+        try {
+            $stmt->execute() || throw new Exception('Failed to delete document');
```
eldadfux, line 653: Transaction could be useful here.

src/Database/Adapter/MariaDB.php
```diff
@@ -640,21 +789,22 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
...
-            unset($value['_uid']);
-            unset($value['_id']);
-            unset($value['_read']);
-            unset($value['_write']);
-
-            $value = new Document($value);
+        foreach ($results as $key => $value) {
+            $results[$key]['$id'] = $value['_uid'];
+            $results[$key]['$internalId'] = $value['_id'];
+            $results[$key]['$read'] = json_decode($value['_read'], true) ?? [];
```
eldadfux, line 795: Have we ended up storing the `read` and `write` in both tables?

src/Database/Adapter/MariaDB.php
```diff
@@ -720,52 +877,55 @@ public function count(string $collection, array $queries = [], int $max = 0): in
...
      * @param string $attribute
      * @param Query[] $queries
      * @param int $max
-     *
      * @return int|float
+     * @throws Exception
+     * @throws PDOException
      */
-    public function sum(string $collection, string $attribute, array $queries = [], int $max = 0)
+    public function sum(string $collection, string $attribute, array $queries = [], int $max = 0): int|float
```
eldadfux, line 888: In what case are we returning float? can we add it to the docblock?

src/Database/Adapter/MariaDB.php
TorstenDittmann: No, but we aggregate a JSON array from the permissions table - so we just need to parse it 👍

src/Database/Adapter/MariaDB.php
TorstenDittmann: Well, when we get the sum of a float attribute.

src/Database/Adapter/MariaDB.php
TorstenDittmann: At this point they have already been through the filter 👍🏻

src/Database/Adapter/MySQL.php
```diff
@@ -103,21 +103,22 @@ protected function getSQLIndex(string $collection, string $id, string $type, arr
...
+     * Get SQL query to aggregate permissions as JSON array
+     *
+     * @param string $collection 
+     * @param string $type 
+     * @param string $alias 
+     * @return string 
+     * @throws Exception 
      */
-    protected function getSQLPermissions(array $roles): string
+    protected function getSQLPermissionsQuery(string $collection, string $type, string $alias): string
```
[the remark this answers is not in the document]
TorstenDittmann: The behaviour is different for both. MySQL does distinct by default - MariaDB doesn't.

Mysql doesn't allow passing Distinct.

PR #113, opened 2022-03-15
fix: mariadb search with @ char

TorstenDittmann:
- allow using MariaDB BOOLEAN search keywords and replace them with space

https://github.com/appwrite/appwrite/issues/2394

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -1267,7 +1267,16 @@ class MariaDB extends Adapter
     {
         switch ($operator) {
             case Query::TYPE_SEARCH:
-                $value = "\"{$value}*\”";
+                /**
+                 * Replace reservec chars with space.
+                 */
+                $value = str_replace(['@'], ' ', $value);
+
+                /**
+                 * Prepend wildcard by default on the back.
+                 */
+                $value = "'{$value}*'";
+
                 return 'MATCH(' . $attribute . ') AGAINST(' . $this->getPDO()->quote($value) . ' IN BOOLEAN MODE)';
                 break;
 
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -1270,7 +1270,7 @@ class MariaDB extends Adapter
                 /**
                  * Replace reservec chars with space.
                  */
-                $value = str_replace(['@'], ' ', $value);
+                $value = str_replace(['@', '+', '-', '*'], ' ', $value);
 
                 /**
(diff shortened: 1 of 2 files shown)


tests/Database/Base.php
```diff
@@ -847,6 +847,15 @@ public function testFind(Document $document)
             ]);
 
             $this->assertEquals(2, count($documents));
+
+            /**
+             * Allow reserved keywords for search
+             */
+            $documents = static::getDatabase()->find('movies', [
+                new Query('name', Query::TYPE_SEARCH, ['*test+alias@email-provider.com']),
```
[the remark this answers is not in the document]
TorstenDittmann: I can add this 👍🏻

src/Database/Adapter/MariaDB.php
```diff
@@ -1267,7 +1267,16 @@ protected function getSQLCondition(string $attribute, string $operator, string $
     {
         switch ($operator) {
             case Query::TYPE_SEARCH:
-                $value = "\"{$value}*\”";
+                /**
+                 * Replace reservec chars with space.
+                 */
+                $value = str_replace(['@', '+', '-', '*'], ' ', $value);
```
stnguyen90, line 1273: Is there any way to preserve the search? If the @ is removed, there's no way to search for a specific email.

TorstenDittmann: Internally Mariadb already removes that sign. You can search for `"torsten@appwrite.io"`.

The quotes will make the word coherent and it should give you the proper results.

But there still are edge cases where this will not work a 100%.

Because the above search will also show you `other.torsten.appwrite.io`.

To fully fix this we would need to introduce the queries from the database service to user service.

stnguyen90: In addition, I think `+` is useful to do a "starts with" search

TorstenDittmann: True, but looking at the bigger picture for the future - we need to support the same syntax in all Adapters. This behavior is different for Mongo/Postgres & co.

stnguyen90: How will Appwrite support wildcard search? I believe the biggest driver for switching from natural language mode to boolean mode was to allow for wildcard search: https://github.com/appwrite/appwrite/issues/2676

stnguyen90 (approved):

tests/Database/Base.php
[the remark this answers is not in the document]
TorstenDittmann: Added them 👍🏻

christyjacob4 (approved):

PR #117, opened 2022-03-30
Feat: Cache Upgrade

Meldiron:
Upgrade cache library from 0.4 to 0.5. This is required if people want to use new 0.5 version of Cache alongside Utopia database

### DIFF
--- a/bin/tasks/index.php
+++ b/bin/tasks/index.php
@@ -26,7 +26,6 @@ $cli
     ->param('adapter', '', new Text(0), 'Database adapter', false)
     ->param('name', '', new Text(0), 'Name of created database.', false)
     ->action(function ($adapter, $name) {
-        Cache::setCaseSensitivity(false);
         $database = null;
 
         switch ($adapter) {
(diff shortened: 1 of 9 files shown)


bin/tasks/index.php
```diff
@@ -26,6 +26,7 @@
     ->param('adapter', '', new Text(0), 'Database adapter', false)
     ->param('name', '', new Text(0), 'Name of created database.', false)
     ->action(function ($adapter, $name) {
+        Cache::setCaseSensitivity(false);
```
christyjacob4, line 29: Question for you , if false is meant to be used in most cases, why not use `false` as default ? 🤔  and avoid all these changes?

Meldiron: To prevent breaking change in cache library, mainly.

christyjacob4: How would it be a breaking change in utopia-php/cache if we change this line to false ? 
https://github.com/utopia-php/cache/blob/02ffbe165f9632e05519eb75a2ce32deb19638a0/src/Cache/Cache.php#L15

Meldiron: New cache version now defaults to false

christyjacob4 (approved):

PR #118, opened 2022-03-31
Feat: Disallow failed and pending indexes

Meldiron:
- [x] Verify index status in filter query
- [x] Verify index status in sort query
- [x] Implement filter validator test
- [x] Implement sort validator test
- [x] Manual QA:

CleanShot 2022-03-31 at 12 03 14
CleanShot 2022-03-31 at 12 02 46

### DIFF
--- a/src/Database/Validator/OrderAttributes.php
+++ b/src/Database/Validator/OrderAttributes.php
@@ -55,7 +55,9 @@ class OrderAttributes extends Validator
         ];
 
         foreach ($indexes as $index) {
-            $this->indexes[] = $index->getArrayCopy(['attributes', 'type']);
+            if($index->getAttribute('status') === 'available') {
+                $this->indexes[] = $index->getArrayCopy(['attributes', 'type']);
+            }
         }
 
         $this->strict = $strict;
--- a/src/Database/Validator/Queries.php
+++ b/src/Database/Validator/Queries.php
@@ -45,9 +45,11 @@ class Queries extends Validator
             'type' => Database::INDEX_UNIQUE,
             'attributes' => ['$id']
         ];
-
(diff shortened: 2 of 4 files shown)


src/Database/Validator/OrderAttributes.php
```diff
@@ -55,7 +55,9 @@ public function __construct($attributes, $indexes, $strict = true)
         ];
 
         foreach ($indexes as $index) {
-            $this->indexes[] = $index->getArrayCopy(['attributes', 'type']);
+            if($index->getAttribute('status') === 'available') {
```
christyjacob4, line 58: Only concern is the usage of `available` here 
the status of attributes `processing, available, failed, deleting, stuck` are appwrite specific concepts and we're coupling the library with Appwrite by doing this.

What happens if we decide to change `available` to `ready` in Appwrite. this means we have to release a new library version as well. What about other developers using utopia/database ? Will they be forced to fix their code because we changed a attribute name in Appwrite ? 

So we should find a better way to do this. Defining `status` values as constants can be a good approach. 

Would like a better discussion on this :)

christyjacob4 (changes requested):
I think this implementation needs more work because of the concern I had.

src/Database/Validator/OrderAttributes.php
christyjacob4: Another solution is to leave this validator as it is and override it in Appwrite. 
So the Appwrite implementation of OrderAttrbutes can have this specific implementation :)

Meldiron:
Closing issue. This is appwrite-specific behaviour and should be solved by appwrite-specific validator class (in Appwrite repo)

PR #121, opened 2022-04-06
Feat: renameCollection [BREAKING]

Meldiron:
New method: `renameCollection(string $id, string $newId)`
Updated adapter signature, adding `documentId`: `updateDocument(string $collection, string $documentId, Document $document)`

Change to `updateDocument` was required as we now need this method to also be able to change value of _uid. To do that properly, you need both IDs, current one (`documentId`), and new one (`docuent->getId()`)

- [x] Implement adapter query

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -247,6 +247,32 @@ class MariaDB extends Adapter
     }
 
     /**
+     * Rename Collection
+     *
+     * @param string $id
+     * @param string $name
+     * @return bool
+     * @throws Exception
+     * @throws PDOException
+     */
+    public function renameCollection(string $id, string $name): bool
+    {
+        $id = $this->filter($id);
+        $name = $this->filter($name);
+
+        $responses = [];
+        $responses[] = $this->getPDO()
+            ->prepare("ALTER TABLE `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$id}` RENAME TO `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}`;")
+            ->execute();
+
+        $responses[] = $this->getPDO()
+            ->prepare("ALTER TABLE `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$id}_perms` RENAME TO `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}_perms`;")
+            ->execute();
+
+        return \array_search(false, $responses) === false ? true : false;
+    }
+
+    /**
      * Rename Attribute
      *
      * @param string $collection
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -477,13 +477,14 @@ class MariaDB extends Adapter
      * Update Document
      *
      * @param string $collection
+     * @param string $documentId
      * @param Document $document
      * @return Document
      * @throws Exception
      * @throws PDOException
      * @throws Duplicate
      */
-    public function updateDocument(string $collection, Document $document): Document
+    public function updateDocument(string $collection, string $documentId, Document $document): Document
     {
         $attributes = $document->getAttributes();
         $name = $this->filter($collection);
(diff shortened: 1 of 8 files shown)


src/Database/Database.php
```diff
@@ -366,6 +366,50 @@ public function createCollection(string $id, array $attributes = [], array $inde
...
+     */
+    public function renameCollection(string $id, string $newId): Document
+    {
+        // Check current ID
+        $metaDocument = $this->findOne(self::METADATA, [
+            new Query('name', Query::TYPE_EQUAL, [$id])
+        ]);
+
+        if(!$metaDocument) {
+            throw new DuplicateException('Collection not found');
```
Meldiron, line 386: Better exception name here

tests/Database/Base.php
```diff
@@ -1930,4 +1930,57 @@ public function testGetId()
...
+        $this->assertEquals('planets3', $planets->getAttribute('name'));
+        $this->assertEquals('planets2', $planets2->getId());
+        $this->assertEquals('planets2', $planets2->getAttribute('name'));
+    }
+
+    /**
+     * @depends testRenameCollection
+     * @expectedException Exception
+     */
+    public function textRenameMissing()
```
Meldiron, line 1969:
```suggestion
    public function textRenameCollectionMissing()
```

tests/Database/Base.php
```diff
@@ -1930,4 +1930,57 @@ public function testGetId()
...
+        $database = static::getDatabase();
+        $this->expectExceptionMessage('Collection not found');
+        $database->renameCollection('planets', 'planets4');
+    }
+
+    /**
+     * @depends testRenameCollection
+     * @expectedException Exception
+     */
+    public function testRenameExisting()
```
Meldiron, line 1980:
```suggestion
    public function testRenameCollectionExisting()
```

Meldiron:
Mongo adapter failing.

Meldiron:

src/Database/Adapter/MariaDB.php
```diff
@@ -477,13 +503,14 @@ public function createDocument(string $collection, Document $document): Document
...
      * @param string $collection
+     * @param string $documentId
      * @param Document $document
      * @return Document
      * @throws Exception
      * @throws PDOException
      * @throws Duplicate
      */
-    public function updateDocument(string $collection, Document $document): Document
+    public function updateDocument(string $collection, string $documentId, Document $document): Document
```
TorstenDittmann, line 513: It might not be too obvious, what the documentId is actually doing here.

Also, it can easily be used in the wrong way. So it might make more sense to have it as its own function? 🤔

src/Database/Database.php
```diff
@@ -366,6 +366,48 @@ public function createCollection(string $id, array $attributes = [], array $inde
...
+        }
+
+        // Check new ID
+        $newMetaDocument = $this->findOne(self::METADATA, [
+            new Query('name', Query::TYPE_EQUAL, [$name])
+        ]);
+
+        if($newMetaDocument) {
+            throw new DuplicateException('Collection ID already used');
+        }
```
TorstenDittmann, line 395: Not sure this is needed, it might make more sense to expect the exception for this on `renameCollection`

src/Database/Adapter/MariaDB.php
Meldiron: I agree. Ill create method to `renameDocument`.

src/Database/Database.php
Meldiron: We are inside renameCollection. Not sure what you mean.

TorstenDittmann: I meant the `renameCollection` from the adapter 👍🏻

abnegate:
@Meldiron Is this still relevant or can we close it?

Meldiron:
@abnegate We didnt need this yet, but might be cool to have in future for migrations. We could keep this as draft?
It requires some breaking changes to existing logic which is a bit scarry. We should try to avoid it unless nessessary.

abnegate:
@Meldiron Let's close for now to keep the number of PRs down, the closed PR and branch will still be there if we need to pick it up again

PR #122, opened 2022-04-06
Feat: Improve renameAttribute

Meldiron:
- Implemented tests
- Improved PHPDocs
- Added new exceptions

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -615,12 +615,18 @@ class Database
         $attributes = $collection->getAttribute('attributes', []);
         $indexes = $collection->getAttribute('indexes', []);
 
-        $attribute = \in_array($old, \array_column($attributes, '$id'));
+        $attribute = \in_array($old, \array_map(function($attribute) {
+            return $attribute['$id'];
+        }, $attributes));
+
         if($attribute === false) {
             throw new Exception('Attribute not found');
         }
 
-        $attributeNew = \in_array($new, \array_column($attributes, '$id'));
+        $attributeNew = \in_array($new, \array_map(function($attribute) {
+            return $attribute['$id'];
+        }, $attributes));
+        
         if($attributeNew !== false) {
             throw new DuplicateException('Attribute name already used');
         }
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -604,33 +604,31 @@ class Database
      * Rename Attribute
      *
      * @param string $collection
-     * @param string $id
-     * @param string $name
+     * @param string $old Current attribute ID
+     * @param string $name New attribute ID
      *
      * @return bool
      */
-    public function renameAttribute(string $collection, string $id, string $name): bool
+    public function renameAttribute(string $collection, string $old, string $new): bool
     {
         $collection = $this->getCollection($collection);
         $attributes = $collection->getAttribute('attributes', []);
         $indexes = $collection->getAttribute('indexes', []);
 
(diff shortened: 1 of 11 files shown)


Meldiron:
- [ ] you might have to update the metadata 👍 since in the indexes - the attribute will still be with the old name

Meldiron:
Mongo adapter failing.

Meldiron:

src/Database/Database.php
```diff
@@ -601,18 +601,35 @@ public function deleteAttribute(string $collection, string $id): bool
...
     public function renameAttribute(string $collection, string $id, string $name): bool
     {
         $collection = $this->getCollection($collection);
-
         $attributes = $collection->getAttribute('attributes', []);
+        $indexes = $collection->getAttribute('indexes', []);
+
+        $attribute = \array_search($id, \array_map(function($attribute) {
+            return $attribute['$id'];
+        }, $attributes));
```
TorstenDittmann, line 620: Pretty sure this will iterate twice over the array.

TorstenDittmann: You could use `array_column($attributes, '$id')` - which will not iterate like map does.

src/Database/Database.php
```diff
@@ -601,18 +601,35 @@ public function deleteAttribute(string $collection, string $id): bool
...
+
+        if($attribute === false) {
+            throw new Exception('Attribute not found');
+        }
+
+        $attributeNew = \array_search($name, \array_map(function($attribute) {
+            return $attribute['$id'];
+        }, $attributes));
```
TorstenDittmann, line 628: same comment as above.

src/Database/Database.php
```diff
@@ -622,7 +639,21 @@ public function renameAttribute(string $collection, string $id, string $name): b
...
 
+        foreach ($indexes as $index) {
+            $indexAttributes = $index->getAttribute('attributes', []);
+            $indexAttributes = \array_map(function($attribute) use($id, $name) {
+                if($attribute === $id) {
+                    return $name;
+                }
+
+                return $attribute;
+            }, $indexAttributes);
```
TorstenDittmann, line 650:
```suggestion
            $indexAttributes = \array_map(fn($attribute) => (attribute === $id) ? $name : $attribute , $indexAttributes);
```

Arrow functions allow you to avoid the `use` keyword.

src/Database/Database.php
Meldiron: Updated

src/Database/Database.php
Meldiron: Updated

Meldiron:
Additionally, I switched from `array_search` to `in_array`, suggested on different PR doing a similar thing.

src/Database/Database.php
```diff
@@ -601,34 +601,60 @@ public function deleteAttribute(string $collection, string $id): bool
...
+    public function renameAttribute(string $collection, string $old, string $new): bool
     {
         $collection = $this->getCollection($collection);
-
         $attributes = $collection->getAttribute('attributes', []);
+        $indexes = $collection->getAttribute('indexes', []);
+
+        $attribute = \in_array($old, \array_map(function($attribute) {
+            return $attribute['$id'];
+        }, $attributes));
```
TorstenDittmann, line 620: Just a minor thing, imo arrow functions are less noisy here.
```suggestion
        $attribute = \in_array($old, \array_map(fn($attribute) => $attribute['$id'], $attributes));
```

TorstenDittmann (approved):

PR #123, opened 2022-04-06
Feat: renameIndex method

Meldiron:
Adds `renameIndex` method

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -632,6 +632,38 @@ class Database
     }
 
     /**
+     * Rename Index
+     *
+     * @param string $collection
+     * @param string $id
+     * @param string $name
+     *
+     * @return bool
+     */
+    public function renameIndex(string $collection, string $id, string $name): bool
+    {
+        $collection = $this->getCollection($collection);
+
+        $indexes = $collection->getAttribute('indexes', []);
+
+        foreach ($indexes as $key => $value) {
+            if (isset($value['$id']) && $value['$id'] === $id) {
+                $indexes[$key]['key'] = $name;
+                $indexes[$key]['$id'] = $name;
+                break;
+            }
+        }
+
+        $collection->setAttribute('indexes', $indexes);
+
+        if ($collection->getId() !== self::METADATA) {
+            $this->updateDocument(self::METADATA, $collection->getId(), $collection);
+        }
+
+        return $this->adapter->renameIndex($collection->getId(), $id, $name);
+    }
+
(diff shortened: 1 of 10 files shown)


Meldiron:

src/Database/Database.php
```diff
@@ -631,6 +631,54 @@ public function renameAttribute(string $collection, string $id, string $name): b
...
+     */
+    public function renameIndex(string $collection, string $id, string $name): bool
+    {
+        $collection = $this->getCollection($collection);
+
+        $indexes = $collection->getAttribute('indexes', []);
+
+        $index = \array_search($id, \array_map(function($index) {
+            return $index['$id'];
+        }, $indexes));
```
TorstenDittmann, line 651: Instead of `array_map`, you can use `array_columns` - which will not do a iteration.

TorstenDittmann: and if you don't need the value of `$index`, you can use `in_array` - then you don't need to iterate once.

src/Database/Database.php
```diff
@@ -631,6 +631,54 @@ public function renameAttribute(string $collection, string $id, string $name): b
...
+
+        if($index === false) {
+            throw new Exception('Index not found');
+        }
+
+        $indexNew = \array_search($name, \array_map(function($index) {
+            return $index['$id'];
+        }, $indexes));
```
TorstenDittmann, line 659: same comment as above.

src/Database/Database.php
Meldiron: Done both, in both cases.

src/Database/Database.php
Meldiron: Done both, in both cases.

TorstenDittmann (approved):

src/Database/Database.php
```diff
@@ -631,6 +631,54 @@ public function renameAttribute(string $collection, string $id, string $name): b
...
+     */
+    public function renameIndex(string $collection, string $old, string $new): bool
+    {
+        $collection = $this->getCollection($collection);
+
+        $indexes = $collection->getAttribute('indexes', []);
+
+        $index = \in_array($old, \array_map(function($index) {
+            return $index['$id'];
+        }, $indexes));
```
TorstenDittmann, line 651: I think arrow functions are less noisy 👀

Meldiron: Updated.

TorstenDittmann (approved):

PR #124, opened 2022-04-07
Feat: Implement updateAttribute()

Meldiron:


Meldiron:
Mongo tests are expected to fail, implementation missing.

Meldiron:
Letting the database natively rebuild index (and decide when it needs to do it) is a better solution because:
- If it fails, the whole thing fails. You don't need to rollback anything
- The index regeneration SHOULD behave the same as CREATE INDEX, so there isn't any advantage or disadvantage

src/Database/Database.php
```diff
@@ -412,6 +412,150 @@ public function deleteCollection(string $id): bool
...
+        // Can get to false during function
+        $success = true;
+
+        $collection = $this->getCollection($collection);
+
+        $attributes = $collection->getAttribute('attributes', []);
+
+        $attributeIndex = \array_search($id, \array_map(function($attribute) {
+            return $attribute['$id'];
+        }, $attributes));
```
TorstenDittmann, line 449: see my other PR reviews regarding this.

src/Database/Database.php
```diff
@@ -412,6 +412,150 @@ public function deleteCollection(string $id): bool
...
+            throw new Exception('Attribute not found');
+        }
+
+        $attribute = $attributes[$attributeIndex];
+
+        if($type !== null || $size !== null || $signed !== null || $array !== null) {
+            $type = $type === null ? $attribute->getAttribute('type') : $type;
+            $size = $size === null ? $attribute->getAttribute('size') : $size;
+            $signed = $signed === null ? $attribute->getAttribute('signed') : $signed;
+            $array = $array === null ? $attribute->getAttribute('array') : $array;
```
TorstenDittmann, line 461:
```suggestion
            $type ??= $attribute->getAttribute('type');
            $size ??= $attribute->getAttribute('size');
            $signed ??= $attribute->getAttribute('signed');
            $array ??= $attribute->getAttribute('array');
```

src/Database/Database.php
```diff
@@ -412,6 +412,150 @@ public function deleteCollection(string $id): bool
...
+                    break;
+                default:
+                    throw new Exception('Unknown attribute type: ' . $type);
+                    break;
+            }
+
+            $attribute->setAttribute('type', $type);
+            $attribute->setAttribute('size', $size);
+            $attribute->setAttribute('signed', $signed);
+            $attribute->setAttribute('array', $array);
```
TorstenDittmann, line 487:
```suggestion
            $attribute
                ->setAttribute('type', $type)
                ->setAttribute('size', $size)
                ->setAttribute('signed', $signed)
                ->setAttribute('array', $array);
```

this can be chained.

TorstenDittmann (changes requested):

TorstenDittmann:
Looking at this, it might have been smarter to implement more smaller functions instead of a universal one 🤔

src/Database/Database.php
Meldiron: Updated. Cant use in_array here, as I need index later on.

src/Database/Database.php
Meldiron: Applied

src/Database/Database.php
Meldiron: Applied

Meldiron:
> Looking at this, it might have been smarter to implement more smaller functions instead of a universal one 🤔

Yes, there is also an edge case when index rebuild might be triggered, and query results would depend on how strict MariaDB settings are. Before making this user-faced in Appwrite, we need bigger research (what queries cost more) and separate index-rebuilding-logic from the one that doesn't. For now, for migration purposes, this should be quick and simple solution.

Do you want me to work on improving this PR?

src/Database/Database.php
```diff
@@ -412,6 +412,151 @@ public function deleteCollection(string $id): bool
...
+                    break;
+            }
+
+            $attribute
+                ->setAttribute('type', $type)
+                ->setAttribute('size', $size)
+                ->setAttribute('signed', $signed)
+                ->setAttribute('array', $array);
+
+            $success = $this->adapter->updateAttribute($collection->getId(), $id, $type, $size, $signed, $array);
```
TorstenDittmann, line 490: Don't we want to throw an exception here if something fails?

tests/Database/Base.php
```diff
@@ -1930,4 +1931,244 @@ public function testGetId()
...
+            null, // type
+            10, // size
+            false, // required
+            "Unnamed", // default
+            null, // signed
+            null, // array
+            null, // format
+            null, // formatOptions
+            null, // filters
+        ));
```
TorstenDittmann, line 1995: For readability we can use named arguments here:
```suggestion
        $this->assertTrue($database->updateAttribute('stocks', 'brand',
            size: 10,
            required: false,
            default: "Unnamed"
        ));
```

src/Database/Database.php
Meldiron: It will throw automatically, providing a proper message of what happened. For instance `Truncated incorrect INTEGER value: "Matej"`

TorstenDittmann: So, why do we need to keep success?

Let's say it returns `false` -> indicating that something went wrong. We are still executing all of its remaining code.

Meldiron: Hmm, I see the confusion now. First, I added `return $success` as it was missing.

The boolean comes from `PDO->execute()`. Not sure why it returns a boolean but also throws an exception. Looking at other adapter methods, they all return `execute()`, so they return a boolean.

TorstenDittmann: Now we are returning it too early 👍🏻 

The PDO has a configuration, where we can set either throw an exception or return false/true. Let's check if its false here and return false. If not just continue. At the bottom we can just return `true`. That way we make sure, we do actions only when previous steps are successful.

Meldiron:
Closing PR. Problem addressed in separate PR that makes more developer-friendly interfaces: https://github.com/utopia-php/database/pull/132

Most of the code copied from this PR.

PR #129, opened 2022-04-18
Fix structure validation for creating new attributes in a collection:

fogelito:
Fix Error:
"Invalid document structure: Attribute "$internalId" has invalid type. Value must be a valid string."

When we use the funtion getDocument:
$document['$internalId'] = $document['_id'];
The value is the autoincrement primary key which is an Integer not a string.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,7 +56,7 @@ services:
       MONGO_INITDB_ROOT_PASSWORD: example
 
   mysql:
-    image: mysql:8.0.25
+    image: mysql:8.0.28-oracle
     container_name: utopia-mysql
     networks:
       - database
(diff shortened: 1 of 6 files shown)


fogelito:
Screen Shot 2022-04-18 at 12 01 19
Screen Shot 2022-04-18 at 12 03 08

docker-compose.yml
```diff
@@ -56,7 +56,7 @@ services:
       MONGO_INITDB_ROOT_PASSWORD: example
 
   mysql:
-    image: mysql:8.0.25
+    image: mysql:8.0.28-oracle
```
TorstenDittmann, line 59: What does this change do? 🙂

TorstenDittmann (approved):

PR #130, opened 2022-04-27
Fix: Auth skip

Meldiron:
- Better reversion mechanism for skip() method
- Improved docs
- Improved error message

Let's imagine the following pseudo code:
```php
Auth::disable();

$doc = $db->getDocument('users', 'user1'); // Works

$doc = Auth::skip(fn() => $db->getDocument('users', 'user1')); // Works

$doc = $db->getDocument('users', 'user1'); // No longer work
```

If you disable auth, and later use `skip`, it re-enabled auth. After `skip` is finished, the status should be restored to what it was before skipping, instead of enabling it forcefully. 
Can be seen on: https://github.com/utopia-php/database/pull/130/files#diff-f0b0bb401b111034042efdb8f446197096a3fa9592d55d36ca9c6213770e5e25L161

### DIFF
--- a/src/Database/Validator/Authorization.php
+++ b/src/Database/Validator/Authorization.php
@@ -12,7 +12,7 @@ class Authorization extends Validator
      * 
      * @var array
      */
-    static $roles = ['role:all' => true];
+    static array $roles = ['role:all' => true];
 
     /**
      * Action of the authorization validation.
@@ -20,7 +20,7 @@ class Authorization extends Validator
      * 
      * @var string
      */
-    protected $action = '';
+    protected string $action = '';
 
     /**
      * Error message explaining why validation failed.
@@ -179,7 +179,7 @@ class Authorization extends Validator
      * 
      * @return mixed
      */
-    public static function skip(callable $callback)
+    public static function skip(callable $callback): mixed
     {
         $enabled = self::$status;
 
--- a/src/Database/Validator/Authorization.php
+++ b/src/Database/Validator/Authorization.php
@@ -27,7 +27,7 @@ class Authorization extends Validator
      * 
      * @var string
      */
-    protected $message = 'Authorization Error';
+    protected string $message = 'Authorization Error';
 
     /**
      * @param string $action Action to check for during validation.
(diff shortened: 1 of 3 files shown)


src/Database/Validator/Authorization.php
```diff
@@ -8,24 +8,31 @@
 class Authorization extends Validator
 {
     /**
+     * Array of allowed roles.
+     * 
      * @var array
      */
     static $roles = ['role:all' => true];
```
TorstenDittmann, line 15:
```suggestion
    static array $roles = ['role:all' => true];
```

src/Database/Validator/Authorization.php
```diff
@@ -8,24 +8,31 @@
...
 
     /**
+     * Action of the authorization validation.
+     * Mainly 'read' or 'write'.
+     * 
      * @var string
      */
     protected $action = '';
```
TorstenDittmann, line 23:
```suggestion
    protected string $action = '';
```

src/Database/Validator/Authorization.php
```diff
@@ -148,23 +173,25 @@ public static function setDefaultStatus(bool $status): void
...
     /**
-     * Skip Authorization
+     * Skip Authorization for the code executed inside the callback.
      * 
-     * Skips authorization for the code to be executed inside the callback
+     * @param callable $callback function to run without authorization
      * 
      * @return mixed
      */
     public static function skip(callable $callback)
```
TorstenDittmann, line 182:
```suggestion
    public static function skip(callable $callback): mixed
```

TorstenDittmann (commented):
Try to also fill out these types when missing 👍🏻

Meldiron:
Oooh now I see what you mean. All updated.

eldadfux:
@Meldiron it would have been very helpful to get some background and context in the PR description. What has caused us to change this mechanism?

Since this is a very sensitive security method in the entire architecture any extra information will help review the changes with more confidence.

Meldiron:
@eldadfux Updated. What we did until now was a bug I believe.

eldadfux:
@Meldiron please carefully test this doesn't break anything in Appwrite before we can merge it.

eldadfux:
Any chance this was already solved at #130 ?

eldadfux:
After checking with @abnegate, this was fixed and merged here: https://github.com/utopia-php/database/pull/186

PR #131, opened 2022-05-06
Fix: Force object

Meldiron:
We currently use `json_encode` inside `json` filter. Problem is, if you pass empty array `[]`, it will be represented as array `[]`.

In the context of JSON, I would like to open a discussion around adding `JSON_FORCE_OBJECT` to convert an empty PHP array into `{}`.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -142,7 +142,7 @@ class Database
                     return $value;
                 }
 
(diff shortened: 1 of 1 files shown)


Meldiron:
Closing PR, the issue can be solved outside of library by passing string as attribute value.

PR #132, opened 2022-05-13
Feat: Add update attribute methods

Meldiron:
New methods (for migration use, as of right now):

- updateAttributeRequired
- updateAttributeFormat
- updateAttributeFormatOptions
- updateAttributeFormatFilters
- updateAttributeDefault
- updateAttribute (to update size&type&signed&array)

The new private method introduced to prevent code duplication:
- updateAttributeMeta

TODO:

- [x] Write tests
- [x] Implement `adapter->updateAttribute`

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -703,10 +703,8 @@ class Database
      * 
      * @return bool
      */
-    public function updateAttribute(string $collection, string $id = null, string $type = null, int $size = null, bool $signed = null, bool $array = null): bool
+    public function updateAttribute(string $collection, string $id, string $type = null, int $size = null, bool $signed = null, bool $array = null): bool
     {
-        $success = true;
-
         $this->updateAttributeMeta($collection, $id, function($attribute) use($collection, $id, $type, $size, $signed, $array, &$success) {
             if($type !== null || $size !== null || $signed !== null || $array !== null) {
                 $type ??= $attribute->getAttribute('type');
@@ -741,14 +739,13 @@ class Database
                     ->setAttribute('signed', $signed)
                     ->setAttribute('array', $array);
     
-                $success = $this->adapter->updateAttribute($collection, $id, $type, $size, $signed, $array);
+                $this->adapter->updateAttribute($collection, $id, $type, $size, $signed, $array);
             }
-
 
             $attribute->setAttribute('array', $array);
         });
 
-        return $success;
+        return true;
     }
 
     /**
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -552,7 +552,7 @@ class Database
 
         $attributes = $collection->getAttribute('attributes', []);
 
-        $attributeIndex = \in_array($id, \array_map(fn($attribute) => $attribute['$id'], $attributes));
+        $attributeIndex = \array_search($id, \array_map(fn($attribute) => $attribute['$id'], $attributes));
 
         if($attributeIndex === false) {
             throw new Exception('Attribute not found');
@@ -561,18 +561,20 @@ class Database
         $attribute = $attributes[$attributeIndex];
 
         // Execute update from callback
-        $attribute = call_user_func($updateCallback, $attribute);
+        call_user_func($updateCallback, $attribute);
 
         // Save
         $attributes[$attributeIndex] = $attribute;
         $collection->setAttribute('attributes', $attributes, Document::SET_TYPE_ASSIGN);
 
+        /*
         if (
             $this->adapter->getRowLimit() > 0 &&
             $this->adapter->getAttributeWidth($collection) >= $this->adapter->getRowLimit()
         ) {
             throw new LimitException('Row width limit reached. Cannot create new attribute.');
         }
+        */
 
         if ($collection->getId() !== self::METADATA) {
             $this->updateDocument(self::METADATA, $collection->getId(), $collection);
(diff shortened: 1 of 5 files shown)


src/Database/Database.php
```diff
@@ -535,6 +535,224 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+     * @return Document
+     */
+    private function updateAttributeMeta(string $collection, string $id, callable $updateCallback): void
+    {
+        // Load
+        $collection = $this->getCollection($collection);
+
+        $attributes = $collection->getAttribute('attributes', []);
+
+        $attributeIndex = \in_array($id, \array_map(fn($attribute) => $attribute['$id'], $attributes));
```
TorstenDittmann, line 555: This returns `true` here and not the index 👍🏻

TorstenDittmann (commented):
This looks already so much better 💪🏻

src/Database/Database.php
```diff
@@ -535,6 +535,227 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+            throw new Exception('Attribute not found');
+        }
+
+        $attribute = $attributes[$attributeIndex];
+
+        // Execute update from callback
+        call_user_func($updateCallback, $attribute);
+
+        // Save
+        $attributes[$attributeIndex] = $attribute;
```
TorstenDittmann, line 567:
```suggestion
        // Execute update from callback
        call_user_func($updateCallback, $attributes[$attributeIndex]);
```

Not sure if this does the same? 🤔

src/Database/Database.php
```diff
@@ -535,6 +535,227 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+        $collection->setAttribute('attributes', $attributes, Document::SET_TYPE_ASSIGN);
+
+        /*
+        if (
+            $this->adapter->getRowLimit() > 0 &&
+            $this->adapter->getAttributeWidth($collection) >= $this->adapter->getRowLimit()
+        ) {
+            throw new LimitException('Row width limit reached. Cannot create new attribute.');
+        }
+        */
```
TorstenDittmann, line 577: Why is this outcommented?

src/Database/Database.php
Meldiron: It does, still works as a refference. Test passing.

src/Database/Database.php
Meldiron: I re-enabled it, and tests pass. I remember it was not needed (as we don't touch row limit by changing meta).

Should we keep it or remove? 🤔

src/Database/Database.php
```diff
@@ -535,6 +535,222 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+     * @param bool $signed
+     * @param bool $array
+     * 
+     * To update attribute key (ID), use renameAttribute instead.
+     * 
+     * @return bool
+     */
+    public function updateAttribute(string $collection, string $id = null, string $type = null, int $size = null, bool $signed = null, bool $array = null): bool
+    {
+        $success = true;
```
TorstenDittmann, line 708: You can remove this and just return `true` on success

src/Database/Database.php
```diff
@@ -535,6 +535,222 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+                        break;
+                }
+    
+                $attribute
+                    ->setAttribute('type', $type)
+                    ->setAttribute('size', $size)
+                    ->setAttribute('signed', $signed)
+                    ->setAttribute('array', $array);
+    
+                $success = $this->adapter->updateAttribute($collection, $id, $type, $size, $signed, $array);
```
TorstenDittmann, line 744: You don't need to assign this, it should throw an error

src/Database/Database.php
```diff
@@ -535,6 +535,222 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+     * @param string $type
+     * @param int $size utf8mb4 chars length
+     * @param bool $signed
+     * @param bool $array
+     * 
+     * To update attribute key (ID), use renameAttribute instead.
+     * 
+     * @return bool
+     */
+    public function updateAttribute(string $collection, string $id = null, string $type = null, int $size = null, bool $signed = null, bool $array = null): bool
```
TorstenDittmann, line 706: Why is `$id` optional?

src/Database/Database.php
```diff
@@ -535,6 +535,222 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+                    ->setAttribute('type', $type)
+                    ->setAttribute('size', $size)
+                    ->setAttribute('signed', $signed)
+                    ->setAttribute('array', $array);
+    
+                $success = $this->adapter->updateAttribute($collection, $id, $type, $size, $signed, $array);
+            }
+
+
+            $attribute->setAttribute('array', $array);
```
TorstenDittmann, line 748: leftover?

src/Database/Database.php
Meldiron: No more. Should not be.

src/Database/Database.php
Meldiron: Nope, on purpose. `array` status is metadata, so we update it just like any other metadata. But we need to do it in `updateAttribute`, because depending if its array or not, a type will change

src/Database/Database.php
Meldiron: PDO `execute` returns boolean, so we just return that. I have seen this logic in `createAttribute` or `createAttribute`.

src/Database/Database.php
TorstenDittmann: But we already do it in line `740`

src/Database/Database.php
Meldiron: Removed, not nessessary. Added it into `updateAttribute()`tho

TorstenDittmann (approved):

PR #133, opened 2022-05-30
Feat: Improve SQL escaping

Meldiron:
- Fix problems with reserved keywords such as `order`
- Implement bunch of tests to test for reserved keywords


tests/Database/Adapter/MariaDBTest.php
```diff
@@ -39,6 +39,27 @@ static function getAdapterRowLimit(): int
...
+            'create',
+            'delete',
+            'update',
+            'table',
+            'from',
+            'by',
+            'in'
+        ];
+    }
+
```
TorstenDittmann, line 62: I don't think this needs to be a function, it can be a property of the class 👍🏻

stnguyen90: Btw, here's the list of reserved keywords for mariadb: https://mariadb.com/kb/en/reserved-words/

eldadfux: I think it make sense, this forces us to provide the list in each adapter. This way tests coverage can be much better.

eldadfux (approved):

TorstenDittmann (approved):

PR #144, opened 2022-06-22
Fix: Required string structure validator

Meldiron:
Currently, you can pass an empty string into a string attribute that is required. This is not expected behavior. This PR makes empty sting invalid in the required string attribute.

0 for integer still works, true/false for boolean still works.

- [x] Added new tests


Meldiron:
We consider this intended behavior. Closing the PR

PR #145, opened 2022-06-23
Fix: Internal attribute protection

Meldiron:
PR makes sure library users cannot edit internal attributes when updating documents.

- [ ] Implement new tests


Meldiron:
Closing PR to keep this logic appwrite-specific. That allows more migration capabilities.

Meldiron:
Just revisited this. Good think we made this appwrite-specific. We are now considering allowing updating dates ($createdAt, $updatedAt) from Server SDK.

PR #147, opened 2022-06-25
Fix avoid duplicate permissions

gepd:
Currently if you pass duplicate permission to `updateDocument` you will get `Error: Document already exists`

```js
const read = ["user:abc", "user:cde", "user:abc"];
updateDocument("collection","documentId", { key: "value"}, read); // Error: Document already exists
```

This is because [this lines](https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L594-L601) doesn't remove the duplicate items:

```js
$readAdded = array_diff($document->getRead(), $permissions['read']);
$writeAdded = array_diff($document->getWrite(), $permissions['write']);
```

I have used `array_unique` in `getRead` and `getWrite` as this way we make sure we always get unique permission when we request permissions

### DIFF
--- a/src/Database/Document.php
+++ b/src/Database/Document.php
@@ -78,7 +78,7 @@ class Document extends ArrayObject
      */
     public function getRead(): array
     {
-        return array_unique($this->getAttribute('$read', []));
+        return $this->getAttribute('$read', []);
     }
 
     /**
@@ -86,7 +86,7 @@ class Document extends ArrayObject
      */
     public function getWrite(): array
     {
-        return array_unique($this->getAttribute('$write', []));
+        return $this->getAttribute('$write', []);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Database/Document.php
```diff
@@ -78,15 +78,15 @@ public function getCollection(): string
      */
     public function getRead(): array
     {
-        return $this->getAttribute('$read', []);
+        return array_unique($this->getAttribute('$read', []));
```
Meldiron, line 81: Can we please add tests to check this functionality? Best way to check:

1. Remove these changes
2. Write tests that you expect to fail
3. Tests fail
4. Re-implement this
5. Run tests
6. Tests pass

gepd: done!

TorstenDittmann (approved):

PR #151, opened 2022-07-13
feat: index validation without order

TorstenDittmann:
Ignores query order for validation.

These Queries are Allowed:
- A,B,C
- A,C,B
- B,A,C
- B,C,A
- C,A,B
- C,B,A

Previous implementation only allowed following Queries:
- A,B,C

This will make Indexes more accessible 👍🏻

### DIFF
--- a/tests/Database/Validator/QueriesTest.php
+++ b/tests/Database/Validator/QueriesTest.php
@@ -254,60 +254,6 @@ class QueriesTest extends TestCase
         ]));
     }
 
-    public function testCoveringIndexQueries()
-    {
-        $validator = new Queries($this->queryValidator, [new Document([
-            '$id' => 'testindex5',
-            'type' => 'key',
-            'attributes' => [
-                'title',
-                'price',
-                'rating'
-            ],
-            'orders' => []
-        ])], true);
-
-        // Test for SUCCESS
-        $this->assertTrue($validator->isValid([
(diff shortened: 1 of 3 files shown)


tests/Database/Validator/QueriesTest.php
```diff
@@ -164,46 +167,146 @@ public function tearDown(): void
...
+            Query::parse('price.lesserEqual(6.50)'),
+            Query::parse('year.lesserEqual(2002)')
+        ]));
+
+        $this->assertFalse($validator->isValid([
+            Query::parse('rating.lesserEqual(2002)'),
+            Query::parse('price.lesserEqual(6.50)')
+        ]));
+    }
+
```
fogelito, line 303: $this->assertFalse($validator->isValid([
            Query::parse('price.greaterEqual(7.5)'),
            Query::parse('price.lesserEqual(6.50)')
        ]));
will this require 2 indexes of "price"?

TorstenDittmann: Added it 👍🏻 

Is working as expected 🙂

eldadfux:
For the sake of our records. This PR has changed to tackle only the index orders limitation. Queries with proper indexes will be accepted now regardless of the attributes order inside the index.

Covering indexes if accepted, will be tackled in a separate PR.

eldadfux (approved):

PR #153, opened 2022-07-15
Feat: Queries Gen 2

Meldiron:
Implementation of new queries syntax.

Goal of new implementation: https://github.com/appwrite/appwrite/discussions

- [x] Existing tests updated
- [x] Implemented new tests


src/Database/Query.php
```diff
@@ -140,106 +105,130 @@ public static function isOperator(string $value): bool
...
+            $param = substr($param, 1, -1); // Remove '' or ""
+            return $param;
+        }
+
+        // Unknown format
+        return $param;
     }
 }
+
+Query::$TYPE_ALIASES = [
```
eldadfux, line 227: This is an anti-pattern. + we can and should treat aliases similar to how we use regular methods to help reduce complexity of this class.

src/Database/Validator/QueryValidator.php
```diff
@@ -90,32 +90,32 @@ public function getDescription(): string
     public function isValid($query): bool
     {
         // Validate operator
-        if (!in_array($query->getOperator(), $this->operators)) {
-            $this->message = 'Query operator invalid: ' . $query->getOperator();
+        if (!in_array($query->getMethod(), $this->operators)) {
```
eldadfux, line 93: If we move from operators to methods we should adapt the terminology library-wide.

eldadfux (changes requested):
Some thoughts:
1. We should try and make the Query class behave closer to our external API for consistency, we can have the same existing constructor with some factory method to represent different methods. 
2. I feel the Query interface can tell a better story. A lot of the new methods could be private, and expose a ,ore intuitive interface, I would love to rethink the structure together.

tests/Database/Validator/QueriesTest.php
```diff
@@ -104,8 +104,8 @@ public function setUp(): void
 
         $this->queryValidator = new QueryValidator($attributes);
 
-        $query1 = Query::parse('title.notEqual("Iron Man", "Ant Man")');
-        $query2 = Query::parse('description.equal("Best movie ever")');
+        $query1 = Query::parse('notEqual("title", ["Iron Man", "Ant Man"])');
+        $query2 = Query::parse('equal("description", "Best movie ever")');
```
stnguyen90, line 108: @Meldiron, would you please make sure to have a test for text with a comma to address https://github.com/appwrite/appwrite/issues/3089? Foe example:

```php
 $query = Query::parse('equal("text", ["Contains,Comma", "No Comma"])');
 ```

stnguyen90 (changes requested):

Meldiron:
Closing in favour of new branch `2.1`

src/Database/Query.php
Meldiron: We simplified aliases. Tho we can't support the page now, the system is much simpler. We can allow complex aliases in future iterations.

tests/Database/Validator/QueriesTest.php
Meldiron: Addressed.

src/Database/Validator/QueryValidator.php
Meldiron: Adapted. All operators renamed to methods.

PR #155, opened 2022-07-19
Feat attr datetime

fogelito:
Adding new Class DateTime for PHP \Datetime changes from Unix time to DateTime easily


src/Database/DateTime.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param string|null $datetime
+     * @return bool
+     */
+    public static function isValid(?string $datetime): bool
+    {
+        if (empty($datetime)) {
+            return false;
+        }
+
+        if(!preg_match("/(\d{4})-(\d{2})-(\d{2}) (\d{2}):(\d{2}):(\d{2})/", $datetime)){
```
TorstenDittmann, line 29: Why do we need this?

Shouldn't the `new \DateTime(...)` take care of this?

src/Database/DateTime.php
```diff
@@ -0,0 +1,74 @@
+<?php
+
+namespace Utopia\Database;
+
+use Exception;
+
+class DateTime
```
TorstenDittmann, line 7: We need Unit tests for this 👍🏻

src/Database/DateTime.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    protected static string $format = 'Y-m-d H:i:s';
```
TorstenDittmann, line 9: I think we can start using more precision now 👍🏻 what precision can work for all Databases?

src/Database/Validator/DatetimeValidator.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace Utopia\Database\Validator;
+
+use Utopia\Validator;
+
+class DatetimeValidator extends Validator
```
TorstenDittmann, line 7: We should be able to use the Validator in `src/`

src/Database/DateTime.php
fogelito: ok!

fogelito: yep 3 :)

src/Database/DateTime.php
fogelito: Does functions like these are considered tests?
DateTime::isValid
DateTime::now
DateTime::addSeconds?
I am using these on tests
will add some more...

src/Database/DateTime.php
fogelito: Removed it for now

src/Database/DateTime.php
```diff
@@ -0,0 +1,70 @@
...
+class DateTime
+{
+    protected static string $format = 'Y-m-d H:i:s.v';
+
+    /**
+     * @throws Exception
+     */
+    public function __construct()
+    {
+        Throw new Exception("Use static right now", 500);
```
TorstenDittmann, line 16: We can make the class static so it wont have a constructor right?

tests/Database/Validator/DateTimeTest.php
```diff
@@ -0,0 +1,33 @@
...
+    public function tearDown(): void
+    {
+    }
+
+    public function testCreateDatetime()
+    {
+        $this->assertGreaterThan(DateTime::addSeconds(new \DateTime(), -3), DateTime::now());
+        $this->assertEquals(false, DateTime::isValid("2022-13-04 11:31:52.680"));
+        $this->assertGreaterThan('2022-7-2', '2022-7-2 11:31:52.680');
+        $this->assertEquals(23, strlen(DateTime::now()));
```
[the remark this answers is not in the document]
fogelito: added some more tests

src/Database/Database.php
```diff
@@ -1254,6 +1282,8 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
 
         $cursor = empty($cursor) ? [] : $cursor->getArrayCopy();
 
+        $queries = DateTime::queries($collection, $queries);
```
TorstenDittmann, line 1285: We should make this more generic, so future possible edge cases can be handled by the same method.

Something like `$this->convertQueries($collection, $queries);` and take care of that there.

So take the code from `DateTime::queries` and instead of an if condition have switch case, so we can extend it in the future 👍🏻

PR #156, opened 2022-07-21
DRAFT: Next&Previous support

Meldiron:
DRAFT PR!

I believe it is possible to figure out if the query has the next and previous document within one existing query we currently do.

I got stuck on cursor pagination, tho it should be possible since we compare internal integer IDs anyway.

This answer makes me think it is possible: https://stackoverflow.com/questions/71699810/it-is-possible-in-cursor-based-pagination-to-get-the-prev-and-next-cursor-in-the

- `limit + 1` should give `next` document
- `<=` instead of `<` in cursor condition should give `previous` document


Meldiron:
Closing the PR as this is most likely not possible due to sorting while using cursor.

PR #157, opened 2022-07-22
Feat: Query gen 2(.1)

Meldiron:
Previous PR: https://github.com/utopia-php/database/pull/153

TODO after this PR:

- [ ] Remove parameters from `find()` (such as limit, offset, order, cursor, ...) and get it from $queries instead
- [ ] Upgrade Appwrite to new method signatures
- [ ] Implement new queries into Appwrite Database service
- [ ] Update Appwrite `list` endpoints so they all allow queries


TorstenDittmann (approved):

src/Database/Query.php
```diff
@@ -4,134 +4,129 @@
...
     /**
-     * Get all query details as array
+     * Helper method changing first param. In first param we often store attribute
      *
-     * @return array
+     * @param string $value
+     * @return self
      */
-    public function getQuery(): array
+    public function setFirstParam(string $value): self
```
eldadfux, line 75: The API of the class feels very oriented towards the technical implementation instead of telling the story of the use case for this class and method. I think, `getAttribute`, `getMethod`, and `getValues` would make much more sense and be more intuitive when reading the code.

We could have also introduce static builder methods for this class that would make the API closer to what we have in our Appwrite SDK (not meaning they should be coupled, but they just feel more convenient). For example:
```php
Query::equal(string $attribute, array $values):self;
```

tests/Database/Benchmark/QueryBench.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Utopia\Tests\Benchmark;
+
+use Generator;
+use Utopia\Database\Query;
+use Utopia\Database\QueryV1;
+use PhpBench\Attributes as Bench;
+
+class QueryBench
+{
+    public function provideAttributes(): Generator
```
eldadfux, line 12: We should properly document this new test and how to run it, alongside other tests related info.

eldadfux (changes requested):

src/Database/Query.php
```diff
@@ -140,106 +135,286 @@ public static function isOperator(string $value): bool
...
+        $stack = []; // Stack of syntactical symbols
+
+        $stringStackState = false;
+        $arrayStackState = false;
+
+        // Loop thorough all characters
+        for ($i = $parametersStart; $i < $paramsEnd; $i++) {
+            $char = $filter[$i];
+
+            $isStringStack = $stringStackState;
```
stnguyen90, line 187: Would it be simpler to not use a stack for string?

src/Database/Query.php
```diff
@@ -140,106 +135,286 @@ public static function isOperator(string $value): bool
...
+            $isArrayStack = $arrayStackState;
+            
+            // String support + escaping support
+            if (
+                (self::isQuote($char)) && // Must be string indicator
+                $filter[$i - 1] !== '\\' // Must not be escaped; first cant be
```
stnguyen90, line 193: Would this still be able to catch something like: `Query.equal("attr", "\\")`?

src/Database/Query.php
```diff
@@ -140,106 +135,286 @@ public static function isOperator(string $value): bool
...
+                }
+
+                continue;
+            }
+
+            // Array support
+            if (!($isStringStack)) {
+                if ($char === static::CHAR_BRACKET_START) {
+                    // Start of array
+                    $stack[] = $char;
```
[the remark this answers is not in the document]
Meldiron: We don't have support for nested arrays, as they are not required for now. It will be added when it becomes relevant (with array filters).

For now we only support 2 layers - one being params array, one being array value. So for example:

Works: `[ 'equal', ['Matej', 'Eldad'] ]`
Doesn't work: `['equal', [ ['editor','owner'] ] ]`

src/Database/Query.php
```diff
@@ -4,134 +4,129 @@
...
-     * Construct.
-     *
      * Construct a new query object
-     *
-     * @param string $attribute
-     * @param string $operator
-     * @param array $values
      */
-    public function __construct(string $attribute, string $operator, array $values)
+    public function __construct(string $method, array $params)
```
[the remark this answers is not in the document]
Meldiron: We removed `operator`. In `equals('name', 'Matej')` all is good, but in a query like `limit(5)` we don't have an operator.

`attribute` was also removed, since that is also not present in `limit(5)`.

We added `method` which replaced `attribute` in the placement of the query. A method is for example `equals`, `limit`, or `offset`.

---

I can see what you mean with a step back. I was thinking about implementing separate interfaces(classes?) for each method. That would give you more high-level of building a query. Would that make sense?

src/Database/QueryV1.php
```diff
@@ -0,0 +1,245 @@
+<?php
+
+namespace Utopia\Database;
+
+class QueryV1
```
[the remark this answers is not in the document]
Meldiron: It's for benchmark. @TorstenDittmann  do we want to keep benchmarks as part of PR?

tests/Database/Benchmark/QueryBench.php
Meldiron: @TorstenDittmann  do we want to keep benchmarks as part of PR?

src/Database/Query.php
Meldiron: Not sure if I understand the comment.

We keep a state variable for performance improvement reasons. It simply holds true/false whether the last symbol in the stack array is a string (`"'`).

We need to know if we are in a string stack since `]` inside string should be simply a character in string, but outside of string it ends an array.

We need stack variable since, at the end of anything (for example, end of string), we should still know the previous stack - for example. if we are still in array parameter or not.

src/Database/Query.php
Meldiron: Added a new test that passes:
```php
        $query = Query::parse('equal(1, "Hello\\\\World")');
        $this->assertCount(2, $query->getParams());
        $this->assertEquals(1, $query->getParams()[0]);
        $this->assertEquals("Hello\\\\World", $query->getParams()[1]);
```

It's there twice in reality. We need four because of PHP escaping.

src/Database/Query.php
Meldiron: I wanted to avoid calling it `getMethod` since you can still call it on `limit(5)` and there you would get `5` as method. Tho, I am opened for suggestions. Should I rename them to original methods?

We can introduce it, yes. That's what I though about in this comment: https://github.com/utopia-php/database/pull/157#discussion_r928212894

> I can see what you mean with a step back. I was thinking about implementing separate interfaces(classes?) for each method. That would give you more high-level of building a query. Would that make sense?

If you give me green, I can go ahead and implement it as static method

src/Database/Query.php
stnguyen90: I was thinking it might be simpler to have $lastQuote that's either ' or " or empty/null instead of using the stack and state variables.

src/Database/Query.php
stnguyen90: I was thinking of this case:

```php
$query = Query::parse('equal(1, "Hello\\\\")');
```

stnguyen90: Actually, how about this case: 

```php
$query = Query::parse('equal(1, "Hello\\\\", "World")');
```

Meldiron: Good catch with the second test. escaped `\` was still considered escaping symbol in case of the end of the string. Fixed now.

Both tests added and are passing now.

src/Database/Query.php
Meldiron: Good point. this is possible because we never have a string inside a string - we don't need a stack. We need stack only because we could (in the future) have an array in an array.

Refactored a little to address this. Changes have no performance impact (or really small).

src/Database/QueryV1.php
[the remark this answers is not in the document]
Meldiron: Removing for now, can be addressed in separate PR in future.

tests/Database/Benchmark/QueryBench.php
Meldiron: Removing for now, can be addressed in separate PR in future.

src/Database/Query.php
eldadfux: I think that with the builder/factory static methods in place we could just set attribute to null, and have proper access methods in place.

TorstenDittmann: We might want to have multiple Query Classes, since they are really hard to mix in combination with Validators.

- The Equal query has stuff like attribute and values (any data type)
- The Limit query has only a single value which is int
- The Offset query has only a single value which is int
- The Order Query has only a single values which are string (maybe even multiple values if we don't want the user to pass multiple orders)
- ...

I am not sure if we can get an interface that makes sense for all. So what about a Base class, extending it for necessary Queries. Otherwise we will have so many conditions for validating everything scattered around the one class for Query and Validator.

stnguyen90 (approved):
query parsing LGTM

PR #158, opened 2022-07-23
Feat: Improved performance

Meldiron:
Added benchmarks and improved code performance of parser


Meldiron:
Original:

CleanShot 2022-07-23 at 13 14 13


After changes:

image

PR #161, opened 2022-08-02
Refactor permissions

abnegate:
- [BREAKING] Introduce new permissions syntax replacing `write` permission with `create`, `update` and `delete`.
- [BREAKING] Remove `$read` attribute from `Document`
- [BREAKING] Remove `$write` attribute from `Document`
- Add `$permissions` array attribute to `Document`
- Add `PERMISSIONS` const array to `Database` to allow extending existing permission types
- Add `getPermissions`, `getPermission($type)`, `getCreate`, `getUpdate` and `getDelete` functions to `Document`
- Update `Permissions` validator to handle input of new syntax
- Update tests to support new syntax


src/Database/Validator/Permissions.php
```diff
@@ -60,138 +66,138 @@ public function getDescription(): string
...
-                        $teamId = \substr($value, 0, $pos);
-                        $teamRole = \substr($value, $pos + 1);
-
-                        // $teamRole must not be empty
-                        // Case $teamId < 1 already covered.
-                        if (strlen($teamRole) < 1) {
-                            $this->message = 'Team role must not be empty.';
-
+            $submatches = [];
+            \array_shift($matches);
```
TorstenDittmann, line 134: for what do we remove the first value?

src/Database/Validator/Permissions.php
```diff
@@ -60,138 +66,138 @@ public function getDescription(): string
...
+            //          - If the permission ID is not found, the entire permission parameter will be considered invalid.
+            //          - If the permissions ID is not a valid ID, the entire permission parameter will be considered invalid.
+            //          - If the dimension should be from the known set and is not found, the entire permission parameter will be considered invalid.
+            //        - If any invalid permission name is found, the entire permission string will be considered invalid.
+
+            // Inner permission string matcher (e.g. "user:123abc", "team:123abc/role") where permission name is matched against known permissions.
+            $permissionMatcher = '((?:' . \implode('|', $this->permissions) . ')(?::(?:[a-z\d]+))?(?:\/(?:[a-z]+))?)';
+
+            // Captures the permissions type (e.g. "read", "update") and ensures at least 1 permission string is provided.
+            $permissionString = '/^(?:' . \implode('|', $this->methods) . ')\(' . $permissionMatcher . '(?:,\s*' . $permissionMatcher . ')*\)$/';
```
TorstenDittmann, line 122: this will probably get more readable if we move logic out of the string building.

something like this:
```suggestion
            $permissions = \implode('|', $this->permissions);
            $methods = \implode('|', $this->methods);
            $permissionMatcher = "((?:{$permissions})(?::(?:[a-z\d]+))?(?:\/(?:[a-z]+))?)";

            // Captures the permissions type (e.g. "read", "update") and ensures at least 1 permission string is provided.
            $permissionString = "/^(?:{$methods})\({$permissionMatcher}(?:,\s*{$permissionMatcher})*\)$/";
```

src/Database/Document.php
```diff
@@ -76,17 +72,87 @@ public function getCollection(): string
...
+    /**
+     * @param string $type
+     * @return array
+     */
+    public function getPermission(string $type): array
+    {
+        if (\method_exists($this, 'get' . \ucfirst($type))) {
+            return [$this, 'get' . \ucfirst($type)]();
+        }
+        return [];
```
TorstenDittmann, line 89: we should probably have a unit test for that, so that it will fail if we are missing such method in the future.

src/Database/Validator/Permissions.php
```diff
@@ -60,138 +66,138 @@ public function getDescription(): string
...
+            //          - If the permission ID is not found, the entire permission parameter will be considered invalid.
+            //          - If the permissions ID is not a valid ID, the entire permission parameter will be considered invalid.
+            //        - Then the third iteration will match "123abc" as a permissions ID and "edit" as a dimension.
+            //          - If the permission ID is not found, the entire permission parameter will be considered invalid.
+            //          - If the permissions ID is not a valid ID, the entire permission parameter will be considered invalid.
+            //          - If the dimension should be from the known set and is not found, the entire permission parameter will be considered invalid.
+            //        - If any invalid permission name is found, the entire permission string will be considered invalid.
+
+            // Inner permission string matcher (e.g. "user:123abc", "team:123abc/role") where permission name is matched against known permissions.
+            $permissionMatcher = '((?:' . \implode('|', $this->permissions) . ')(?::(?:[a-z\d]+))?(?:\/(?:[a-z]+))?)';
```
TorstenDittmann, line 119: we need uppercase and some allowed letters as well right?

```suggestion
            $permissionMatcher = '((?:' . \implode('|', $this->permissions) . ')(?::(?:[a-zA-Z._\-\d]+))?(?:\/(?:[a-zA-Z._\-]+))?)';
```

src/Database/Validator/Permissions.php
abnegate: The first match contains the text that matched the full pattern, we only care that it _did_ match, then we want to shift past that and process the capture groups.

Matches looks like:

```php
array(3) {
  [0]=>
  string(32) "create(user:123abc, team:123abc)"
  [1]=>
  string(11) "user:123abc"
  [2]=>
  string(11) "team:123abc"
}
```

eldadfux (changes requested):

src/Database/Validator/Authorization.php
```diff
@@ -10,7 +10,10 @@ class Authorization extends Validator
     /**
      * @var array
      */
-    static $roles = ['role:all' => true];
+    static $roles = [
+        'any' => true,
+        'role:all' => true,
```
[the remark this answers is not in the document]
abnegate: Have added a task for all deprecated components to be removed

src/Database/Validator/Permissions.php
```diff
@@ -2,33 +2,40 @@
...
-     */
-    protected $roles = [
+    protected array $legacyDimensions = [
         'all',
-        'guest',
         'member',
+        'guest'
+    ];
+
+    protected array $statusDimensions = [
```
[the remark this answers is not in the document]
abnegate: They are being verified when used positionally, so:

`users:verified`,
`user:123/verified`,
`status:verified`

would be considered valid, but

`users:invalid`,
`user:123/invalid`,
`status:invalid`

would not. Note that validation depends on the permission type, so a team role in the dimension position is still considered valid e.g.:

`team:123/anythingisvalid`

eldadfux (commented):
All looks good, just waiting for the ID class, and I think we miss tests for the new Role class.

src/Database/Document.php
```diff
@@ -107,20 +141,19 @@ public function getUpdatedAt(): ?string
...
      */
     public function getAttributes(): array
     {
         $attributes = [];
 
         foreach ($this as $attribute => $value) {
-            if(array_key_exists($attribute, [
-                '$id' => true,
+            if (\array_key_exists($attribute, [
+                '$id' => ID::custom(true),
```
eldadfux, line 153:
```suggestion
                '$id' => true,
```

I think this should stay just true.

src/Database/ID.php
```diff
@@ -0,0 +1,23 @@
...
+
+class ID
+{
+
+    /**
+     * Create a new unique ID
+     */
+    public static function unique(): string
+    {
+        return uniqid();
```
eldadfux, line 13:
```suggestion
        return 'uniqid()';
```

The custom ID is generated from a different method. We just use the unique function from some tests.

src/Database/Permission.php
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public static function create(Role $role): string
+    {
+        $permission =  new Permission(
+            'create',
+            $role->getRole(),
+            $role->getIdentifier(),
+            $role->getDimension()
+        );
+        return $permission->toString();
```
eldadfux, line 84: In the Role and Query classes we just return the object, why do we return a string here?

src/Database/Role.php
```diff
@@ -0,0 +1,150 @@
...
+    {
+    }
+
+    /**
+     * Parse a role string into a Role object
+     *
+     * @param string $role
+     * @return Role
+     */
+    public static function parse(string $role): Role
```
eldadfux, line 21: Let's always keep static methods at the bottom of the class.

eldadfux (commented):
Some more small comments.

src/Database/Permission.php
abnegate: The result is a string here so we can pass it directly to the `$permissions` array

Alternatively, we could call `->toString()` for each permission instead?

eldadfux: I think this is fine, but I was wondering if we can avoid string internally and just use the object inside the adapters.

PR #162, opened 2022-08-08
Feat query gen 2.2

stnguyen90:
Based on [feedback in previous PR](https://github.com/utopia-php/database/pull/157#discussion_r928206123), update Query class to be used like:

```php
// signature of constructor is (string $method, string $attribute = '', array $values = [])
$query = new Query('equal', 'actors', ['Brad Pitt', 'Johnny Depp']);

$this->assertEquals('equal', $query->getMethod());
$this->assertEquals('actors', $query->getAttribute());
$this->assertEquals('Brad Pitt', $query->getValues()[0]);

// other examples
$query = new Query('lessThan', 'score', [10]);
$query = new Query('orderDesc', 'score');
$query = new Query('limit', values: [10]);

// static helper methods
$query = Query::equal('title', ['Iron Man']); // the only helper method that accepts arrays
$query = Query::greaterThan('score', 10); // value is not array
$query = Query::search('search', 'John Doe'); // also not array
$query = Query::orderAsc('score');
$query = Query::limit(10);
$query = Query::cursorAfter('cursor');
```

Update parsing of queries to be like:

```php
$query = Query::parse('equal("actors", ["Brad Pitt", "Johnny Depp"])');

$this->assertEquals('equal', $query->getMethod());
$this->assertEquals('actors', $query->getAttribute());
$this->assertEquals("Brad Pitt", $query->getValues()[0]);
$this->assertEquals("Johnny Depp", $query->getValues()[1]);

$query = Query::parse('equal("actors", "Brad Pitt")');

$this->assertEquals('equal', $query->getMethod());
$this->assertEquals('actors', $query->getAttribute());
$this->assertEquals("Brad Pitt", $query->getValues()[0]);
```


src/Database/Query.php
```diff
@@ -140,106 +117,401 @@ public static function isOperator(string $value): bool
...
+    public static function greaterThanEqual(string $attribute, $value): self
+    {
+        return new self(self::TYPE_GREATEREQUAL, $attribute, [$value]);
+    }
 
-            // Trim whitespace from around $value
+    /**
+     * Helper method to create Query with contains method
+     */
+    public static function contains(string $attribute, $value): self
```
stnguyen90, line 457: Should contains accept `array $values` or 1 `$value`? Probably an array, right? Since you may want to check if the array attribute contains value1 or value2?

src/Database/Query.php
```diff
@@ -140,106 +117,401 @@ public static function isOperator(string $value): bool
...
+    {
+        return new self(self::TYPE_CONTAINS, $attribute, [$value]);
+    }
 
-            $value = trim($value);
+    /**
+     * Helper method to create Query with search method
+     */
+    public static function search(string $attribute, $value): self
```
stnguyen90, line 465: Right now, we can do an OR search by passing 1 string with each value separated by space. Should this signature be changed to `array $values`, though, and then we do `implode(' ', $values)` server side?

src/Database/Query.php
```diff
@@ -140,106 +117,401 @@ public static function isOperator(string $value): bool
...
+    public static function cursorAfter(string $value): self
+    {
+        return new self(self::TYPE_CURSORAFTER, values: [$value]);
+    }
 
-        return [$operator, $values];
+    /**
+     * Helper method to create Query with cursorBefore method
+     */
+    public static function cursorBefore(string $value): self
```
stnguyen90, line 513: Looks like this `string $value` should be `Document $value`, right?

src/Database/Query.php
```diff
@@ -140,106 +117,401 @@ public static function isOperator(string $value): bool
...
+    }
 
-            switch (true) {
-                // type casted to int or float by "+" operator
-                case is_numeric($value):
-                    return $value + 0;
+    /**
+     * Helper method to create Query with orderDesc method
+     */
+    public static function orderDesc(string $attribute): self
```
stnguyen90, line 473: In some of our test cases, we order by "natural order" where we specify direction but no attribute. Is it to weird to do `Query::orderDesc('')` for those cases?

src/Database/Query.php
fogelito: On MYSQL Fulltext can take a few attributes such as 
SELECT * FROM collection WHERE MATCH(column1, column2) AGAINST ('text')
I know today there is no support, only a single attribute

PR #163, opened 2022-08-09
Feat query gen 2.3

stnguyen90:
Based on previous feedback:

* https://github.com/utopia-php/database/pull/157#discussion_r928206123
* https://github.com/utopia-php/database/pull/162

This PR updates the Query class to be used like:

```php
// signature of constructor is (string $method, string $attribute = '', array $values = [])
$query = new Query('equal', 'actors', ['Brad Pitt', 'Johnny Depp']);

$this->assertEquals('equal', $query->getMethod());
$this->assertEquals('actors', $query->getAttribute());
$this->assertEquals('Brad Pitt', $query->getValues()[0]);

// other examples
$query = new Query('lessThan', 'score', [10]);
$query = new Query('orderDesc', 'score');
$query = new Query('limit', values: [10]);

// static helper methods
$query = Query::equal('title', ['Iron Man']); // accepts array
$query = Query::greaterThan('score', 10); // value is not array
$query = Query::search('search', 'John Doe'); // also not array
$query = Query::orderAsc('score');
$query = Query::orderDesc(''); // empty string to sort on natural order
$query = Query::limit(10);
$query = Query::cursorAfter($cursorDocument); // takes a Document
$query = Query::contains('arrayAttribute', [1, 2, 3]); // accepts array
```

Update parsing of queries to be like:

```php
$query = Query::parse('equal("actors", ["Brad Pitt", "Johnny Depp"])');

$this->assertEquals('equal', $query->getMethod());
$this->assertEquals('actors', $query->getAttribute());
$this->assertEquals("Brad Pitt", $query->getValues()[0]);
$this->assertEquals("Johnny Depp", $query->getValues()[1]);

$query = Query::parse('equal("actors", "Brad Pitt")');

$this->assertEquals('equal', $query->getMethod());
$this->assertEquals('actors', $query->getAttribute());
$this->assertEquals("Brad Pitt", $query->getValues()[0]);
```

In addition, this PR [updates the Database find methods to only use queries](https://github.com/utopia-php/database/pull/163/commits/4348f07d232678452b47b2744a2c281198e0a606) (the Adapter's find method remains the same).

Finally, this PR also [updates the Queries Validator to validate limit, offset, cursor, and order queries](https://github.com/utopia-php/database/pull/163/commits/478098a4ddabbcf1a0425c6dae1a7e417123331d).


src/Database/Adapter/MariaDB.php
```diff
@@ -770,23 +770,60 @@ public function deleteDocument(string $collection, string $id): bool
...
+                    $cursor = $values[0];
+                    $cursorDirection = $method == Query::TYPE_CURSORAFTER ? Database::CURSOR_AFTER : Database::CURSOR_BEFORE;
+                    break;
+
+                default:
+                    $filters[] = $query;
+                    break;
+            }
+        }
+
```
stnguyen90, line 826: Should this be refactored into the base class? Something like:

```php
function processQueries(array $queries = []): array
```

where the array returned will look like:

```php

[
  'filters' => $filters,
  'limit' => $limit,
  'offset' => $offset,
  'orderAttributes' => $orderAttributes,
  'orderTypes' => $orderTypes,
  'cursor' => $cursor,
  'cursorDirection' => $cursorDirection,
]
```

src/Database/Adapter/MariaDB.php
```diff
@@ -770,23 +770,60 @@ public function deleteDocument(string $collection, string $id): bool
...
+                case Query::TYPE_ORDERDESC:
+                    if (!empty($attribute)) {
+                        $orderAttributes[] = $attribute;
+                    }
+
+                    $orderTypes[] = $method == Query::TYPE_ORDERASC ? Database::ORDER_ASC : Database::ORDER_DESC;
+                    break;
+
+                case Query::TYPE_LIMIT:
+                    $limit = $values[0];
```
christyjacob4, line 808: Let's check for null safety here 
`$values[0] ?? $limit`

src/Database/Adapter/MariaDB.php
```diff
@@ -770,23 +770,60 @@ public function deleteDocument(string $collection, string $id): bool
...
+                    break;
+
+                case Query::TYPE_OFFSET:
+                    $offset = $values[0];
```
christyjacob4, line 812: null-safety here as well `$values[0] ?? $offset`

src/Database/Adapter/MariaDB.php
```diff
@@ -770,23 +770,60 @@ public function deleteDocument(string $collection, string $id): bool
...
+                    break;
+
+                case Query::TYPE_CURSORAFTER:
+                case Query::TYPE_CURSORBEFORE:
+                    $cursor = $values[0];
```
christyjacob4, line 817: `$values[0] ?? $cursor`

src/Database/Adapter/MariaDB.php
```diff
@@ -770,23 +770,60 @@ public function deleteDocument(string $collection, string $id): bool
...
+                    $cursorDirection = $method == Query::TYPE_CURSORAFTER ? Database::CURSOR_AFTER : Database::CURSOR_BEFORE;
+                    break;
+
+                default:
+                    $filters[] = $query;
```
christyjacob4, line 822: Are all the filters queries simply appended to the list ? Don't they need to be parsed to their SQL forms ?  

```
const TYPE_EQUAL = 'equal';
    const TYPE_NOTEQUAL = 'notEqual';
    const TYPE_LESSER = 'lessThan';
    const TYPE_LESSEREQUAL = 'lessThanEqual';
    const TYPE_GREATER = 'greaterThan';
    const TYPE_GREATEREQUAL = 'greaterThanEqual';
    const TYPE_CONTAINS = 'contains';
    const TYPE_SEARCH = 'search';
```

stnguyen90: That's handled later on at https://github.com/utopia-php/database/blob/e48bdaa6ab40b79615673442b5fcdd27f6a4a769/src/Database/Adapter/MariaDB.php#L902

src/Database/Adapter/MariaDB.php
christyjacob4: We aren't reusing this logic elsewhere right? So I don't think we need to wrap it in a function at this point

src/Database/Adapter/MariaDB.php
christyjacob4: Okay got it 👍  Thanks

src/Database/Adapter/MariaDB.php
stnguyen90: We would probably also use it in the Mongo Adapter 🧐

src/Database/Adapter/MariaDB.php
stnguyen90: Good tip!

Also, how and where should we handle if there are multiple limit, offset, or cursor attributes in the queries array?

src/Database/Adapter/MariaDB.php
TorstenDittmann: Correct me if I am wrong, but shouldn't this whole block be part of the Database Class and not of the Adapter?

The MariaDB Adapter is taking care for SQL stuff and this has nothing todo with it.

I would actually keep the interface of the find method in the adapter as it was before 🤔 otherwise it makes the Adapters more complicated. wdyt @eldadfux ?

src/Database/Adapter/MariaDB.php
```diff
@@ -770,23 +770,60 @@ public function deleteDocument(string $collection, string $id): bool
...
+            $attribute = $query->getAttribute();
+            $values = $query->getValues();
+            switch ($method) {
+                case Query::TYPE_ORDERASC:
+                case Query::TYPE_ORDERDESC:
+                    if (!empty($attribute)) {
+                        $orderAttributes[] = $attribute;
+                    }
+
+                    $orderTypes[] = $method == Query::TYPE_ORDERASC ? Database::ORDER_ASC : Database::ORDER_DESC;
```
TorstenDittmann, line 804:
```suggestion
                    $orderTypes[] = $method === Query::TYPE_ORDERASC ? Database::ORDER_ASC : Database::ORDER_DESC;
```

in PHP I would do a strict equal everytime, otherwise you might get type juggling 👍🏻

TorstenDittmann: just as an example if you didn't know:
```php
var_dump("0000" == 0);      // bool(true)
var_dump("0e12345" == "0"); // bool(true)
var_dump("0e34983" == 0);   // bool(true)
```

ref: https://www.php.net/manual/en/language.types.type-juggling.php

src/Database/Adapter/MariaDB.php
stnguyen90: > Correct me if I am wrong, but shouldn't this whole block be part of the Database Class and not of the Adapter?

This makes sense to me especially if we're okay with keeping the existing signature for Adapter->find(). @eldadfux ?

src/Database/Adapter/MariaDB.php
stnguyen90: So I'm thinking about moving this to the Queries Validator because I think I'll need this logic in there too:

```php
    /**
     * Iterates through $queries and returns an array with:
     * - filters: array of filter queries
     * - limit: int
     * - offset: int
     * - orderAttributes: array of attribute keys
     * - orderTypes: array of Database::ORDER_ASC or Database::ORDER_DESC
     * - cursor: Document
     * - cursorDirection: Database::CURSOR_BEFORE or Database::CURSOR_AFTER
     * 
     * @param Query[] $queries
     * @param int $defaultLimit
     * @param int $defaultOffset
     * @param string $defaultCursorDirection
     * 
     * @return array
     */
    public static function byMethod(array $queries, int $defaultLimit = 25, int $defaultOffset = 0, string $defaultCursorDirection = Database::CURSOR_AFTER): array
```

and this use it like:

```php
        $queriesByMethod = Queries::byMethod($queries);
        /** @var Query[] */ $filters = $queriesByMethod['filters'];
        /** @var int */ $limit = $queriesByMethod['limit'];
        /** @var int */ $offset = $queriesByMethod['offset'];
        /** @var string[] */ $orderAttributes = $queriesByMethod['orderAttributes'];
        /** @var string[] */ $orderTypes = $queriesByMethod['orderTypes'];
        /** @var Document */ $cursor = $queriesByMethod['cursor'];
        /** @var string[] */ $cursorDirection = $queriesByMethod['cursorDirection'];
```

in Database->find() or Adapter->find().

Thoughts, @christyjacob4 and @TorstenDittmann ?

src/Database/Adapter/MariaDB.php
stnguyen90: Makes sense! Thanks for the tip! 🙏

src/Database/Adapter/MariaDB.php
stnguyen90: Updated! I also updated the processing of queries to only factor the 1st limit/offset/cursor encountered.

src/Database/Adapter/MariaDB.php
stnguyen90: updated!

src/Database/Adapter/MariaDB.php
stnguyen90: updated!

src/Database/Adapter/MariaDB.php
stnguyen90: updated!

src/Database/Adapter/MariaDB.php
eldadfux: It might be a good idea to keep the adapter signature as it was and change only the `Database` class signature.

src/Database/Validator/Queries.php
```diff
@@ -34,30 +39,32 @@ class Queries extends Validator
      * Queries constructor
      *
      * @param QueryValidator $validator
+     * @param Document[] $attributes
      * @param Document[] $indexes
      * @param bool $strict
      */
-    public function __construct($validator, $indexes, $strict = true)
+    public function __construct($validator, $attributes, $indexes, $strict = true)
```
fogelito, line 46: Does it make sense to pass the whole collection? 
instead of $attributes, $indexes

src/Database/Validator/Queries.php
```diff
@@ -184,4 +199,86 @@ protected function arrayMatch($indexes, $queries): bool
...
+     * - cursorDirection: Database::CURSOR_BEFORE or Database::CURSOR_AFTER
+     * 
+     * @param Query[] $queries
+     * @param int $defaultLimit
+     * @param int $defaultOffset
+     * @param string $defaultCursorDirection
+     * 
+     * @return array
+     */
+    public static function byMethod(array $queries, int $defaultLimit = 25, int $defaultOffset = 0, string $defaultCursorDirection = Database::CURSOR_AFTER): array
```
eldadfux, line 220: I don't think this method should handle default values. We should decouple this responsibility. I think we should take care of them when creating the query or in the consuming app (AKA Appwrite).

src/Database/Validator/QueryValidator.php
```diff
@@ -19,51 +19,45 @@ class QueryValidator extends Validator
...
+    protected int $maxOffset;
+    protected int $maxValuesCount;
 
     /**
      * Expression constructor
      *
      * @param Document[] $attributes
      */
-    public function __construct(array $attributes)
+    public function __construct(array $attributes, int $maxLimit = 100, int $maxOffset = 5000, int $maxValuesCount = 100)
```
eldadfux, line 31: Out of context, but I don't remember why we call this class `QueryValidator` over `Query`. Can we change it? The namespace is their to indicate this exact scope.

src/Database/Validator/QueryValidator.php
```diff
@@ -19,51 +19,45 @@ class QueryValidator extends Validator
...
-        'contains',
-        'search',
-    ];
+    protected int $maxLimit;
+    protected int $maxOffset;
+    protected int $maxValuesCount;
 
     /**
      * Expression constructor
      *
```
eldadfux, line 28: I think we should add a good chunk of docblock to explain the logic we follow.

src/Database/Database.php
```diff
@@ -1266,29 +1268,41 @@ public function deleteCachedDocument(string $collection, string $id): bool
...
      *
      * @return Document[]
      * @throws Exception
      */
-    public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], Document $cursor = null, string $cursorDirection = self::CURSOR_AFTER): array
+    public function find(string $collection, array $queries = []): array
     {
         $collection = $this->getCollection($collection);
 
+        $queriesByMethod = Queries::byMethod($queries);
```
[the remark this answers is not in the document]
stnguyen90: I'm definitely open to the name, however, I have a couple concerns about doing something like `getMethodByType($type);`

1. we don't really want queries by `Query::TYPE_*`, but more like query type groupings (filters, orders, limit, offset, curors)
2. I wanted to avoid iterating over the $queries array multiple times (ie. `getMethodByType(Query::TYPE_LIMIT)` and `getMethodByType(Query::TYPE_OFFSET)` would iterate over `$queries` twice).
3. I wanted to simplify what `Adapter->find()` needed. For example, given `$queries`, `Adapter->find()` needs `$orderAttributes` AND `$orderTypes`

src/Database/Validator/QueryValidator.php
stnguyen90: @eldadfux, the logic for if a Query is valid? Should that go in the docblock for `isValid()`?

src/Database/Database.php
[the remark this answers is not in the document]
eldadfux: We also usually try to keep the return values as close as possible to primitives or arrays of primitives for simplicity.
What about filtering them inside the database instance and saving the results once in a private attribute? That way we can keep the state, simplify the signature, and avoid extra iteration.

src/Database/Validator/QueryValidator.php
eldadfux: Yes.

src/Database/Database.php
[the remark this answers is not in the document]
stnguyen90: > What about filtering them inside the database instance and saving the results once in a private attribute? That way we can keep the state, simplify the signature, and avoid extra iteration.

Saving state into the Database instance makes me a little wary in case of any concurrency problems. In addition, this method is needed in the Queries Validator. TBD whether it needs to be used elsewhere in the appwrite repo too.

src/Database/Validator/Queries.php
stnguyen90: @eldadfux, the trouble with removing the defaults here is there is a possibility of it being null in `Database->find()`, but we kind of need it for `Adapter->find()`:

```php
        $queriesByMethod = Queries::byMethod($queries);
        /** @var Query[] */ $filters = $queriesByMethod['filters'];
        /** @var int */ $limit = $queriesByMethod['limit'];
        /** @var int */ $offset = $queriesByMethod['offset'];
        /** @var string[] */ $orderAttributes = $queriesByMethod['orderAttributes'];
        /** @var string[] */ $orderTypes = $queriesByMethod['orderTypes'];
        /** @var Document */ $cursor = $queriesByMethod['cursor'];
        /** @var string */ $cursorDirection = $queriesByMethod['cursorDirection'];

        if (!empty($cursor) && $cursor->getCollection() !== $collection->getId()) {
            throw new Exception("cursor Document must be from the same Collection.");
        }

        $cursor = empty($cursor) ? [] : $cursor->getArrayCopy();

        $queries = self::convertQueries($collection, $filters);

        $results = $this->adapter->find(
            $collection->getId(),
            $queries,
            $limit,
            $offset,
            $orderAttributes,
            $orderTypes,
            $cursor,
            $cursorDirection,
        );
```

stnguyen90: @eldadfux, unless you think it's okay to pass in null for some of these:

```php
        $results = $this->adapter->find(
            $collection->getId(),
            $queries,
            $limit,
            $offset,
            $orderAttributes,
            $orderTypes,
            $cursor,
            $cursorDirection,
        );
```

even though the default values for `Adapter->find()` won't be used in those cases.

src/Database/Validator/QueryValidator.php
stnguyen90: renamed!

src/Database/Validator/QueryValidator.php
stnguyen90: Added docblocks for both Query and Queries!

src/Database/Validator/Queries.php
stnguyen90: updated!

src/Database/Validator/Queries.php
eldadfux: I think we should handle the defaults when calling the adapter->find() method for keeping separation of concerns. Something like:
```php
$results = $this->adapter->find(
    $collection->getId(),
    $queries,
    $limit ?? 25,
    $offset ?? 0,
    ...
 );
```

src/Database/Database.php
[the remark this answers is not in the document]
eldadfux: As discussed, I think it should be fine concurrency, we can check with @TorstenDittmann later.

src/Database/Validator/Queries.php
eldadfux: Do we need to take care of the defaults in the validator that uses this method as well?

src/Database/Validator/Query.php
```diff
@@ -0,0 +1,202 @@
...
+            };
+
+            if (!$condition) {
+                $this->message = 'Query type does not match expected: ' . $attributeType;
+                return false;
+            }
+        }
+
+        // Contains method only supports array attributes
+        if (!$attributeSchema['array'] && $query->getMethod() === DatabaseQuery::TYPE_CONTAINS) {
```
eldadfux, line 172: I don't recall we publicly support this query type.

src/Database/Validator/Queries.php
```diff
@@ -78,50 +84,67 @@ public function getDescription(): string
     /**
      * Is valid.
      *
-     * Returns true if all $queries are valid as a set.
+     * Returns false if:
+     * 1. any query in $value is invalid based on $validator
+     * 
+     * In addition, if $strict is true, this returns false if:
+     * 1. there is no index with an exact match of the filters
+     * 2. there is no index with an exact mathc of the order attributes
```
eldadfux, line 92:
```suggestion
     * 2. there is no index with an exact match of the order attributes
```

src/Database/Validator/Queries.php
```diff
@@ -78,50 +84,67 @@ public function getDescription(): string
     /**
      * Is valid.
```
eldadfux, line 85: A few questions and comments:

- Are we preventing search on attributes with no proper fulltext index?
- We should also allow to limit the type of operators we allow to use. For example, future GEO operators will be disabled when listing users.
- Not sure if passing the collections is the best thing, for listing users, files and more we won't have easy access to it.
- Another problem with this, is that for users, files, etc we won't necessarily allow all attributes in the collections, but a set of very specific and predefined ones.

src/Database/Validator/Query.php
```diff
@@ -0,0 +1,202 @@
...
+    public function isValid($query): bool
+    {
+        // Validate method
+        $method = $query->getMethod();
+        if (!DatabaseQuery::isMethod($method)) {
+            $this->message = 'Query method invalid: ' . $method;
+            return false;
+        }
+
+        if ($method === DatabaseQuery::TYPE_LIMIT) {
```
eldadfux, line 105: I think it will be easier to maintain if we delegate each method check to a different protected method and use a switch case here.

eldadfux (changes requested):
Added a few comments, I think we should be close to a merge :)

src/Database/Validator/Query.php
stnguyen90: We don't. This is leftover from previous implementation: https://github.com/utopia-php/database/blob/0.19.0/src/Database/Validator/QueryValidator.php#L121-L124

src/Database/Validator/Queries.php
stnguyen90: @eldadfux,

> Are we preventing search on attributes with no proper fulltext index?

Yes, here:

https://github.com/utopia-php/database/blob/6dd00dd746395fd992a6a00d7abf96038e280eb8/src/Database/Validator/Queries.php#L136-L140

> We should also allow to limit the type of operators we allow to use. For example, future GEO operators will be disabled when listing users.

Is it okay to tackle that later? Will some sort of index be required for that type of operator? Perhaps we can validate that by checking query method against attribute type? 

> Not sure if passing the collections is the best thing, for listing users, files and more we won't have easy access to it.
Another problem with this, is that for users, files, etc we won't necessarily allow all attributes in the collections, but a set of very specific and predefined ones.

I'll change this back to take attributes and indexes.

> Another problem with this, is that for users, files, etc we won't necessarily allow all attributes in the collections, but a set of very specific and predefined ones.

If I change back to $attributes, we can pass in a set of attributes into the validator.

src/Database/Database.php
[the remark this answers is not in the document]
stnguyen90: In addition, this method is needed in the Queries Validator.

src/Database/Validator/Queries.php
stnguyen90: I have removed default values and hardcoded the defaults in `Database->find()`

src/Database/Validator/Queries.php
eldadfux: > Is it okay to tackle that later? Will some sort of index be required for that type of operator? Perhaps we can validate that by checking query method against attribute type?

Might be OK, I'll let you know if I think of a problem.

> If I change back to $attributes, we can pass in a set of attributes into the validator.

Are we passing only array of attributes names or array of attribute objects?

src/Database/Validator/Queries.php
```diff
@@ -184,4 +208,86 @@ protected function arrayMatch($indexes, $queries): bool
...
+     * - cursorDirection: Database::CURSOR_BEFORE or Database::CURSOR_AFTER
+     * 
+     * @param Query[] $queries
+     * @param int $defaultLimit
+     * @param int $defaultOffset
+     * @param string $defaultCursorDirection
+     * 
+     * @return array
+     */
+    public static function byMethod(array $queries): array
```
eldadfux, line 229: Maybe the name `group` or `groupByType` will better explain this logic now.

src/Database/Validator/Query.php
eldadfux: Should we remove it? It might be confusing to other team members, are we even using this anywhere?

stnguyen90: I think it's fine to keep as progress in adding support for it. The only thing left is adding support in the Adapter layer. 

If you want me to remove it, I can, but would you want to remove everything about it in Utopia\Database\Query?

src/Database/Validator/Queries.php
stnguyen90: Attribute Documents because OrderAttributes Validator required Attribute Documents.

src/Database/Validator/Queries.php
stnguyen90: 👍 I'll rename to `groupByType()`

src/Database/Validator/Query.php
eldadfux: No worries, let's tackle it later.

src/Database/Validator/Queries.php
eldadfux: We might need to extend this validator in Appwrite so we don't need to work hard to mock these attributes inside the controllers.

eldadfux (approved):

PR #165, opened 2022-08-15
Refactor permissions update

abnegate:
- Allow passing list of allowed permission types to permission validator

### DIFF
--- a/src/Database/Validator/Permissions.php
+++ b/src/Database/Validator/Permissions.php
@@ -9,12 +9,12 @@ class Permissions extends Validator
 {
     protected string $message = 'Permissions Error';
 
-    protected static array $defaultPermissions = [
+    protected static array $defaults = [
         ...Database::PERMISSIONS,
         'write',
     ];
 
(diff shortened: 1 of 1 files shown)


src/Database/Validator/Permissions.php
```diff
@@ -32,10 +33,16 @@ class Permissions extends Validator
      * Permissions constructor.
      *
      * @param int $length maximum amount of permissions. 0 means unlimited.
+     * @param array $allowedPermissions
      */
-    public function __construct(int $length = 0)
+    public function __construct(int $length = 0, array $allowedPermissions = [])
```
eldadfux, line 38:
```suggestion
    public function __construct(int $length = 0, array $allowed = [])
```

We're already under the permissions namespace. $allowed should be good enough

eldadfux: Same for the protected property.

eldadfux (changes requested):

eldadfux (approved):

PR #166, opened 2022-08-15
Datetime tz filter

christyjacob4:


src/Database/Database.php
```diff
@@ -221,7 +221,7 @@ function (?string $value) {
              * @return string|null
```
eldadfux, line 221: What is the default `date_default_timezone_get` for our server?

eldadfux (approved):

src/Database/Database.php
stnguyen90: From my experience, the appwrite docker images have the timezone set to UTC:

```shell
$ docker run appwrite/appwrite php -r 'print date_default_timezone_get() . "\n";'
UTC
```

fogelito: $app = new App('UTC');

tests/Database/Validator/DateTimeTest.php
```diff
@@ -44,6 +44,7 @@ public function testCreateDatetime()
         $this->assertEquals(true, DateTime::isValid("2022-12-04 11:31:52.680+02:00"));
         $this->assertEquals('UTC', date_default_timezone_get());
         $this->assertEquals("2022-12-04 09:31:52.680", DateTime::setTimezone("2022-12-04 11:31:52.680+02:00"));
+        $this->assertEquals("2022-12-04T09:31:52.681+00:00", DateTime::formatTz("2022-12-04 09:31:52.681"));
```
stnguyen90, line 47: I can confirm Dart and Javascript can parse this easily:

Dart: 

```dart
final nz = DateTime.parse("2022-12-04T09:31:52.681+00:00");
print(nz.toIso8601String()); // 2022-12-04T09:31:52.681Z
```

Javascript:

```javascript
d = new Date("2022-12-04T09:31:52.681+00:00");
console.log(d.toISOString()); // 2022-12-04T09:31:52.681Z
```

src/Database/DateTime.php
```diff
@@ -73,5 +74,22 @@ public static function setTimezone(string $datetime): string
...
+     */
+    public static function formatTz(?string $dbFormat): ?string
+    {
+        if (is_null($dbFormat)) return null;
+
+        try {
+            $value = new \DateTime($dbFormat);
+            return $value->format(self::$formatTz);
+        } catch (\Throwable $th) {
+            return $dbFormat;
```
stnguyen90, line 89: Can we get a test case for this?

src/Database/DateTime.php
```diff
@@ -6,7 +6,8 @@
 
 class DateTime
 {
-    protected static string $format = 'Y-m-d H:i:s.v';
+    protected static string $formatDb = 'Y-m-d H:i:s.v';
```
stnguyen90, line 9: This is probably okay for now, but we may want to figure out a way to handle this for different databases.

stnguyen90 (approved):
It would be nice to have an additional test case, but not a blocker.

tests/Database/Validator/DateTimeTest.php
fogelito: Thanks a lot

src/Database/DateTime.php
fogelito: There is one assertion test, most will be on Appwrite , since the filter happen there

src/Database/DateTime.php
fogelito: I agree

src/Database/Database.php
```diff
@@ -1429,18 +1430,51 @@ static public function addFilter(string $name, callable $encode, callable $decod
         ];
     }
 
+
+    /**
+     * @return array Document
+     * @throws Exception
+     */
+    public function getInternalAttributes(): array
```
eldadfux, line 1438: I think the more common way would be to just have a protected property named `attributes`, no need for a method. Definitely not a public one. When we do decide to add a public method, it should have tests.

eldadfux (changes requested):

src/Database/DateTime.php
fogelito: OK!

src/Database/Database.php
```diff
@@ -1429,18 +1430,51 @@ static public function addFilter(string $name, callable $encode, callable $decod
...
+                '$id' => ID::custom('$updatedAt'),
+                'type' => Database::VAR_DATETIME,
+                'format' => '',
+                'size' => 0,
+                'signed' => false,
+                'required' => false,
+                'default' => null,
+                'array' => false,
+                'filters' => ['datetime']]
+            )
```
christyjacob4, line 1461: Since we have 4 internal attributes, its better to keep all of them consistent and include the remaining in this array as well. 

`$id` and `$collection`

Please refer 
https://github.com/utopia-php/database/blob/479b4b0f84417253c4f749d731773b27c980b800/src/Database/Database.php#L1175-L1178

tests/Database/Base.php
```diff
@@ -2061,8 +2061,8 @@ public function testEncodeDecode()
         $this->assertEquals(false, $result->getAttribute('emailVerification'));
         $this->assertEquals(1, $result->getAttribute('status'));
         $this->assertEquals('randomhash', $result->getAttribute('password'));
-        $this->assertEquals('2000-06-12 14:12:55.000', $result->getAttribute('passwordUpdate'));
-        $this->assertEquals('1975-06-12 13:12:55.000', $result->getAttribute('registration'));
+        $this->assertEquals('2000-06-12T14:12:55.000+00:00', $result->getAttribute('passwordUpdate'));
+        $this->assertEquals('1975-06-12T13:12:55.000+00:00', $result->getAttribute('registration'));
```
christyjacob4, line 2065: Can we add more assertions to check the values of `$createdAt` and `$updatedAt` attributes ?

src/Database/Database.php
fogelito: I added:
protected array $attributes
I need the function still for to convert to list of Documents

src/Database/Database.php
fogelito: Sure! 
create : 
protected array $internals ....

src/Database/Database.php
```diff
@@ -93,6 +94,54 @@ class Database
         self::VAR_BOOLEAN => true,
     ];
 
+    /**
+     * List of Internal Ids
+     * @var array
+     */
+
```
eldadfux, line 101:
```suggestion
```

src/Database/Database.php
```diff
@@ -212,7 +261,7 @@ function (?string $value) {
                     $value = new \DateTime($value);
                     $value->setTimezone(new \DateTimeZone(date_default_timezone_get()));
                     return DateTime::format($value);
-                } catch (\Throwable $th) {
+                } catch (Throwable $th) {
```
eldadfux, line 264:
```suggestion
                } catch (\Throwable $th) {
```

christyjacob4 (commented):
Looks good 👍

src/Database/Database.php
christyjacob4, line 101:
```suggestion
```
My OCD wants to remove this line

src/Database/Database.php
eldadfux: php 8 don't allow us to set them in the protected attribute? 😒

christyjacob4: not really :(

src/Database/Database.php
```diff
@@ -1581,21 +1644,18 @@ public function casting(Document $collection, Document $document): Document
...
         }
 
-        try {
-            if (array_key_exists($name, $this->instanceFilters)) {
-                $value = $this->instanceFilters[$name]['encode']($value, $document, $this);
-            } else {
-                $value = self::$filters[$name]['encode']($value, $document, $this);
-            }
-        } catch (\Throwable $th) {
-            throw $th;
```
christyjacob4, line 1598: why did we remove try catch ?

src/Database/Database.php
eldadfux: Bummer.

eldadfux (approved):

PR #169, opened 2022-08-19
Feat: Allow null queries

Meldiron:
This PR tries to solve queries with null values failing, such as `Query.equal("name", null)`.

Marked as a draft because the solution sounds too simple to be correct 😅

- [x] New tests implemented

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -836,11 +836,11 @@ class MariaDB extends Adapter
                 }
 
                 $where[] = "(
-                        table_main.{$attribute} {$this->getSQLOperator($orderMethod, false)} :cursor 
+                        table_main.{$attribute} {$this->getSQLOperator($orderMethod, $cursor === null)} :cursor 
                         OR (
                             table_main.{$attribute} = :cursor 
                             AND
-                            table_main._id {$this->getSQLOperator($orderMethodInternalId, false)} {$cursor['$internalId']}
+                            table_main._id {$this->getSQLOperator($orderMethodInternalId, $cursor === null)} {$cursor['$internalId']}
                         )
                     )";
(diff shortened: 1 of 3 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -836,11 +836,11 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
                 }
 
                 $where[] = "(
-                        table_main.{$attribute} {$this->getSQLOperator($orderMethod)} :cursor 
+                        table_main.{$attribute} {$this->getSQLOperator($orderMethod, $cursor === null)} :cursor
```
eldadfux, line 839: Why do we call this param cursor? Feels out of context.

Meldiron: We call the param `isNull`. What should be passed inside is if the value of the SQL query is null or not. In this case we do `$cursor === null`, because $cursor is our value passed into SQL query later (thanks to `:cursor` part).

Meldiron: Should we keep it like this, or make a variable `$value = $cursor;` just so we can then have `$value === null`?

Alternatively, we could pass the whole value into the method as a named parameter `value: $cursor`, and do the check inside `getSQLOperator` method.

eldadfux: > Alternatively, we could pass the whole value into the method as a named parameter value: $cursor, and do the check inside getSQLOperator method.

I feel like this is the better approach.

Meldiron: Will do.

eldadfux:
Any chance this problem is already solved in #194 ?

abnegate:
Closing as this will be superseded by future features

PR #172, opened 2022-08-25
Refactor permissions

abnegate:
- Revert permissions checks to check document level
- **Remove authorisation check for document create**
- Implement aggregate functionality permission helper

We have verified that the removal of the authorisation check in `createDocument` should not have an impact in Appwrite, as all resource types that are publicly creatable, already have the necessary checks in place to ensure they can not be abused.

### DIFF
--- a/src/Database/Permission.php
+++ b/src/Database/Permission.php
@@ -104,12 +104,15 @@ class Permission
     /**
      * Map aggregate permissions into the set of individual permissions they represent.
      *
-     * @param array $permissions
+     * @param ?array $permissions
      * @param array $allowed
      * @return array
      */
-    public static function aggregate(array $permissions, array $allowed = Database::PERMISSIONS): array
+    public static function aggregate(?array $permissions, array $allowed = Database::PERMISSIONS): ?array
     {
+        if (\is_null($permissions)) {
+            return null;
+        }
         foreach ($permissions as $i => $permission) {
             $permission = Permission::parse($permission);
             foreach (self::$aggregates as $type => $subTypes) {
--- a/src/Database/Permission.php
+++ b/src/Database/Permission.php
@@ -6,6 +6,14 @@ class Permission
 {
     private Role $role;
 
+    private static $aggregates = [
+        'write' => [
+            Database::PERMISSION_CREATE,
+            Database::PERMISSION_UPDATE,
+            Database::PERMISSION_DELETE,
+        ]
+    ];
+
     public function __construct(
         private string $permission,
         string $role,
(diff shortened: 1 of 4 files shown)


christyjacob4 (approved):
Looks good 👍

tests/Database/Base.php
```diff
@@ -2111,10 +2111,6 @@ public function testReadPermissionsSuccess(Document $document)
 
     public function testReadPermissionsFailure()
     {
-        $this->expectException(ExceptionAuthorization::class);
-
-        Authorization::cleanRoles();
```
Meldiron, line 2116: Why did we re,oveclearRoles from the top? Are we expecting to get roles from a different test? Sounds scarry.
Could we clear roles  here at the top and set roles we need? For example `any` or `user:1` or whatever we need.

tests/Database/Base.php
```diff
@@ -2130,43 +2126,22 @@ public function testReadPermissionsFailure()
...
 
     /**
      * @depends testCreateDocument
      */
-    public function testWritePermissionsFailure(Document $document)
+    public function testWritePermissionsSuccess(Document $document)
     {
-        $this->expectException(ExceptionAuthorization::class);
-
-        Authorization::cleanRoles();
```
Meldiron, line 2168: Same as above, why did we remove this? Can we keep it? Should work the same, as creation should be allowed to everyone.

src/Database/Permission.php
```diff
@@ -93,6 +101,41 @@ public static function parse(string $permission): Permission
...
+                        continue;
+                    }
+                    $permissions[] = (new Permission(
+                        $subType,
+                        $permission->getRole(),
+                        $permission->getIdentifier(),
+                        $permission->getDimension()
+                    ))->toString();
+                }
+                unset($permissions[$i]);
```
Meldiron, line 133: I can imagine this to be hard to maintain with new changes (pushing and unsetting). Can we separate it? Use $permissions as read-only input, and have a new array for aggregated stuff.

src/Database/Permission.php
```diff
@@ -6,6 +6,14 @@ class Permission
 {
     private Role $role;
 
+    private static $aggregates = [
+        'write' => [
+            Database::PERMISSION_CREATE,
+            Database::PERMISSION_UPDATE,
+            Database::PERMISSION_DELETE,
+        ]
+    ];
```
Meldiron, line 15: Just a question, does this mean that `write` permission is supported? So I can do `createDocument(Permissions.write('any'))` and it would give create+update+delete permissions?

If so, can we have one more for `admin` that would allow everything? (read+write)

Meldiron (changes requested):

src/Database/Permission.php
abnegate: Correct, you can use write with SDK's or the API. Admin will be implemented in a future release

Meldiron (approved):

stnguyen90:
Closes: https://github.com/appwrite/appwrite/issues/4543

PR #174, opened 2022-09-01
Feat: toString() for queries

Meldiron:
In Appwrite, we currently test queries by passing string versions into HTTP requests, such as
```php
'queries' => [ 'limit(1)' ]
```

Thanks to introduction of `toString()`, we can now build actual query classes in Appwrite tests, such as
```php
'queries' => [ Query::limit(1)->toString() ]
```

That will improve code quality and will make sure no unexpected test breaks occur when internal change is done to queries syntax.

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -72,6 +72,40 @@ class Query
     }
 
     /**
+     * Create a query string from this Query instance
+     * This method is the opposide of 'parse' method.
+     *
+     * @return string
+     */
+    public function toString(): string
+    {
+        $method = static::getMethodFromAlias($this->getMethod());
+        $values = $this->getValues();
+        $attribute = $this->getAttribute();
+
+        switch ($method) {
+            case self::TYPE_EQUAL:
+            case self::TYPE_NOTEQUAL:
+            case self::TYPE_LESSER:
+            case self::TYPE_LESSEREQUAL:
+            case self::TYPE_GREATER:
+            case self::TYPE_GREATEREQUAL:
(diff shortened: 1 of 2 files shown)


abnegate (approved):

src/Database/Query.php
```diff
@@ -71,6 +71,40 @@ public function getValue($default = null)
         return $this->values[0] ?? $default;
     }
 
+    /**
+     * Create a query string from this Query instance
+     * This method is the opposide of 'parse' method.
+     *
+     * @return string
+     */
+    public function toString(): string
```
stnguyen90, line 80: Would it be better to implement [`__toString()`](https://www.php.net/manual/en/language.oop5.magic.php#object.tostring)?

eldadfux: @Meldiron I think this is a pretty cool suggestion, can we have the same thing for Role and Permission?

abnegate: `__toString()` would only be used in the case where the object is contextually expected to be a string, e.g.

```php
echo Query::limit(1);
```

For the case in the description, we would have to call the magic method:

```php
'queries' => [ Query::limit(1)->__toString() ]
```

I think it's cleaner to implement an instance method `toString()` instead

abnegate:
@Meldiron What's left to move this out of draft state?

Meldiron:
This is not relevant anymore, closing the PR.
It was relevant as it was supposed to improve code and tests maintainability of Appwrite. Hit some hiccups that would require more refactoring. The benefits no longer outweighed the downsides.

PR #176, opened 2022-09-04
Create update permissions in document

fogelito:
Add a new Internal attribute in Document _permissions for avoiding subqueries on getting Documents and lists

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -438,8 +438,7 @@ class MariaDB extends Adapter
         $document['$internalId'] = $document['_id'];
         $document['$createdAt'] = $document['_createdAt'];
         $document['$updatedAt'] = $document['_updatedAt'];
-        if(empty($document['_permissions']))$document['_permissions'] = '[]';
-        $document['$permissions'] = json_decode($document['_permissions'], true);
+        $document['$permissions'] = json_decode($document['_permissions'] ?? '[]', true);
 
         unset($document['_id']);
         unset($document['_uid']);
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -580,6 +580,7 @@ class MariaDB extends Adapter
         $attributes['_createdAt'] = $document->getCreatedAt();
         $attributes['_updatedAt'] = $document->getUpdatedAt();
         $attributes['_permissions'] = json_encode($document->getPermissions());
+
         $name = $this->filter($collection);
         $columns = '';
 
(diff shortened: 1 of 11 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -418,15 +419,10 @@ public function getDocument(string $collection, string $id): Document
...
         $stmt = $this->getPDO()->prepare("
-            SELECT 
-                table_main.*,
-                {$permissions}
-            FROM `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}` table_main
-            WHERE _uid = :_uid
-            LIMIT 1;
+            SELECT * 
+            FROM `{$this->getDefaultDatabase()}`.`{$this->getNamespace()}_{$name}` 
+            WHERE _uid = :_uid;
```
eldadfux, line 425: Why did we remove LIMIT 1? Isn't it important for performance?

fogelito: _uid is a unique field

eldadfux: Does this mean MySQL/MariaDB will automatically stop looking for more results?

eldadfux (approved):

src/Database/Adapter/MariaDB.php
```diff
@@ -442,25 +438,13 @@ public function getDocument(string $collection, string $id): Document
...
-            if (!empty($document[$key] ?? '')) {
-                $permissions = \json_decode($document[$key]);
-                foreach ($permissions as $permission) {
-                    $document['$permissions'][] = "{$type}(\"{$permission}\")";
-                }
-
-            }
-            unset($document[$key]);
-        }
+        $document['$permissions'] = json_decode($document['_permissions'], true);
```
TorstenDittmann, line 441: not sure if we need to make sure that `$document['_permissions']` actually contains a value. It might fail when its `NULL` or and empty string 🤔

fogelito: We can double check to force it back to an array...

if(empty($document['_permissions']))$document['_permissions'] = '[]';
?
@TorstenDittmann

src/Database/Adapter/MariaDB.php
```diff
@@ -442,25 +438,14 @@ public function getDocument(string $collection, string $id): Document
...
-                $permissions = \json_decode($document[$key]);
-                foreach ($permissions as $permission) {
-                    $document['$permissions'][] = "{$type}(\"{$permission}\")";
-                }
-
-            }
-            unset($document[$key]);
-        }
+        if(empty($document['_permissions']))$document['_permissions'] = '[]';
+        $document['$permissions'] = json_decode($document['_permissions'], true);
```
TorstenDittmann, line 442:
```suggestion
        $document['$permissions'] = json_decode($document['_permissions'] ?? '[]', true);
```

TorstenDittmann (approved):

PR #177, opened 2022-09-05
Refactor permissions valdiators

abnegate:
- Remove regex matching permissions/roles
- Return specific error messages for invalid permissions/roles
- Improve permissions/roles string parsing
- Make status a dimension instead of role for consistency
  - `status:verified`-> `users/verified` 
- Add optional status param to `user` and `users` role helpers
- Remove all legacy roles to avoid future migrations

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -53,15 +53,19 @@ class Database
         self::PERMISSION_DELETE,
     ];
 
+    // Roles
+    const ROLE_ANY = 'any';
+    const ROLE_GUESTS = 'guests';
+    const ROLE_USERS = 'users';
+    const ROLE_USER = 'user';
+    const ROLE_TEAM = 'team';
+
     const ROLES = [
-        'any',
-        'users',
-        'user',
-        'team',
-        'member',
-        'guests',
-        'status',
-        'role',
+        self::ROLE_ANY,
+        self::ROLE_GUESTS,
+        self::ROLE_USERS,
+        self::ROLE_USER,
+        self::ROLE_TEAM,
     ];
 
     // Collections
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -70,6 +70,15 @@ class Database
         self::ROLE_MEMBER,
     ];
 
+    // Dimensions
+    const DIMENSION_VERIFIED = 'verified';
+    const DIMENSION_UNVERIFIED = 'unverified';
+
+    const USER_DIMENSIONS = [
+        self::DIMENSION_VERIFIED,
+        self::DIMENSION_UNVERIFIED,
+    ];
+
     // Collections
     const METADATA = '_metadata';
 
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -59,6 +59,7 @@ class Database
     const ROLE_USERS = 'users';
     const ROLE_USER = 'user';
     const ROLE_TEAM = 'team';
+    const ROLE_MEMBER = 'member';
 
     const ROLES = [
         self::ROLE_ANY,
@@ -66,6 +67,7 @@ class Database
         self::ROLE_USERS,
         self::ROLE_USER,
         self::ROLE_TEAM,
+        self::ROLE_MEMBER,
     ];
 
     // Collections
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -70,6 +70,71 @@ class Database
         self::ROLE_MEMBER,
     ];
 
+    const ROLE_CONFIGURATION = [
+        Database::ROLE_ANY => [
+            'identifier' => [
+                'allowed' => false,
+                'required' => false,
+            ],
+            'dimension' =>[
+                'allowed' => false,
+                'required' => false,
+            ],
+        ],
+        Database::ROLE_GUESTS => [
+            'identifier' => [
+                'allowed' => false,
+                'required' => false,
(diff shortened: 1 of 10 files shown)


src/Database/Database.php
```diff
@@ -53,15 +53,19 @@ class Database
...
-        'team',
-        'member',
-        'guests',
-        'status',
-        'role',
+        self::ROLE_ANY,
+        self::ROLE_GUESTS,
+        self::ROLE_USERS,
+        self::ROLE_USER,
+        self::ROLE_TEAM,
```
eldadfux, line 68: Missing member.

src/Database/Permission.php
```diff
@@ -72,33 +72,52 @@ public function getDimension(): string
...
-
-        return new Permission(
-            $permission,
-            $role,
-            $identifier,
-            $dimension,
-        );
+
+        if (!$hasIdentifier && $hasDimension) {
+            $parts = \explode('/', $fullRole);
```
eldadfux, line 102: We're missing some null safety after explode.

src/Database/Validator/Permissions.php
```diff
@@ -72,127 +58,105 @@ public function isValid($permissions): bool
...
-                    default:
-                        $this->message = 'Permission must begin with one of: ' . \implode(", ", $permissions);
+            switch ($role) {
+                case Database::ROLE_USERS:
+                    if (!empty($identifier)) {
+                        $this->message = 'Role "' . $role . '"' . ' can not have an ID value.';
                         return false;
-                }
+                    }
+                    if (!empty($dimension) && !\in_array($dimension, $this->userDimensions)) {
```
eldadfux, line 113: Could have been a bit simpler if we had a config that tells us which characteristics each role can have.

Something like:

```php
'any' => [
  'id' => true,
  'dimension' => false
]
``` 

What do you think?

src/Database/Validator/Permissions.php
```diff
@@ -72,127 +58,105 @@ public function isValid($permissions): bool
...
+                        $this->message = 'Users dimension "' . $dimension . '" is not allowed. Must be one of: ' . \implode(', ', $this->userDimensions);
+                        return false;
+                    }
+                    break;
+                case Database::ROLE_GUESTS:
+                case Database::ROLE_ANY:
```
eldadfux, line 119: Are we replicating the same logic we have in the Role validator? Could we reuse some of it?

src/Database/Validator/Permissions.php
abnegate: I think it may make it more complex as we would need a config schema closer to:

```php
    [
        'user' => [
            'id' => [
                'allowed' => true,
                'required' => true,
            ],
            'dimension' => [
                'allowed' => true,
                'required' => false,
                'options' => [
                    'verified',
                    'unverified',
                ],
            ],
        ],
    ]
```

And reading through a lot of config `isset` and `if` checks might be harder than the switch cases

src/Database/Permission.php
abnegate: The following `count` check should be sufficient as `explode` will always return an array in this case

src/Database/Validator/Permissions.php
eldadfux: I feel the current implementation seems very error prone, and depends too much on the maintainer focus. Can we try to start with a config array and iterate again if it goes out of control?

eldadfux (approved):

src/Database/Permission.php
```diff
@@ -72,33 +72,52 @@ public function getDimension(): string
      *
      * @param string $permission
      * @return Permission
+     * @throws \Exception
      */
     public static function parse(string $permission): Permission
```
TorstenDittmann, line 77: Nothing improtant, would it make sense to call this method `fromString`?

src/Database/Permission.php
```diff
@@ -72,33 +72,52 @@ public function getDimension(): string
...
         $parts = \explode('("', $permission);
+
+        if (\count($parts) !== 2) {
+            throw new \Exception('Invalid permission string format: "' . $permission . '".');
+        }
+
         $permission = $parts[0];
         $fullRole = \str_replace('")', '', $parts[1]);
         $parts = \explode(':', $fullRole);
+        $role = $parts[0];
```
TorstenDittmann, line 88: This might get confusing with the naming here, since 4 lines before we have `$permission = $parts[0];` and now again.

We might wanna use another variable for the second `$parts`

src/Database/Role.php
```diff
@@ -58,50 +58,66 @@ public function getDimension(): string
...
+            return new Role($role, $parts[1]);
         }
-        $role = $parts[0];
-        $fullIdentifier = $parts[1];
-        $parts = \explode('/', $fullIdentifier);
 
-        if (\count($parts) === 1) {
-            return new Role($role, $fullIdentifier);
+        if (!$hasIdentifier && $hasDimension) {
+            $parts = \explode('/', $role);
```
TorstenDittmann, line 79: Same here, re-using the same variable might create confusion.

src/Database/Permission.php
abnegate: It initially was 😅 it was changed to `parse` to be consistent with the `Query` class

TorstenDittmann (approved):

christyjacob4 (approved):

PR #181, opened 2022-09-06
Fix: datetime default values

Meldiron:
Library was not forcing datetime filter on datetime-type attributes, even tho it is necessary for proper conversion on adapter level. This PR forces the adapter internally, and introduces tests to ensure the use case works.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -573,6 +573,10 @@ class Database
             }
         }
 
+        if($type === self::VAR_DATETIME && !\array_key_exists('datetime', $filters)) {
+            $filters[] = 'datetime';
+        }
+
         $collection->setAttribute('attributes', new Document([
             '$id' => ID::custom($id),
             'key' => $id,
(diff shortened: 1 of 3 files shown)


src/Database/Database.php
```diff
@@ -785,7 +789,7 @@ public function updateAttributeDefault(string $collection, string $id, $default
                 case self::VAR_INTEGER:
                 case self::VAR_FLOAT:
                 case self::VAR_BOOLEAN:
-                    if ($attribute->getAttribute('type') !== \gettype($default)) {
+                        if ($attribute->getAttribute('type') !== \gettype($default) && $attribute->getAttribute('type') !== self::VAR_DATETIME) {
```
TorstenDittmann, line 792:
```suggestion
                    if ($attribute->getAttribute('type') !== \gettype($default) && $attribute->getAttribute('type') !== self::VAR_DATETIME) {
```

src/Database/Database.php
```diff
@@ -635,7 +639,7 @@ public function createAttribute(string $collection, string $id, string $type, in
                 case self::VAR_INTEGER:
                 case self::VAR_FLOAT:
                 case self::VAR_BOOLEAN:
-                    if ($type !== \gettype($default)) {
+                    if ($type !== \gettype($default) && $type !== self::VAR_DATETIME) {
```
TorstenDittmann, line 642: I feel like the condition for datetime should only apply for the case of a string right?

Otherwise its hard to follow why the condition is there.

Meldiron:
Problem addressed in a better way in this PR: 
https://github.com/utopia-php/database/pull/185

PR #182, opened 2022-09-08
Fix: updateAttribute for Datetime

Meldiron:
- [x] Implement feature
- [x] Implement new tests

### DIFF
--- a/tests/Database/Base.php
+++ b/tests/Database/Base.php
@@ -2726,8 +2726,10 @@ abstract class Base extends TestCase
         $flowers = $database->createCollection('flowers');
         $database->createAttribute('flowers', 'name', Database::VAR_STRING, 128, true);
         $database->createAttribute('flowers', 'inStock', Database::VAR_INTEGER, 0, false);
+        $database->createAttribute('flowers', 'date', Database::VAR_STRING, 128, false);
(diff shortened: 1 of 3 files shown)


tests/Database/Base.php
```diff
@@ -2890,6 +2893,7 @@ public function testUpdateAttributeStructure()
         $this->assertEquals(500, $doc->getAttribute('price'));
 
         $database->updateAttribute('flowers', 'price', Database::VAR_STRING, 255, false, false);
+        $database->updateAttribute('flowers', 'date', Database::VAR_DATETIME, 0, false);
 
         // Delete cache to force read from database with new schema
         $database->deleteCachedDocument('flowers', 'LiliPriced');
```
fogelito, line 2899: Do we have a clean cache on  all 'flowers' Documents after the update?

Meldiron: We should, thanks to code right below this `$database->deleteCachedDocument('flowers', 'LiliPriced');`. We need to do this manually, since `updateAttribute` can't know if cache is still valid or not after changes.

TorstenDittmann (approved):

christyjacob4 (approved):

PR #185, opened 2022-09-12
feat: update datetime filters issue

christyjacob4:


src/Database/Database.php
```diff
@@ -573,10 +573,6 @@ public function createAttribute(string $collection, string $id, string $type, in
             }
         }
 
-        if($type === self::VAR_DATETIME && !\array_key_exists('datetime', $filters)) {
-            $filters[] = 'datetime';
-        }
-
```
TorstenDittmann: any reason for this removal?

if so, we should add a test on what happens when an attribute is create that does not have the datetime filter. because it is required and under any circumstance should throw an error :+1:

src/Database/Database.php
```diff
@@ -654,6 +634,51 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+        switch ($type) {
+            case self::VAR_STRING:
+            case self::VAR_INTEGER:
+            case self::VAR_FLOAT:
+            case self::VAR_BOOLEAN:
+                if ($type !== $defaultType) {
+                    throw new Exception('Default value ' . $default . ' does not match given type ' . $type);
+                }
+                break;
+            case self::VAR_DATETIME:
```
christyjacob4, line 671: This is the only check for datetime. To ensure that datetime attribute allows a default value of type string

tests/Database/Base.php
```diff
@@ -142,9 +142,10 @@ public function testCreateDeleteAttribute()
         $this->assertEquals(true, static::getDatabase()->createAttribute('attributes', 'integer_default', Database::VAR_INTEGER, 0, false, 1));
         $this->assertEquals(true, static::getDatabase()->createAttribute('attributes', 'float_default', Database::VAR_FLOAT, 0, false, 1.5));
         $this->assertEquals(true, static::getDatabase()->createAttribute('attributes', 'boolean_default', Database::VAR_BOOLEAN, 0, false, false));
+        $this->assertEquals(true, static::getDatabase()->createAttribute('attributes', 'datetime_default', Database::VAR_DATETIME, 0, false, '2000-06-12T14:12:55.000+00:00', true, false, null, [], ['datetime']));
```
christyjacob4, line 145: When we create a datetime attribute, we should pass the filter explicitly..

src/Database/Database.php
christyjacob4: So the `createAttribute` function has a parameter called `$filters` 

When we create a new attribute, it becomes the responsibility of the caller to pass in the relevant filters. I've updated the tests to ensure we pass in the  relevant filter when creating a datetime attribute

TorstenDittmann: yeah, but is it possible to create that attribute without the filter?

TorstenDittmann: In the end, you should be able to use the library and if no exception is thrown - means everything worked fine.

This will not be true in this case. So we either add the filter if its datetime or we throw an error if its missing.

christyjacob4: Yes, it will allow you to create the attribute, however there will be an exception when you try to create a document since the filter is not applied.

christyjacob4: I can think of 2 ways to make it more generic

1. We make `$filters` a required parameter instead of defaulting to `null`. This way, the caller must at least pass an empty array.
2. The second approach is to maintain a mapping of data types to their default filters
```php
$defaultFilters = [
   self::VAR_DATETIME => ['datetime'],
   self::VAR_STRING => [],
   .... 
];
```

And apply these filters in create attribute.. 


Even if we have to throw an exception or just use the default values, we need an array mapping like this to know which filters are required for which data types. 

LMK your thoughts...

TorstenDittmann: 1. will still make the attribute not work without passing it
2. can mess up adding additional filters where they order might matter

We should probably have a map of required filters and check. If missing we throw an exception 👍🏻

TorstenDittmann (commented):
I think if we add tests for required attributes it should be good 👍🏻

TorstenDittmann (approved):

Meldiron:
This closes https://github.com/utopia-php/database/pull/181 I believe

PR #186, opened 2022-09-14
Cache and reset to intial auth status in skips

abnegate:
- Fixes nested skips resetting status before the callback is completed

### DIFF
--- a/src/Database/Validator/Authorization.php
+++ b/src/Database/Validator/Authorization.php
@@ -158,9 +158,13 @@ class Authorization extends Validator
      */
     public static function skip(callable $callback)
     {
+        $initialStatus = self::$status;
+
         self::disable();
+
         $result = $callback();
-        self::reset();
+
+        self::$status = $initialStatus;
 
         return $result;
     }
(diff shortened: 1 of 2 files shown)


src/Database/Validator/Authorization.php
```diff
@@ -158,9 +158,13 @@ public static function setDefaultStatus(bool $status): void
...
     public static function skip(callable $callback)
     {
+        $initialStatus = self::$status;
+
         self::disable();
+
         $result = $callback();
-        self::reset();
+
+        self::$status = $initialStatus;
```
stnguyen90, line 167: This is okay for now, but what if there are more than 2 nested skips? 👀 

Can we get a test case for this?

stnguyen90 (changes requested):

src/Database/Validator/Authorization.php
abnegate: Since it is always setting status back to what it was before the callback executes, it shouldn't matter how many levels deep it goes. Will add a test case now 👍

stnguyen90 (changes requested):
LGTM! Just pending successful build 🔨

stnguyen90 (approved):

PR #189, opened 2022-09-19
Fix: Improve count

Meldiron:
Currently we do this filtering of queries in before running count() method in Appwrite. Problem is, we have a lot of duplicate code because of that. Also there is no reason why you would need non-filter (order, for example) queries when counting.

Addresses: https://github.com/appwrite/appwrite/pull/3702#discussion_r958287285

- [x] Existing tests must pass

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1520,6 +1520,7 @@ class Database
             throw new Exception("Collection not found");
         }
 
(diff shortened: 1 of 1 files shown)


Meldiron:
Fun fact: It even fails with `Message: Unknown method:limit` if you try to put non-filter queries into `count` method.

eldadfux (approved):

PR #194, opened 2022-09-23
Fix bugs where Query parser ignored 0 strings

stnguyen90:
When `Query::parse()` parsed a string like `equal("attr", [0])` it ignored the 0 string so the values was an empty array. This caused a 500 error in Appwrite.

```
appwrite  | array(1) {
appwrite  |   [0]=>
appwrite  |   object(Utopia\Database\Query)#1556 (3) {
appwrite  |     ["method":protected]=>
appwrite  |     string(5) "equal"
appwrite  |     ["attribute":protected]=>
appwrite  |     string(7) "intAttr"
appwrite  |     ["values":protected]=>
appwrite  |     array(0) {
appwrite  |     }
appwrite  |   }
appwrite  | }
appwrite  | [Error] Type: Utopia\Exception
appwrite  | [Error] Message: Error handler had an error
appwrite  | [Error] File: /usr/src/code/vendor/utopia-php/framework/src/App.php
```


src/Database/Query.php
```diff
@@ -251,7 +251,9 @@ public static function parse(string $filter): self
                     } else {
                         // Append from parap builder. Either value, or array
                         if (empty($currentArrayParam)) {
-                            if (!empty($currentParam)) {
+                            // if (!empty($currentParam)) {
```
eldadfux, line 254: leftover

eldadfux (approved):
Looks good, left one small comment. 👍

src/Database/Query.php
```diff
@@ -251,7 +251,8 @@ public static function parse(string $filter): self
                     } else {
                         // Append from parap builder. Either value, or array
                         if (empty($currentArrayParam)) {
-                            if (!empty($currentParam)) {
+                            if (strlen($currentParam)) {
+                                var_dump($currentParam);
```
christyjacob4, line 255: leftovers

tests/Database/QueryTest.php
```diff
@@ -148,6 +148,21 @@ public function testParseV2()
         $this->assertEquals("attr", $query->getAttribute());
         $this->assertEquals([1], $query->getValues());
 
+        $query = Query::parse('equal("attr", [0])');
+        $this->assertCount(1, $query->getValues());
+        $this->assertEquals("attr", $query->getAttribute());
+        $this->assertEquals([0], $query->getValues());
+
+        $query = Query::parse('equal("attr", 0,)');
```
christyjacob4, line 156: Is this supposed to be a valid input ? with a trailing comma after the zero? 
Shouldn't this throw an error

stnguyen90: We're lenient and allow it like how some languages allow trailing commas.

src/Database/Query.php
stnguyen90: oye such shame! 🫣 Sorry about that. I've removed it now.

christyjacob4: No worries! Happens to the best of us!

PR #107, opened 2022-02-04
Add Postgres Adapter

vermakhushboo:
### Work in progress

**Edge cases in Appwrite but not in Utopia**

- Had to change the exception type to ‘Throwable’ instead of ‘PDOException’ for Appwrite as the exception wasn’t getting captured in createDocument and updateDocument methods
TODO: 
   - Add more tests that throw exceptions not caught by PDOException
   - Find places where we throw the wrong errors
   - Rethrow exceptions that Appwrite can understand

- There are not enough tests in Utopia to verify that double-quotes were needed around {$attribute} in the find() method to handle reserved words
TODO:
   - Find specific keywords that break this method
   - Automate iteration to test for keywords 
   - Ex: https://www.drupal.org/docs/develop/coding-standards/list-of-sql-reserved-words#s-reserved-words 
   - Find more reserved words for different databases, example, https://en.wikipedia.org/wiki/SQL_reserved_words

- Had to remove all alphanumeric values in getSQLCondition() method. Need to add better edge cases for full-text search in Utopia, it is missing many cases
TODO:
  - Add more tests to cover similar scenarios in Utopia 
  - For example email ids etc
  - Add tests specific to databases

**Improvements needed in Appwrite for future adapters**
TODO: 
- Hardcoded DB names in init.php, realtime.php, worker.php. Create generic interface for all adapters, allowing configurations using env variables
- Add health check method to each adapter
- Had to create a file PostgresSQLStatement.php and override Swoole methods, potentially move it to Utopia in the future

PR #195, opened 2022-09-30
Feat postgress adapter (eldad)

eldadfux:


src/Database/Adapter/Postgres.php
```diff
@@ -1,129 +1,1160 @@
...
+        if (!empty($permissions)) {
+            $queryPermissions = "INSERT INTO {$this->getSQLTable($name.'_perms')}
+            (_type, _permission, _document) VALUES " . implode(', ', $permissions);
+            $stmtPermissions = $this->getPDO()->prepare($queryPermissions);
+        }
+
+        try {
+            $stmt->execute();
+
+            $document['$internalId'] = $this->getDocument($collection, $document->getId())->getInternalId();
```
fogelito, line 375: We can get last insert using native RETURNING id? We better not use the get Document which can be a very large doc for this?

src/Database/Adapter/Postgres.php
```diff
@@ -1,129 +1,1160 @@
...
+
+            if (isset($stmtPermissions)) {
+                $stmtPermissions->execute();
+            }
+        } catch (Throwable $e) {
+            switch ($e->getCode()) {
+                case 1062:
```
fogelito, line 382: Are this values same as mariadb?

src/Database/Adapter/Postgres.php
```diff
@@ -1,129 +1,1160 @@
...
+            foreach ($additions as $type => $permissions) {
+                foreach ($permissions as $i => $_) {
+                    $values[] = "( :_uid, '{$type}', :_add_{$type}_{$i} )";
+                }
+            }
+
+            $stmtAddPermissions = $this->getPDO()
+                ->prepare(
+                    "INSERT INTO {$this->getSQLTable($name.'_perms')}
+                    (_document, _type, _permission) VALUES" . \implode(', ', $values)
```
fogelito, line 515: General comment , we should use Internal id's instead of _document

src/Database/Adapter/Postgres.php
```diff
@@ -1,129 +1,1160 @@
...
+     * @param Query[] $queries
+     * @param int $max
+     *
+     * @return int|float
+     */
+    public function sum(string $collection, string $attribute, array $queries = [], int $max = 0): int|float
+    {
+        $name = $this->filter($collection);
+        $roles = Authorization::getRoles();
+        $where = ['1=1'];
```
fogelito, line 889: Better not using 1=1 I do not think there is a performance issue but it looks like sql injection :)

abnegate:
Closing as changes here will be merged into the main PR https://github.com/utopia-php/database/pull/107

PR #201, opened 2022-10-12
add event, triggers and listeners support

lohanidamodar:
- Introduces events and listeners on the database functions
- Each function will generate relevant events
- All event constants are added in the Database class
- Relevant tests added in the base test
- Revert `metadata` back to `_metadata`

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -83,6 +83,8 @@ class Database
     const EVENT_DOCUMENT_READ = 'document_read';
     const EVENT_DOCUMENT_UPDATE = 'document_update';
     const EVENT_DOCUMENT_DELETE = 'document_delete';
+    const EVENT_DOCUMENT_COUNT = 'document_count';
+    const EVENT_DOCUMENT_SUM = 'document_sum';
 
     const EVENT_ATTRIBUTE_CREATE = 'attribute_create';
     const EVENT_ATTRIBUTE_UPDATE = 'attribute_update';
@@ -1527,7 +1529,9 @@ class Database
         $queries = Query::groupByType($queries)['filters'];
         $queries = self::convertQueries($collection, $queries);
 
-        return $this->adapter->count($collection->getId(), $queries, $max);
+        $count = $this->adapter->count($collection->getId(), $queries, $max);
+        $this->trigger(self::EVENT_DOCUMENT_COUNT, $count);
+        return $count;
     }
 
     /**
@@ -1552,7 +1556,9 @@ class Database
         }
 
         $queries = self::convertQueries($collection, $queries);
-        return $this->adapter->sum($collection->getId(), $attribute, $queries, $max);
+        $sum = $this->adapter->sum($collection->getId(), $attribute, $queries, $max);
+        $this->trigger(self::EVENT_DOCUMENT_SUM, $sum);
+        return $sum;
     }
 
     /**
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -322,7 +322,13 @@ class Database
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -65,6 +65,15 @@ class Database
     // Cache
     const TTL = 60 * 60 * 24; // 24 hours
 
+    // Events
+    const EVENT_CREATE_DATABASE = 'create_database';
```
eldadfux, line 69:
```suggestion
    const EVENT_DATABASE_CREATE = 'database_create';
```

Makes more sense to have the namespace before the action

src/Database/Database.php
```diff
@@ -265,6 +281,40 @@ function (?string $value) {
...
+    /**
+     * Trigger callback for events
+     *
+     * @param string $event
+     * @param array|null $args
+     * @return void
+     */
+    protected function trigger(string $event, mixed $args = null): void
+    {
+        foreach ($this->listeners['*'] as $callback) {
```
eldadfux, line 309:
```suggestion
        foreach ($this->listeners['*'] as $callback) {
```

We can have a constant for all events as well

src/Database/Database.php
```diff
@@ -502,7 +557,9 @@ public function deleteCollection(string $id): bool
     {
         $this->adapter->deleteCollection($id);
 
-        return $this->deleteDocument(self::METADATA, $id);
+        $deleted = $this->deleteDocument(self::METADATA, $id);
+        $this->trigger(self::EVENT_DELETE_COLLECTION, ['id' => $id, 'deleted' => $deleted]);
```
eldadfux, line 561: Can we be more consistent with the payload we're passing? deleted can be just named as document or data? Do we need to specifically pass `id` and `collections`? Don't we already have them in the deleted/created/updated document/collection/database

eldadfux (changes requested):

src/Database/Database.php
lohanidamodar: updated

eldadfux (commented):
All looks good, I would add some unit tests to check all events are executed.

src/Database/Database.php
```diff
@@ -265,6 +306,49 @@ function (?string $value) {
...
+    public function on(string $event, callable $callback): self
+    {
+        if(!isset($this->listeners[$event])) {
+            $this->listeners[$event] = [];
+        }
+        $this->listeners[$event][] = $callback;
+        return $this;
+    }
+
+    public function silent(callable $callback) {
```
eldadfux, line 325: Let's add a docblock (I know...), can we also have a return type?

tests/Database/Base.php
```diff
@@ -3180,4 +3180,77 @@ public function testWritePermissions()
         $this->assertEquals('cat', $docs[0]['$id']);
         $this->assertEquals('newCat', $docs[0]['type']);
     }
+
+    public function testEvents()
+    {
```
eldadfux, line 3185: Does this tests validate all events have actually been triggered? Can we know if an event was triggered more than once? It will be important to make sure no future maintainers forget to silent internal calls.

lohanidamodar: Yes this test does validate events actually been triggered and this test will fail if event was triggered more than once

src/Database/Database.php
lohanidamodar: Return type will depend on the callback as it returns whatever the callback returns so? Auth::skip doesn't have one as well? Can we add mixed?

eldadfux: Not sure from which version mixed is supported, please check.

eldadfux (approved):

PR #207, opened 2022-10-15
Added ping method

eldadfux:
Added `ping` method to test the database and adapter connections.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -33,6 +33,20 @@ class MariaDB extends Adapter
     }
 
     /**
+     * Ping Database
+     *
+     * @return bool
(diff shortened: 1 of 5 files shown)


lohanidamodar (approved):
Looks good.

src/Database/Adapter/MariaDB.php
```diff
@@ -32,6 +32,20 @@ public function __construct($pdo)
...
     }
 
+    /**
+     * Ping Database
+     *
+     * @return bool
+     * @throws Exception
+     * @throws PDOException
+     */
+    public function ping(): bool
```
fogelito, line 42: If there is a connection issue this will never happen, an exception will be thrown in the connection function.
On the health check, it is essential to put a short timeout.

PR #221, opened 2022-11-24
Fix benchmarks

wess:
Updates benchmarks to actually work.


bin/tasks/query.php
```diff
@@ -133,23 +134,33 @@
...
 
     // Popular posts
-    $query = ["views.greater(100000)"];
+    $query = ["greaterThan('views', 100000)"];
+
     $results[] = runQuery($query, $database, $limit);
 
     // Fulltext search
-    $query = ["text.search('Alice')"];
+    $query = ["search('text', 'Alice')"];
```
TorstenDittmann, line 163: We can use the Query class here 👍🏻

abnegate (approved):

PR #222, opened 2022-12-12
Feat technical debt

fogelito:
Addresses some technical debt:

- Move ID, Role, Query, Permission to src/Database/Helpers
- Fix DateTime valdiator pattern
- Update getRowLimit to getDocumentSize
- Remove switching on adapter name in tests
- Split up find test
- Update supported DB version on README


src/Database/Adapter/SQLite.php
```diff
@@ -24,7 +24,7 @@
  * 9. MODIFY COLUMN is not supported
  * 10. Can't rename an index directly
  */
-class SQLite extends MySQL
+class SQLite extends MariaDB
```
abnegate, line 27: What's the reasoning for this change?

tests/Database/Base.php
```diff
@@ -32,14 +34,12 @@ abstract static protected function getDatabase(): Database;
      */
     abstract static protected function getAdapterName(): string;
 
-    /**
-     * @return int
-     */
-    abstract static protected function getAdapterRowLimit(): int;
+    static public Validator $dateValidator;
```
abnegate, line 37: We're only using this in one test, let's create one new instance there instead of having the static property

tests/Database/Validator/DateTimeTest.php
```diff
@@ -4,11 +4,16 @@
 
 use Utopia\Database\DateTime;
 use PHPUnit\Framework\TestCase;
+use Utopia\Database\Validator\DatetimeValidator;
+use Utopia\Validator;
 
 class DateTimeTest extends TestCase
 {
+    static public Validator $dateValidator;
```
abnegate, line 12: We're only using this in one test, let's create one new instance there instead of having the static property

abnegate (commented):
Looking great! The find method is SO much better :grin: just a few small comments

src/Database/Adapter/SQLite.php
fogelito: The base Class is MariaDB not MySQL , I asked Eldad about this in the passes he said no reason , he just wants tests to pass.
MySQL extends MariaDB as well.
SQLite should to ?

tests/Database/Base.php
fogelito: cool!

tests/Database/Validator/DateTimeTest.php
fogelito: cool!

src/Database/Adapter/SQLite.php
abnegate: Cool makes sense 👌

abnegate (approved):

eldadfux (approved):

PR #223, opened 2022-12-15
Replaced group by with distinct

vermakhushboo:
After the permission-related changes in the database, the join queries in Appwrite and Utopia have been replaced with subqueries, thus eliminating the need for GROUP BY.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -894,10 +894,9 @@ class MariaDB extends Adapter
         $sqlWhere = !empty($where) ? 'where ' . implode(' AND ', $where) : '';
 
         $sql = "
-            SELECT table_main.*
+            SELECT DISTINCT _uid, table_main.*
             FROM {$this->getSQLTable($name)} as table_main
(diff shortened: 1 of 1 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -894,10 +894,9 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
         $sqlWhere = !empty($where) ? 'where ' . implode(' AND ', $where) : '';
 
         $sql = "
-            SELECT table_main.*
+            SELECT DISTINCT _uid, table_main.*
```
fogelito, line 897: We will have in the fetch twice _uid , is this ok with PHP?
SELECT DISTINCT _uid as `only_for_remove_duplicate_field`, table_main.*

fogelito: Currently, I do not think we need group by and not distinct since we do not have any table joins at all...
So could be can be removed until that day....

abnegate:
Closing as we will need this in future

PR #225, opened 2023-01-10
Merge latest postgres changes into original postgres PR

vermakhushboo:


tests/Database/Base.php
```diff
@@ -712,45 +712,45 @@ public function testGetDocument(Document $document)
...
-         */
-        $documents = static::getDatabase()->find('documents', [
-            Query::search('string', '*test+alias@email-provider.com'),
-        ]);
-
-        $this->assertEquals(1, count($documents));
-
-        return $document;
-    }
+    // /**
```
abnegate, line 715: What's the reasoning for commenting these tests?

tests/Database/Base.php
```diff
@@ -712,45 +712,45 @@ public function testGetDocument(Document $document)
...
+    //  * @depends testCreateDocument
+    //  */
+    // public function testListDocumentSearch(Document $document)
```
abnegate, line 718: There are still a couple of commented tests here

src/Database/Adapter/Postgres.php
```diff
@@ -122,36 +87,50 @@ public function createCollection(string $name, array $attributes = [], array $in
             ->prepare("CREATE TABLE IF NOT EXISTS \"{$database}\".\"{$namespace}_{$id}\" (
                 \"_id\" SERIAL NOT NULL,
                 \"_uid\" VARCHAR(255) NOT NULL,
-                \"_read\" TEXT[] NOT NULL,
-                \"_write\" TEXT[] NOT NULL,
+                \"_createdAt\" TIMESTAMP DEFAULT NULL,
```
fogelito, line 90: Should the format be in millisesonds such as timestamp(3)

src/Database/Adapter/Postgres.php
```diff
@@ -211,11 +190,68 @@ public function deleteAttribute(string $collection, string $id, bool $array = fa
...
+    {
+        $name = $this->filter($collection);
+        $id = $this->filter($id);
+        $type = $this->getSQLType($type, $size, $signed);
+
+        if ($array) {
+            $type = 'LONGTEXT';
+        }
+
+        if ($type == 'TIMESTAMP') {
```
fogelito, line 245: If the format would be in type  timestamp(3) I am not sure we need the TO_TIMESTAMP?

src/Database/Adapter/Postgres.php
```diff
@@ -768,37 +977,41 @@ protected function getSQLType(string $type, int $size, bool $signed = true): str
...
-            break;
+                break;
 
             case Database::VAR_DOCUMENT:
                 return 'VARCHAR';
-            break;
+                break;
+
+            case Database::VAR_DATETIME:
+                return 'TIMESTAMP';
```
fogelito, line 1009: Same here timestamp(3)

src/Database/Adapter/Postgres.php
fogelito: We can add without time zone to be persistant?

tests/Database/Base.php
```diff
@@ -2770,7 +2770,7 @@ public function testUpdateAttributeDefault()
             ],
             'name' => 'Violet',
             'inStock' => 51,
-            'date' => '2000-06-12 14:12:55.000'
+            'date' => '2000-06-12 14:12:55.123'
```
fogelito, line 2773: Since we use milliseconds , and 000 is ignored, and data is fetched as strings, we must add a tests to verify
2000-06-12 14:12:55.000
Is not fetched as 2000-06-12 14:12:55

src/Database/Adapter/Postgres.php
fogelito: I Do not think you need to write without time zone  since it is default , or to be consistent with all

src/Database/Adapter/Postgres.php
vermakhushboo: MariaDB also has without timezone. Not sure if we want to include timezone.

fogelito: We do not want timezone.
I see when we write only timestamp , postgres treats this as "timestamp without time zone"
I think we should add the timestamp(3) since we use 3 digit precision?

abnegate (approved):

PR #229, opened 2023-01-24
Add `isNull` and `isNotNull` queries

abnegate:
## What does this PR do?

Add `Query::isNull` and `Query::isNotNull` operators to allow developers to filter documents based on whether specific attributes are or are not null.

## Test Plan

Added tests for new query operators

## Related Issues

https://github.com/appwrite/appwrite/issues/2680

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -2,50 +2,17 @@
 
 namespace Utopia\Database\Adapter;
 
-use PDO;
 use Exception;
+use PDO;
 use PDOException;
-use Utopia\Database\Adapter;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
 use Utopia\Database\Exception\Duplicate;
-use Utopia\Database\Helpers\ID;
 use Utopia\Database\Query;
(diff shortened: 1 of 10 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -878,11 +878,16 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
             });
 
             $conditions = [];
-            $attributeIndex = 0;
-            foreach ($query->getValues() as $key => $value) {
-                $bindKey = 'key_' . $attributeIndex;
-                $conditions[] = $this->getSQLCondition('table_main.`' . $query->getAttribute() . '`', $query->getMethod(), ':attribute_' . $i . '_' . $key . '_' . $bindKey, $value);
-                $attributeIndex++;
+
+            if (empty($query->getValues())) {
```
fogelito, line 882: Do we want to check here for $query->getMethod() == IsNull or IsNotNull method?

abnegate: Yep better to be explicit, will update :+1:

fogelito (approved):

PR #230, opened 2023-01-25
Fixed No Error when Creating a Document with value 0 for array attribute

apoorvdwi:
When creating a document, no error is thrown if you pass 0 to the optional array field. This PR fixes that issue.

Closes https://github.com/appwrite/appwrite/issues/5013


abnegate (approved):
Looks good, thanks for your contribution!

PR #231, opened 2023-01-26
Add select operator

abnegate:
## What does this PR do

Add select query operator:

```php
Query::select(['name', 'title']];
```

The default projection if no `select` query is passed is:
-  `*` for SQL
- No projection filter set for Mongo

Attributes not selected are not present in the response document at all

## Test plan

Updated query tests


src/Database/Adapter.php
```diff
@@ -283,7 +283,7 @@ abstract public function deleteIndex(string $collection, string $id): bool;
      *
      * @return Document
      */
-    abstract public function getDocument(string $collection, string $id): Document;
+    abstract public function getDocument(string $collection, string $id, array $selections = []): Document;
```
eldadfux, line 286: Is this an array of strings? let's update the doc block and README file.

src/Database/Adapter.php
```diff
@@ -331,7 +332,7 @@ abstract public function deleteDocument(string $collection, string $id): bool;
      *
      * @return Document[]
      */
-    abstract public function find(string $collection, array $queries = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], array $cursor = [], string $cursorDirection = Database::CURSOR_AFTER): array;
+    abstract public function find(string $collection, array $queries = [], array $selections = [], int $limit = 25, int $offset = 0, array $orderAttributes = [], array $orderTypes = [], array $cursor = [], string $cursorDirection = Database::CURSOR_AFTER): array;
```
eldadfux, line 335: Why isn't this part of the queries array?

src/Database/Adapter.php
```diff
@@ -489,6 +490,13 @@ abstract public function getAttributeWidth(Document $collection): int;
      */
     abstract public function getKeywords(): array;
 
+    /**
+     * Get a projection given a list of selected attributes
+     *
+     * @return string[]|string
+     */
+    abstract protected function getProjection(array $selections, string $prefix = ''): mixed;
```
eldadfux, line 498: This name doesn't tell me much.

src/Database/Adapter.php
abnegate: Yep it is, updated docblock

src/Database/Adapter.php
abnegate: Updated

src/Database/Adapter.php
abnegate: Only because the code to process the select queries would be common in each adapter, I will update to pass them in `queries` instead, will be better than changing the signature here 👍

README.md
```diff
@@ -119,10 +119,24 @@ static::getDatabase()->createDocument('movies', new Document([
...
+**Get Document**:
+
+```php
+// Get document with all attributes
+$document = static::getDatabase()->getDocument('movies', '1', [
+    Query::select(['name', 'director', 'year']),
+]);
+
+// Get document with a sub-set of attributes
+$document = static::getDatabase()->getDocument('movies', '1', [
```
fogelito, line 135: Both examples look identical

src/Database/Adapter/MariaDB.php
```diff
@@ -778,9 +779,9 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
         $sqlWhere = !empty($where) ? 'WHERE ' . implode(' AND ', $where) : '';
 
         $sql = "
-            SELECT table_main.*
+            SELECT {$this->getAttributeProjection($selections, 'table_main')}
```
fogelito, line 782: Should the alias be a part of the $selections , since on joins clients will have to pick from 2 tables:
A.id, B.name?

src/Database/Adapter/Postgres.php
```diff
@@ -992,6 +992,41 @@ public function sum(string $collection, string $attribute, array $queries = [],
...
+
+        $selections[] = '_uid';
+        $selections[] = '_id';
+        $selections[] = '_createdAt';
+        $selections[] = '_updatedAt';
+        $selections[] = '_permissions';
+
+        if (!empty($prefix)) {
+            foreach ($selections as &$selection) {
+                $selection = "\"{$prefix}\".\"{$selection}\"";
```
fogelito, line 1019: Can we put this function and the MariaDB function on SQL.php and use a function for the quotes change?

src/Database/Database.php
```diff
@@ -1345,12 +1345,20 @@ public function getDocument(string $collection, string $id): Document
...
+
+        $selections = $this->getSelections($collection, $queries);
 
         $validator = new Authorization(self::PERMISSION_READ);
 
-        if ($cache = $this->cache->load('cache-' . $this->getNamespace() . ':' . $collection->getId() . ':' . $id, self::TTL)) {
+        $cacheKey = 'cache-' . $this->getNamespace() . ':' . $collection->getId() . ':' . $id;
+
+        if (!empty($selections)) {
+            $cacheKey .= ':' . \md5(\implode($selections));
```
fogelito, line 1356: Can we do the data manipulations after reading/writing and stay with one key per document?
Because cleaning from Redis we will have to use wildcards?

tests/Database/Validator/QueryTest.php
```diff
@@ -120,6 +120,7 @@ public function testQuery()
         $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('orderDesc("title")')));
         $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('isNull("title")')));
         $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('isNotNull("title")')));
+        $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('select(["title", "description"])')));
```
fogelito, line 123: 1. Will it be possible with this logic later on to add an aggregation function as 
select sum(price), count(*), distinct('attr') FROM.

2. What about aliases 
select count(*) as alias FROM

fogelito: 3. Curiosity question: Can clients send queries as JSON and then we do not need the parse function?

src/Database/Adapter/MariaDB.php
abnegate: For now yes, in future we'll need to refactor the whole query to allow joins 👍

src/Database/Adapter/Postgres.php
abnegate: Let's do this in a separate PR to keep the scope small

src/Database/Database.php
abnegate: We are already clearing from Redis with wildcard, so no need for a change there. I think if we manipulate the data after reading then we lose some of the benefit of the feature. This way should be closer to being able to support join selections too

tests/Database/Validator/QueryTest.php
abnegate: We will implement aggregation queries separately.

 No alias support as of yet as I can't see a benefit to the developer.

They're currently a custom syntax that make it a little easier to send from client side than JSON, however JSON might be a good option as we expand the query types. Will discuss further in future :+1:

fogelito (commented):
Looking good

src/Database/Adapter.php
```diff
@@ -489,6 +489,38 @@ abstract public function getAttributeWidth(Document $collection): int;
...
+     */
+    abstract protected function getAttributeProjection(array $selections, string $prefix = ''): mixed;
+
+    /**
+     * Get all selected attributes from queries
+     *
+     * @param Query[] $queries
+     * @return string[]
+     */
+    protected function getAttributeSelections(array $queries): array
```
fogelito, line 507: Maybe we can get by type 'select'?
$cursor = Query::getByType($queries, Query::TYPE_CURSORAFTER, Query::TYPE_CURSORBEFORE);

src/Database/Adapter/Mongo.php
```diff
@@ -385,15 +385,24 @@ public function deleteIndex(string $collection, string $id): bool
...
+     * @throws MongoException
      */
-    public function getDocument(string $collection, string $id): Document
+    public function getDocument(string $collection, string $id, array $queries = []): Document
     {
         $name = $this->getNamespace() . '_' . $this->filter($collection);
 
-        $result = $this->client->find($name, ['_uid' => $id], ['limit' => 1])->cursor->firstBatch;
+        $filters = ['_uid' => $id];
+        $options = ['limit' => 1];
```
fogelito, line 397: General comment
I think we need to remove limit 1.
If we have more than one result, that means something is wrong.

abnegate: Yep makes sense, I'll remove it

fogelito (approved):

PR #232, opened 2023-01-29
Between Operator

fogelito:


abnegate (commented):
Looking good, just some small comments and need to merge main in 👍

src/Database/Adapter/Mongo.php
```diff
@@ -958,7 +958,10 @@ protected function buildFilters($queries): array
             $operator = $this->getQueryOperator($query->getMethod());
             $value = (count($query->getValues()) > 1) ? $query->getValues() : $query->getValues()[0];
 
-            if (is_array($value) && $operator === '$eq') {
+            if ($operator === 'between' && isset($value[0]) && isset($value[1])) {
```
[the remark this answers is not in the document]
fogelito: ok!

src/Database/Adapter/MariaDB.php
```diff
@@ -1929,4 +1841,63 @@ public static function getPDOAttributes(): array
...
+            case Query::TYPE_SEARCH:
+                /**
+                 * Replace reserved chars with space.
+                 */
+                $value = trim(str_replace(['@', '+', '-', '*'], ' ', $query->getValues()[0]));
+                /**
+                 * Prepend wildcard by default on the back.
+                 */
+                $value = "'{$value}*'";
+                return 'match('.$attribute.') against('.$this->getPDO()->quote($value).' in boolean mode)';
```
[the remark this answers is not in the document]
fogelito: Done

composer.json
```diff
@@ -15,11 +15,11 @@
...
         "utopia-php/framework": "0.*.*",
         "utopia-php/cache": "0.8.*",
-        "utopia-php/mongo": "0.0.2"
+        "utopia-php/mongo": "0.0.2",
+        "ext-pdo": "*"
     },
     "require-dev": {
-        "ext-pdo": "*",
-        "ext-redis": "*",
+      "ext-redis": "*",
```
abnegate, line 22: Let's move PDO, Mongo, Redis, etc. to the `"suggests"` block. They should not be required or required for dev, since you could use the library with only mongo, then you don't need PDO or Redis etc.

src/Database/Adapter/Postgres.php
```diff
@@ -852,6 +819,8 @@ public function find(string $collection, array $queries = [], int $limit = 25, i
         return $results;
     }
 
+
+
```
abnegate, line 823: Let's remove these redundant lines

src/Database/Query.php
```diff
@@ -145,6 +146,7 @@ public static function isMethod(string $value): bool
             self::TYPE_CURSORAFTER,
             self::TYPE_CURSORBEFORE,
             self::TYPE_IS_NULL,
+            self::TYPE_BETWEEN,
```
abnegate, line 149: Let's move this after `TYPE_IS_NOT_NULL` just to keep things ordered

tests/Database/Base.php
```diff
@@ -1159,6 +1159,21 @@ public function testFindNotEqual()
...
+
+    public function testBetween(){
+        $documents = static::getDatabase()->find('movies', [
+            Query::between('price', 25.94, 25.99),
+        ]);
+
+        $this->assertEquals(2, count($documents));
+
+        $documents = static::getDatabase()->find('movies', [
+            Query::between('$createdAt', '1975-12-06', '2050-12-06'),
```
abnegate, line 1171: Will it work if you pass an actual `DateTime` object here?

tests/Database/Base.php
```diff
@@ -1159,6 +1159,21 @@ public function testFindNotEqual()
 
     }
 
+
+    public function testBetween(){
```
abnegate, line 1163: Let's name this `testFindBetween` for consistency

tests/Database/Base.php
```diff
@@ -1159,6 +1159,21 @@ public function testFindNotEqual()
...
+        $documents = static::getDatabase()->find('movies', [
+            Query::between('price', 25.94, 25.99),
+        ]);
+
+        $this->assertEquals(2, count($documents));
```
abnegate, line 1168: Let's add a case that checks the `price` attribute, but outside the bounds of any document, so we get `count($document) == 0`

composer.json
fogelito: I see them in the suggested block.
On mariaDB we are using etc PDO::PARAM_INT so does it supposed to be required?

src/Database/Adapter/Postgres.php
fogelito: done!

src/Database/Query.php
fogelito: Done!

tests/Database/Base.php
fogelito: Done!

tests/Database/Base.php
fogelito: Yes, if the attributes has a filter 'datetime' it will work, I will add another test for this

tests/Database/Base.php
fogelito: Done!

abnegate (approved):

PR #233, opened 2023-02-06
Update Database to have a timestamp property

stnguyen90:
The timestamp property is then used to provide context for
when the request was initially made so that a Conflict
exception can be thrown if a Document was updated or
deleted after the initial request timestamp.

There were 3 options in how to add this timestamp:

1. Add the timestamp to the updateDocument() function
2. Add setTimestamp() and resetTimestamp()
3. Add a withRequestTimestamp() function

The problem with option 1 is that it would pollute the signature and
make it messy.

The problem with option 2 is it adds a global state that developers
would need to make sure to reset. Not doing so can lead to unexpected
behavior.

Option 3 will require modifying places were we want to use the request
timestamp, but this explicitness is best for maintainability.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -6,6 +6,7 @@ use Exception;
 use Throwable;
 use Utopia\Cache\Cache;
 use Utopia\Database\Exception\Authorization as AuthorizationException;
+use Utopia\Database\Exception\Conflict as ConflictException;
 use Utopia\Database\Exception\Duplicate as DuplicateException;
 use Utopia\Database\Exception\Limit as LimitException;
 use Utopia\Database\Exception\Structure as StructureException;
@@ -234,6 +235,12 @@ class Database
     protected bool $silentEvents = false;
 
     /**
+     * Timestamp for the current request
+     * @var ?\DateTime
+     */
+    protected ?\DateTime $timestamp = null;
+
+    /**
      * @param Adapter $adapter
      * @param Cache $cache
      */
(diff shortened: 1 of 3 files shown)


src/Database/Database.php
```diff
@@ -1455,6 +1483,12 @@ public function updateDocument(string $collection, string $id, Document $documen
             throw new AuthorizationException($validator->getDescription());
         }
 
+        // Check if document was updated after the request timestamp
+        $oldUpdatedAt = new \DateTime($old->getUpdatedAt());
+        if (!is_null($this->timestamp) && $oldUpdatedAt > $this->timestamp) {
```
stnguyen90, line 1488: I've added this check to `Database`, but I'm slightly concerned about any side effects this may cause like when some audit document is being updated during a project document update.

abnegate (commented):
Not sure if we need the timestamp as a property at the database level, what about an optional parameter `$requestedAt` for the update/delete document methods? Feels a bit easier to understand than `setTimeout` and `resetTimeout` cc @eldadfux

eldadfux:
I think this way is better for now in order to avoid signature pollution, we can workout the naming if we have better suggestions. This approach was to try and avoid having a huge list of arguments in every write function.

src/Database/Database.php
```diff
@@ -360,6 +367,26 @@ protected function trigger(string $event, mixed $args = null): void
         }
     }
 
+    /**
+     * Executes $callback with $timestamp set to $requestTimestamp
+     * 
+     * @template T
+     * @param \DateTime $requestTimestamp
+     * @param callable(): T $callback
+     * @return T
```
abnegate, line 376: Does this still work if we add the `mixed` return type to the function declaration as well?

stnguyen90: Looks like it does!

src/Database/Database.php
```diff
@@ -360,6 +367,26 @@ protected function trigger(string $event, mixed $args = null): void
...
+     */
+    function withRequestTimestamp(\DateTime $requestTimestamp, callable $callback): mixed {
+        $previous = $this->timestamp;
+        $this->timestamp = $requestTimestamp;
+        try {
+            $result = $callback();
+        } finally {
+            $this->timestamp = $previous;
+        }
+        var_dump($this->timestamp);
```
abnegate, line 386: Let's remove this

abnegate (approved):

src/Database/Database.php
stnguyen90: Ooops! It should be removed now

eldadfux (approved):

PR #235, opened 2023-02-08
Add `startsWith` and `endsWith` operator

abnegate:
## What's new

- Added startsWith and endsWith query operators that match the start or end of a string. 
- Uses the `LIKE` operator for SQL
- Uses the `$regex` opearator for Mongo.

## Tests

Added new tests

### DIFF
--- a/src/Database/Adapter/SQL.php
+++ b/src/Database/Adapter/SQL.php
@@ -685,32 +685,36 @@ abstract class SQL extends Adapter
         switch ($method) {
             case Query::TYPE_EQUAL:
                 return '=';
-
             case Query::TYPE_NOTEQUAL:
                 return '!=';
-
             case Query::TYPE_LESSER:
                 return '<';
-
             case Query::TYPE_LESSEREQUAL:
                 return '<=';
-
             case Query::TYPE_GREATER:
                 return '>';
-
             case Query::TYPE_GREATEREQUAL:
                 return '>=';
-
             case Query::TYPE_IS_NULL:
                 return 'IS NULL';
-
             case Query::TYPE_IS_NOT_NULL:
                 return 'IS NOT NULL';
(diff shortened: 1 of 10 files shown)


tests/Database/Validator/QueryTest.php
```diff
@@ -120,6 +120,8 @@ public function testQuery()
         $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('orderDesc("title")')));
         $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('isNull("title")')));
         $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('isNotNull("title")')));
+        $this->assertEquals(true, $validator->isValid(DatabaseQuery::parse('startsWith("title", "Fro)')));
```
fogelito, line 123: - Are we ok with limits in case someone does 'startsWith' 'A' for example?
- Maybe to have a test with 2 words?
- Are we ok with special char for mongo regex? as '\' , '*'?

src/Database/Adapter/SQL.php
```diff
@@ -685,34 +685,38 @@ protected function getSQLOperator(string $method): string
...
+                return 'LIKE';
             default:
                 throw new Exception('Unknown method:' . $method);
-                break;
         }
     }
 
+    protected function getSQLValue(string $method, mixed $value)
+    {
+        return match($method) {
```
fogelito, line 712: Are we forcing an error when value is empty?

tests/Database/Validator/QueryTest.php
abnegate: - Yep should be okay in G4
- Added a test for two words
- Added escaping for wildcards

fogelito (approved):

PR #237, opened 2023-02-17
Remove requirement to pass limit/offset/max for find/count/sum

abnegate:
## What's Changed

Remove the requirement to pass a limit/offset/max for find/count/sum. This makes it possible to request documents without passing a limit instead of using `PHP_INT_MAX` or using 0 to mean unlimited.

## Tests

- Existing tests


fogelito (approved):
I think we need to investigate the performance of count/sum as we do now when we remove the limit:
now we are counting the fetched results. maybe is better just count(*)  as aggregation?

eldadfux (approved):

PR #243, opened 2023-03-06
Dynamic attibute for related primary key

fogelito:
Database::VAR_ID
Since some adapters primary key is a string and some is an int , we need a dynamic ID field per adapter


src/Database/Adapter/MariaDB.php
```diff
@@ -1082,6 +1082,9 @@ protected function getSQLType(string $type, int $size, bool $signed = true): str
 
                 return "VARCHAR({$size})";
 
+            case Database::VAR_ID:
+                return 'INT UNSIGNED'; // Same as Primary Key Sqlite does not allow INT(11)
```
abnegate, line 1086: Can we make this `INT(11)` and override for SQLite?

src/Database/Database.php
```diff
@@ -759,7 +760,8 @@ public function createAttribute(string $collection, string $id, string $type, in
             case self::VAR_FLOAT:
             case self::VAR_BOOLEAN:
             case self::VAR_DATETIME:
-                break;
+            case self::VAR_ID:
+            break;
```
abnegate, line 764: Let's fix the indentation here:

```suggestion
                break;
```

src/Database/Validator/Structure.php
```diff
@@ -253,6 +253,10 @@ public function isValid($document): bool
                     $validator = new Integer();
                     break;
 
+                case Database::VAR_ID:
+                    $validator = new Text(24, min: 0); // Mongo Id length
```
abnegate, line 257: I think this can be 36 for non-mongo adapters :thinking: how can we handle both?

fogelito:
Added an Asana task in the backlog

PR #245, opened 2023-03-08
Use VARCHAR instead of CHAR for MariaDB UID

fogelito:


abnegate:
@fogelito Let's create a migration PR for this in Appwrite and fix the conflicts here

abnegate:
@fogelito Now that we have the migration PR let's fix the conflicts and lint issues here

fogelito:
cool

abnegate (approved):

PR #246, opened 2023-03-12
Feat relationships

abnegate:
## What's changed

- Add create/update/delete relationship attributes
- Allow creating documents with relationships, either with existing document ID's or nested documents. If nested documents do not exist, they are created. They are updated if they already exist and the data is different. 
- Add fetching documents with relationships. Allows nesting up to a maximum level.
- Add updating documents with relationships.
- Add deleting documents with relationships.
- ~~Allow querying documents with relationships. Will filter parent documents based on child attributes. Uses dot-paths to specify child attributes. E.g. For a one-to-many, author to posts relationship `Query::equal('posts.title', ['Post 1'])` will return all authors that have at least one post where the title equals 'Post 1'.~~
- Allow selecting attributes from child relationships. Must either select some or all parent attributes, defaults to all. Uses dot-paths to specify child attributes. Can use wildcard `*` to specify selection of all a relationships attributes. E.g. `Query::select(['posts.title'])` or `Query::select(['posts.*'])` if parent attributes are selected, and relationship attributes are not, relationships are not fetched. 

## Test Plan

- Added tests covering CRUD for relationship attributes of all types, and CRUD for documents for each type, as well as querying and selecting.
- [ ] Performance testing and benchmarking


src/Database/Database.php
```diff
@@ -1220,6 +1305,401 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+        $this->updateAttributeMeta($collection, $key, function ($attribute) use ($collection, $key, $newKey, $newTwoWayKey, $twoWay, $onDelete) {
+            $altering = !\is_null($newKey) || !\is_null($newTwoWayKey);
+
+            $relatedCollection = $attribute['options']['relatedCollection'];
+            $type = $attribute['options']['relationType'];
+            $side = $attribute['options']['side'];
+
+            $newKey ??= $attribute['key'];
+            $twoWayKey = $attribute['options']['twoWayKey'];
+            $newTwoWayKey ??= $attribute['options']['twoWayKey'];
```
fogelito, line 1517: Do we want to compare old to new to skip the Rename?

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5426 @@ public function testWritePermissions(): void
...
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Person 10',
+            'library' => [
+                '$id' => 'library10',
+                'name' => 'Library 10 Updated',
+                'area' => 'Area 10 Updated',
+            ],
+        ]));
```
Meldiron, line 3788: I cant see test for this. Can you double check?
I expect this should have library 10 updated.

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5426 @@ public function testWritePermissions(): void
...
+            key: 'library',
+            newKey: 'newLibrary'
+        );
+
+        // Get document with again
+        $person = static::getDatabase()->getDocument('person', 'person1');
+        $library = $person->getAttribute('newLibrary');
+        $this->assertEquals('library4', $library['$id']);
+
+        // Can not delete document while still related to another with on delete set to restrict
```
Meldiron, line 3976: Let's have test for de-attaching library from it (set to null I guess?) and it should succeed after.

Let's also have test for trying to delete library. Should also not work while connection is there, I would guess.

docker-compose.yml
```diff
@@ -13,6 +13,7 @@ services:
       - ./bin:/usr/src/code/bin
       - ./src:/usr/src/code/src
       - ./tests:/usr/src/code/tests
+      - ./tmp:/usr/src/code/tmp
```
TorstenDittmann, line 16: leftover?

src/Database/Database.php
```diff
@@ -861,23 +942,23 @@ private function updateAttributeMeta(string $collection, string $id, callable $u
         }
 
         $attributes = $collection->getAttribute('attributes', []);
+        $i = \array_search($id, \array_map(fn ($attribute) => $attribute['$id'], $attributes));
```
TorstenDittmann, line 945: Not sure if single digit char variable names work so well here, wouldn't something like `$found` or whatever tells the story better?

tests/Database/Base.php
abnegate: Yep assertions were missing, added now

src/Database/Database.php
```diff
@@ -1220,6 +1305,401 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+        $twoWay = $relationship['options']['twoWay'];
+        $twoWayKey = $relationship['options']['twoWayKey'];
+
+        $relatedCollection = $this->silent(fn () => $this->getCollection($relatedCollection));
+        $relatedAttributes = $relatedCollection->getAttribute('attributes', []);
+
+        foreach ($relatedAttributes as $name => $attribute) {
+            if ($attribute['$id'] === $twoWayKey) {
+                unset($relatedAttributes[$name]);
+            }
```
TorstenDittmann, line 1654: same here you might wanna short circuit

src/Database/Database.php
```diff
@@ -1220,6 +1305,401 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+                    }
+                    break;
+                case self::RELATION_ONE_TO_MANY:
+                    $this->deleteIndex($relatedCollection->getId(), $twoWayIndexKey);
+                    break;
+                case self::RELATION_MANY_TO_ONE:
+                    $this->deleteIndex($collection->getId(), $indexKey);
+                    break;
+                case self::RELATION_MANY_TO_MANY:
+                    break;
```
TorstenDittmann, line 1680: if not needed, we could remove it to bring down noise

src/Database/Database.php
```diff
@@ -1417,7 +1901,27 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
+        $selections = $this->validateSelections($collection, $selects);
+        $nestedSelections = [];
+
+        foreach ($queries as $index => $query) {
+            switch ($query->getMethod()) {
+                case Query::TYPE_SELECT:
+                    foreach ($query->getValues() as $value) {
+                        if (\str_contains($value, '.')) {
+                            $nestedSelections[] = Query::select([
+                                \implode('.', \array_slice(\explode('.', $value), 1))
```
TorstenDittmann, line 1914: might wanna leave a comment here about this part. like a `before -> after` of a sample string

src/Database/Database.php
```diff
@@ -1417,7 +1901,27 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
+                            ]);
+                            unset($queries[$index]);
+                            break;
+                        }
+                    }
+                    break;
+                default:
+                    break;
```
TorstenDittmann, line 1922: do we need an empty `default`?

src/Database/Database.php
```diff
@@ -1454,31 +1958,267 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
+            $this->silent(fn () => $this->getDocumentRelationships($collection, $document, $nestedSelections));
+        }
+
         $document = $this->casting($collection, $document);
         $document = $this->decode($collection, $document, $selections);
 
-        $this->cache->save($cacheKey, $document->getArrayCopy()); // save to cache after fetching from db
+        $relationships = \array_filter($collection->getAttribute('attributes', []), function ($attribute) {
+            return $attribute['type'] === Database::VAR_RELATIONSHIP;
+        });
```
TorstenDittmann, line 1970: could use an arrow function here

src/Database/Database.php
```diff
@@ -1454,31 +1958,267 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
     }
 
     /**
-     * Create Document
-     *
-     * @param string $collection
+     * @param Document $collection
      * @param Document $document
-     *
+     * @param array<Query> $queries
```
TorstenDittmann, line 1993: just a fyi that `Query[]` works too 👍🏻

src/Database/Database.php
```diff
@@ -1454,31 +1958,267 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
+                        && $existingRelatedCollection === $collection->getId()
+                        && $existingSide !== $side)
+                    || ($existingKey === $key
+                        && $existingTwoWayKey !== $twoWayKey
+                        && $existingRelatedCollection === $relatedCollection->getId()
+                        && $existingSide !== $side)
+                    || ($existingKey !== $key
+                        && $existingTwoWayKey === $twoWayKey
+                        && $existingRelatedCollection === $relatedCollection->getId()
+                        && $existingSide !== $side));
```
TorstenDittmann, line 2052: thats quite the monster 😁 

would you think its worth leaving comments for each grouped condition? I feel like this would be a part where as an engineer unknown to this part would be quite intimidated 😄

src/Database/Database.php
```diff
@@ -1454,31 +1958,267 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
+                    $fetchedRelationships[] = $relationship;
+
+                    $relatedDocuments = $this->find($relatedCollection->getId(), [
+                        Query::equal($twoWayKey, [$document->getId()]),
+                        Query::limit(PHP_INT_MAX),
+                        ...$queries
+                    ]);
+
+                    $fetchDepth--;
+                    \array_pop($fetchedRelationships);
```
TorstenDittmann, line 2116: any reason we do `fetchDepth++` and `fetchDepth--` without usage in between?

TorstenDittmann: ah nvm, just saw its static

src/Database/Database.php
```diff
@@ -1454,31 +1958,267 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
     {
-        $collection = $this->silent(fn () => $this->getCollection($collection));
+        $attributes = $collection->getAttribute('attributes', []);
+
+        $relationships = \array_filter($attributes, function ($attribute) {
+            return $attribute['type'] === Database::VAR_RELATIONSHIP;
+        });
+
+        static $fetchDepth = 1;
+        static $fetchedRelationships = [];
```
TorstenDittmann, line 2006: is this static to the function of the instance or the class? just having in mind that we potentially can have multiple instances of the Database

src/Database/Database.php
```diff
@@ -1496,15 +2236,276 @@ public function createDocument(string $collection, Document $document): Document
...
+     * @throws Exception
+     * @throws Throwable
+     */
+    private function createDocumentRelationships(Document $collection, Document $document): void
+    {
+        $attributes = $collection->getAttribute('attributes', []);
+
+        $relationships = \array_filter($attributes, function ($attribute) {
+            return $attribute['type'] === Database::VAR_RELATIONSHIP;
+        });
```
TorstenDittmann, line 2269: could use an arrow function here

src/Database/Database.php
```diff
@@ -1560,6 +2570,347 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+
+                    switch (\gettype($value)) {
+                        case 'string':
+                            $related = $this->skipRelationships(fn () => $this->getDocument($relatedCollection->getId(), $value));
+
+                            if (
+                                $oldValue?->getId() !== $value
+                                && $this->findOne($relatedCollection->getId(), [
+                                    Query::equal($twoWayKey, [$document->getId()]),
+                                ]) !== false
```
TorstenDittmann, line 2648:
```suggestion
                                && $this->findOne($relatedCollection->getId(), [
                                    Query::equal($twoWayKey, [$document->getId()]),
                                ])
```

should be the same am i right?

src/Database/Database.php
```diff
@@ -1560,6 +2570,347 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                            break;
+                        case 'object':
+                            if ($value instanceof Document) {
+                                $related = $this->skipRelationships(fn () => $this->getDocument($relatedCollection->getId(), $value->getId()));
+
+                                if (
+                                    $oldValue?->getId() !== $value->getId()
+                                    && $this->findOne($relatedCollection->getId(), [
+                                        Query::equal($twoWayKey, [$document->getId()]),
+                                    ]) !== false
```
TorstenDittmann, line 2668:
```suggestion
                                    && $this->findOne($relatedCollection->getId(), [
                                        Query::equal($twoWayKey, [$document->getId()]),
                                    ])
```

same here

src/Database/Database.php
```diff
@@ -1560,6 +2570,347 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+
+                        if ($related->isEmpty()) {
+                            if (! isset($value['$permissions'])) {
+                                $value->setAttribute('$permissions', $document->getAttribute('$permissions'));
+                            }
+                            $this->createDocument(
+                                $relatedCollection->getId(),
+                                $value
+                            );
+                        } elseif ($related->getAttributes() != $value->getAttributes()) {
```
TorstenDittmann, line 2802:
```suggestion
                        } elseif ($related->getAttributes() !== $value->getAttributes()) {
```
strict equal 👍🏻

src/Database/Database.php
```diff
@@ -1795,12 +3380,119 @@ public function find(string $collection, array $queries = [], ?int $timeout = nu
...
+                            $matched = $value <= $query->getValue();
+                            break;
+                        case Query::TYPE_CONTAINS:
+                            $matched = \in_array($query->getValue(), $value);
+                            break;
+                        case Query::TYPE_SEARCH:
+                            $matched = \str_contains($value, $query->getValue());
+                            break;
+                        case Query::TYPE_IS_NULL:
+                            $matched = $value == null;
```
TorstenDittmann, line 3469:
```suggestion
                            $matched = $value === null;
```

strict equal :)

tests/Database/Base.php
abnegate: Good call there was a bug there. Added tests

docker-compose.yml
abnegate: Yep, removed

src/Database/Database.php
abnegate: Yep updated

src/Database/Database.php
```diff
@@ -1220,6 +1305,401 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+     * @throws StructureException
+     * @throws Throwable
+     */
+    public function createRelationship(
+        string $collection,
+        string $relatedCollection,
+        string $type,
+        bool $twoWay = false,
+        string $id = '',
+        string $twoWayKey = '',
```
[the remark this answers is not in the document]
abnegate: Yep makes sense, updated

src/Database/Database.php
```diff
@@ -1220,6 +1305,401 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+                    break;
+                case self::RELATION_ONE_TO_MANY:
+                    $this->createIndex($relatedCollection->getId(), $twoWayIndexKey, self::INDEX_KEY, [$twoWayKey]);
+                    break;
+                case self::RELATION_MANY_TO_ONE:
+                    $this->createIndex($collection->getId(), $indexKey, self::INDEX_KEY, [$id]);
+                    break;
+                case self::RELATION_MANY_TO_MANY:
+                    // Indexes created on junction collection creation
+                    break;
```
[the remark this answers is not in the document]
abnegate: We need the empty case here so the exception from the default isn't thrown for this type

src/Database/Database.php
```diff
@@ -1220,6 +1305,401 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+    public function deleteRelationship(string $collection, string $id): bool
+    {
+        $collection = $this->silent(fn () =>$this->getCollection($collection));
+        $attributes = $collection->getAttribute('attributes', []);
+        $relationship = null;
+
+        foreach ($attributes as $name => $attribute) {
+            if ($attribute['$id'] === $id) {
+                $relationship = $attribute;
+                unset($attributes[$name]);
```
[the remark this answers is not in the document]
abnegate: Yep good call, done

src/Database/Database.php
abnegate: Very fair 😅 added

src/Database/Database.php
abnegate: Yep, will reduce it to an if since there's only the one case

src/Database/Database.php
abnegate: I prefer this way in case it ever becomes an associative array it's easier to add another type than change formats

src/Database/Database.php
abnegate: This one is intentional, strict check for arrays will compare order which we don't want here

src/Database/Database.php
abnegate: I thought static to the function of the instance but I was wrong 😅 I will them convert to instance properties

src/Database/Database.php
abnegate: This intimidates me and I wrote it, took like 2 days to get it right 😆  added some extra comments to try help explain what they're checking

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5426 @@ public function testWritePermissions(): void
...
+            ->getAttribute('name')
+        );
+
+        $countries = static::getDatabase()->find('country');
+
+        $this->assertEquals(4, \count($countries));
+
+        // Select related document attributes
+        $country = static::getDatabase()->findOne('country', [
+            Query::select(['*', 'city.name'])
```
Meldiron, line 4258: Do we have test for `::select(['*', 'city.*'])`? A test for * in relationship

abnegate: Yep there is a test for that further down here:

https://github.com/utopia-php/database/blob/feat-relationships/tests/Database/Base.php#L7071-L7074

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5426 @@ public function testWritePermissions(): void
...
+        $this->assertArrayNotHasKey('artist', $albums[0]);
+
+        $artist = static::getDatabase()->getDocument('artist', 'artist2');
+        $albums = $artist->getAttribute('albums', []);
+        $this->assertEquals('album2', $albums[0]['$id']);
+        $this->assertArrayNotHasKey('artist', $albums[0]);
+
+        // Get related document
+        $album = static::getDatabase()->getDocument('album', 'album1');
+        $this->assertArrayNotHasKey('albums', $album);
```
Meldiron, line 4698: Would also make sense to assert its one way. so notHasKey artist

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5426 @@ public function testWritePermissions(): void
...
+        $attributes = $collection->getAttribute('attributes', []);
+        foreach ($attributes as $attribute) {
+            if ($attribute['key'] === 'movies') {
+                $this->assertEquals('relationship', $attribute['type']);
+                $this->assertEquals('movies', $attribute['$id']);
+                $this->assertEquals('movies', $attribute['key']);
+                $this->assertEquals('movie', $attribute['options']['relatedCollection']);
+                $this->assertEquals(Database::RELATION_MANY_TO_ONE, $attribute['options']['relationType']);
+                $this->assertEquals(false, $attribute['options']['twoWay']);
+                $this->assertEquals('review', $attribute['options']['twoWayKey']);
```
Meldiron, line 5432: We set twoWayKey ot reviews, and assert its review.

abnegate: Good catch!

composer.lock
christyjacob4, line 1: One of those rare PRs where the lock file has lesser changes than the rest! Love it 😄

fogelito (approved):

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5594 @@ public function testWritePermissions(): void
...
+        static::getDatabase()->createCollection('movie');
+
+        static::getDatabase()->createAttribute('review', 'name', Database::VAR_STRING, 255, true);
+        static::getDatabase()->createAttribute('movie', 'name', Database::VAR_STRING, 255, true);
+        static::getDatabase()->createAttribute('movie', 'length', Database::VAR_INTEGER, 0, true, formatOptions: ['min' => 0, 'max' => 999]);
+
+        static::getDatabase()->createRelationship(
+            collection: 'review',
+            relatedCollection: 'movie',
+            type: Database::RELATION_MANY_TO_ONE,
```
Meldiron, line 5421: This relation is like ONE_TO_ONE, but it doesn't have to be unique. Right?
If so, let's add test where we have 2 reviews wit same related movie. It should not fail to create.

tests/Database/Base.php
```diff
@@ -3704,6 +3705,5594 @@ public function testWritePermissions(): void
...
+        static::getDatabase()->createCollection('song');
+
+        static::getDatabase()->createAttribute('playlist', 'name', Database::VAR_STRING, 255, true);
+        static::getDatabase()->createAttribute('song', 'name', Database::VAR_STRING, 255, true);
+        static::getDatabase()->createAttribute('song', 'length', Database::VAR_INTEGER, 0, true);
+
+        static::getDatabase()->createRelationship(
+            collection: 'playlist',
+            relatedCollection: 'song',
+            type: Database::RELATION_MANY_TO_MANY,
```
Meldiron, line 6235: Let's have a test where 2 playlists share the same song. It should be fine, no unique exceptions

Meldiron (approved):

src/Database/Database.php
```diff
@@ -1795,17 +3405,136 @@ public function find(string $collection, array $queries = [], ?int $timeout = nu
...
-        foreach ($results as &$node) {
+        $attributes = $collection->getAttribute('attributes', []);
+
+        $relationships = $this->resolveRelationships ? \array_filter($attributes, function (Document $attribute) {
+            return $attribute->getAttribute('type') === self::VAR_RELATIONSHIP;
+        }) : [];
+
+        foreach ($results as $index => &$node) {
+            if ($this->resolveRelationships && (empty($selects) || !empty($nestedSelections))) {
+                $node = $this->silent(fn () => $this->getDocumentRelationships($collection, $node, $nestedSelections));
```
stnguyen90, line 3416: I'm running into a problem where nested data isn't being returned as expected and I think it's because the collection I'm calling `listDocuments()` on has document-level security, but the related collection doesn't.

With Admin mode, I get:

```json
{
    "movie": {
        "name": "Lightyear",
        "description": "Legendary Space Ranger Buzz Lightyear embarks on an intergalactic adventure alongside a group of ambitious recruits and his robot companion Sox.",
        "thumbnailImageId": "635721f1e1ae447d46ed",
        "cast": [
            "Chris Evans",
            "James Brolin",
            "Keke Palmer",
            "Taika Waititi",
            "Debra Wilson"
        ],
        "tags": [
            "space travel",
            "alien life-form",
            "stranded",
            "alien planet",
            "spin off",
            "aftercreditsstinger",
            "duringcreditsstinger",
            "origin story"
        ],
        "genres": [
            "Animation",
            "Action",
            "Adventure",
            "Family",
            "Science Fiction"
        ],
        "durationMinutes": 105,
        "netflixReleaseDate": "2023-01-30T00:00:00.000+00:00",
        "releaseDate": "2022-06-15T00:00:00.000+00:00",
        "trendingIndex": 903.525,
        "isOriginal": false,
        "ageRestriction": "AR13",
        "$id": "641e2bc990f204aade06",
        "$createdAt": "2023-03-24T23:01:29.593+00:00",
        "$updatedAt": "2023-03-24T23:01:46.666+00:00",
        "$permissions": [],
        "$databaseId": "default",
        "$collectionId": "movies"
    },
    "user": {
        "name": "Steven",
        "$id": "641e279d436165d3af09",
        "$createdAt": "2023-03-24T22:44:37.627+00:00",
        "$updatedAt": "2023-03-27T22:07:16.883+00:00",
        "$permissions": [
            "read(\"user:641e279d436165d3af09\")"
        ],
        "$databaseId": "default",
        "$collectionId": "users"
    },
    "$id": "64220b9f516b197ef1e0",
    "$createdAt": "2023-03-27T21:33:19.333+00:00",
    "$updatedAt": "2023-03-27T21:33:19.333+00:00",
    "$permissions": [
        "read(\"user:641e279d436165d3af09\")",
        "update(\"user:641e279d436165d3af09\")",
        "delete(\"user:641e279d436165d3af09\")"
    ],
    "$databaseId": "default",
    "$collectionId": "watchlists"
}
```

but with a user, I get:

```json
{
  "movie": { "$databaseId": "default", "$collectionId": "movies" },
  "user": {
      "name": "Steven",
      "$id": "641e279d436165d3af09",
      "$createdAt": "2023-03-24T22:44:37.627+00:00",
      "$updatedAt": "2023-03-27T22:07:16.883+00:00",
      "$permissions": ["read(\"user:641e279d436165d3af09\")"],
      "$databaseId": "default",
      "$collectionId": "users",
  },
  "$id": "64220b9f516b197ef1e0",
  "$createdAt": "2023-03-27T21:33:19.333+00:00",
  "$updatedAt": "2023-03-27T21:33:19.333+00:00",
  "$permissions": [
      "read(\"user:641e279d436165d3af09\")",
      "update(\"user:641e279d436165d3af09\")",
      "delete(\"user:641e279d436165d3af09\")",
    ],
  "$databaseId": "default",
  "$collectionId": "watchlists",
}
```

My `watchlists` collection has document-level security, but my `movies` has collection-level security and should be readable by anyone.

stnguyen90: This also makes me concerned about how permissions would be handled if I'm fetching a collection that has collection-level permissions with related documents that have document-level permissions. I'm worried the auth skip at the top level would cascade to nested relationships.

src/Database/Database.php
```diff
@@ -1496,15 +2254,276 @@ public function createDocument(string $collection, Document $document): Document
...
+                                    $related,
+                                    $relationType,
+                                    $twoWay,
+                                    $twoWayKey,
+                                    $side,
+                                    $this->relationshipCreateDepth
+                                );
+                                break;
+                            default:
+                                throw new Exception('Invalid relationship value. Must be either a document, document ID, or an array of documents or document IDs.');
```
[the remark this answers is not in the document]
abnegate: There are a lot of cases where we throw a global `Exception` from the database, there's a separate PR to address this with an internal base exception :+1:

src/Database/Adapter/MariaDB.php
```diff
@@ -181,63 +181,62 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+     * @param string $type
+     * @param int $size
+     * @param bool $signed
+     * @param bool $array
      * @return bool
      * @throws Exception
      * @throws PDOException
      */
-    public function renameAttribute(string $collection, string $old, string $new): bool
+    public function updateAttribute(string $collection, string $id, string $type, int $size, bool $signed = true, bool $array = false): bool
```
christyjacob4, line 201: Not sure if we want to use $id as the name of this parameter. Since we use `$key` to refer to it within Appwrite, it might be better to use the same convention here.

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
-    public function renameIndex(string $collection, string $old, string $new): bool
+    public function renameAttribute(string $collection, string $old, string $new): bool
     {
         $collection = $this->filter($collection);
         $old = $this->filter($old);
         $new = $this->filter($new);
 
         return $this->getPDO()
-            ->prepare("ALTER TABLE {$this->getSQLTable($collection)} RENAME INDEX `{$old}` TO `{$new}`;")
+            ->prepare("ALTER TABLE {$this->getSQLTable($collection)} RENAME COLUMN `{$old}` TO `{$new}`;")
```
christyjacob4, line 255: When we perform a `renameAttribute`, are the underlying indexes automatically updated as well ?

fogelito: Yep we are, renaming the metadata indexing on Database.php before sending it to the adapter.

src/Database/Adapter/MariaDB.php
```diff
@@ -181,63 +181,62 @@ public function createAttribute(string $collection, string $id, string $type, in
...
-        $type = $this->getSQLType($type, $size, $signed);
-
-        if ($array) {
-            $type = 'LONGTEXT';
-        }
 
         return $this->getPDO()
             ->prepare("ALTER TABLE {$this->getSQLTable($name)}
-                MODIFY `{$id}` {$type};")
+                DROP COLUMN `{$id}`;")
```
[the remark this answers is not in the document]
fogelito: They are automatically deleted in the adapter, who knows to take care of it but, 
I do not think we are taking care of the metadata indexes as we do on rename, I think we are left with a gap, 
Will make tomorrow more tests for this.
@abnegate will love you opinion here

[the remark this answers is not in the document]
christyjacob4: okay sounds good @fogelito

src/Database/Adapter.php
```diff
@@ -242,6 +241,47 @@ abstract public function deleteAttribute(string $collection, string $id): bool;
...
+     * @param string $collection
+     * @param string $relatedCollection
+     * @param string $type
+     * @param bool $twoWay
+     * @param string $id
+     * @param string $twoWayKey
+     * @param string $onDelete
+     * @return bool
+     */
+    abstract public function createRelationship(string $collection, string $relatedCollection, string $type, bool $twoWay = false, string $id = '', string $twoWayKey = '', string $onDelete = Database::RELATION_MUTATE_RESTRICT): bool;
```
christyjacob4, line 254:
```suggestion
    abstract public function createRelationship(string $collection, string $relatedCollection, string $type, bool $twoWay = false, string $key = '', string $twoWayKey = '', string $onDelete = Database::RELATION_MUTATE_RESTRICT): bool;
```

Better to call this a key since we use key in the other method signatures as well

stnguyen90 (approved):

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
-     * @throws PDOException
      */
-    public function deleteAttribute(string $collection, string $id, bool $array = false): bool
-    {
+    public function createRelationship(
+        string $collection,
+        string $relatedCollection,
+        string $type,
+        bool $twoWay = false,
+        string $id = '',
```
christyjacob4, line 275:
```suggestion
        string $key = '',
```

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
+                }
+                break;
+            case Database::RELATION_ONE_TO_MANY:
+                $sql = "ALTER TABLE {$relatedTable} ADD COLUMN `{$twoWayKey}` {$sqlType} DEFAULT NULL;";
+                break;
+            case Database::RELATION_MANY_TO_ONE:
+                $sql = "ALTER TABLE {$table} ADD COLUMN `{$id}` {$sqlType} DEFAULT NULL;";
+                break;
+            case Database::RELATION_MANY_TO_MANY:
+                return true;
```
christyjacob4, line 301: Why do we return true here? Shouldn't an SQL query be executed for this relation type?

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
+            default:
+                throw new Exception('Invalid relationship type.');
+        }
```
christyjacob4, line 304:
```suggestion
        $sql = '';
        switch ($type) {
            case Database::RELATION_ONE_TO_ONE:
                $sql = "ALTER TABLE {$table} ADD COLUMN `{$id}` {$sqlType} DEFAULT NULL;";

                if ($twoWay) {
                    $sql .= "ALTER TABLE {$relatedTable} ADD COLUMN `{$twoWayKey}` {$sqlType} DEFAULT NULL;";
                }
                break;
            case Database::RELATION_ONE_TO_MANY:
                $sql = "ALTER TABLE {$relatedTable} ADD COLUMN `{$twoWayKey}` {$sqlType} DEFAULT NULL;";
                break;
            case Database::RELATION_MANY_TO_ONE:
                $sql = "ALTER TABLE {$table} ADD COLUMN `{$id}` {$sqlType} DEFAULT NULL;";
                break;
            case Database::RELATION_MANY_TO_MANY:
                break;
            default:
                throw new Exception('Invalid relationship type.');
        }
        
        if (empty($sql)) {
            return true;
        }
```

to be consistent with the `updateAttribute` method below, we can declare $sql to be an empty string outside the switch case and return `true` outside the switch case if its still empty

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
+
+        switch ($type) {
+            case Database::RELATION_ONE_TO_ONE:
+                if (!\is_null($newKey)) {
+                    $sql = "ALTER TABLE {$table} RENAME COLUMN `{$key}` TO `{$newKey}`;";
+                }
+                if ($twoWay && !\is_null($newTwoWayKey)) {
+                    $sql .= "ALTER TABLE {$relatedTable} RENAME COLUMN `{$twoWayKey}` TO `{$newTwoWayKey}`;";
+                }
+                break;
```
christyjacob4, line 355: What happens if we try to rename one of the keys in the relationship to the name of another attribute that already exists on the collection ? 

Example :

We have a movies collection that has an attribute `name`. Let's say it has a relation attribute `director`. Now I try to rename `director` to `name` . Since the attribute already exists will it throw an error or overwrite it ? 

If it throws an error, we should probably catch it and throw our own version of `DuplicateException` or something along those lines so that we dont expose any of the PDO internal error messages

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
-    public function deleteAttribute(string $collection, string $id, bool $array = false): bool
-    {
+    public function createRelationship(
+        string $collection,
+        string $relatedCollection,
+        string $type,
+        bool $twoWay = false,
+        string $id = '',
+        string $twoWayKey = '',
+        string $onDelete = Database::RELATION_MUTATE_RESTRICT
```
christyjacob4, line 277: I see `$onDelete` as one of the parameters but its not being used in this function 👀  Why is that ?

christyjacob4: Similarly in the other methods `updateRelationship` and `deleteRelationship`. as well ... How are the changes cascaded to the child collection then ?

src/Database/Adapter/MariaDB.php
```diff
@@ -246,35 +245,222 @@ public function updateAttribute(string $collection, string $id, string $type, in
...
+                if ($twoWay && $side === Database::RELATION_SIDE_CHILD) {
+                    $sql = "ALTER TABLE {$relatedTable} DROP COLUMN `{$twoWayKey}`;";
+                } else {
+                    $sql = "ALTER TABLE {$table} DROP COLUMN `{$key}`;";
+                }
+                break;
+            case Database::RELATION_MANY_TO_MANY:
+                $junction = $side === Database::RELATION_SIDE_PARENT
+                    ? $this->getSQLTable('_' . $collection . '_' . $relatedCollection)
+                    : $this->getSQLTable('_' . $relatedCollection . '_' . $collection);
```
christyjacob4, line 430: Where are these tables created? The `createRelationship` method does not seem to be doing this ?

src/Database/Database.php
```diff
@@ -1220,6 +1318,429 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+            }
+        }
+
+        if (\is_null($relationship)) {
+            throw new Exception('Attribute not found');
+        }
+
+        $collection->setAttribute('attributes', \array_values($attributes));
+
+        $relatedCollection = $relationship['options']['relatedCollection'];
```
christyjacob4, line 1666: I think it's important we store the keys of the `options` array to be constants. There is a high chance these can be overlooked.

src/Database/Database.php
```diff
@@ -1220,6 +1318,429 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+     * @throws Throwable
+     */
+    public function deleteRelationship(string $collection, string $id): bool
+    {
+        $collection = $this->silent(fn () =>$this->getCollection($collection));
+        $attributes = $collection->getAttribute('attributes', []);
+        $relationship = null;
+
+        foreach ($attributes as $name => $attribute) {
+            if ($attribute['$id'] === $id) {
```
christyjacob4, line 1653: If this is the child collection shouldn't we should be checking `if ($attribute['$id'] === $toWayKey)`.

I feel the current code only handles the case when a parent collection & its attribute is passed.

src/Database/Adapter/MariaDB.php
abnegate: `$id` is consistent within the library; I think we should keep concerns separate. In the future, we may need separation between `$key` and `$id` as well, in which case I think `$id` should be the ...key (to add to confusion 😅)

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
abnegate: Yep this will leave a dangling index in the metadata table, will fix :+1:

src/Database/Adapter.php
abnegate: I think it should be $id here and the update method as well, will update the inverse

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
abnegate: Yep but we are able to abstract this part to the `Database` level instead of the adapter since we just call `createCollection`: https://github.com/utopia-php/database/blob/feat-relationships/src/Database/Database.php#L1417-L1418

src/Database/Adapter/MariaDB.php
abnegate: We need to do that for the update and delete because there's a possibility it won't be assigned, but that's not the case here

src/Database/Adapter/MariaDB.php
abnegate: It's only used in the metadata so isn't needed at the adapter level - will remove the param here 👌

src/Database/Database.php
abnegate: If this is the child collection, then the `$id` will equal the `$twoWayKey` of the parent collection, this is explicitly tested here: https://github.com/utopia-php/database/blob/d8752a8a05d5989045a25248b2e219d673f94638/tests/Database/Base.php#L9512

src/Database/Adapter/MariaDB.php
christyjacob4: Okay got it 👍

src/Database/Adapter.php
christyjacob4: I see 👍

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
christyjacob4: okay got it

src/Database/Database.php
```diff
@@ -1221,147 +1342,603 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+            throw new LimitException('Column limit reached. Cannot create new attribute.');
+        }
 
-        $this->adapter->renameIndex($collection->getId(), $old, $new);
+        if (
+            $this->adapter->getDocumentSizeLimit() > 0 &&
+            $this->adapter->getAttributeWidth($collection) >= $this->adapter->getDocumentSizeLimit()
+        ) {
+            throw new LimitException('Row width limit reached. Cannot create new attribute.');
+        }
```
christyjacob4, line 1407: Don't we also need to do the same checks for the relatedCollection ?

christyjacob4 (commented):
Looks good. Just some more questions

src/Database/Database.php
abnegate: Yep will add them 👍

src/Database/Database.php
```diff
@@ -848,10 +908,49 @@ protected function validateDefaultTypes(string $type, mixed $default): void
...
+            throw new Exception('Index not found');
+        }
+
+        // Execute update from callback
+        $updateCallback($indexes[$index], $collection, $index);
+
+        // Save
+        $collection->setAttribute('indexes', $indexes, Document::SET_TYPE_ASSIGN);
+
+        $this->silent(fn () => $this->updateDocument(self::METADATA, $collection->getId(), $collection));
```
christyjacob4, line 937: Once the index metadata is updated, shouldn't its cache be cleared ?

abnegate: @christyjacob4 `updateDocument` purges the cache for the collection document

src/Database/Database.php
```diff
@@ -1221,146 +1348,604 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+                    '$id' => '_index_' . $twoWayKey,
+                    'key' => '_index_' . $twoWayKey,
+                    'type' => self::INDEX_KEY,
+                    'attributes' => [$twoWayKey],
+                ]),
+            ]));
         }
 
-        $this->trigger(self::EVENT_INDEX_RENAME, $indexNew);
+        $relationship = $this->adapter->createRelationship(
```
christyjacob4, line 1486: I believe we should check for `true` or `false` on the return type and throw an exception accordingly ?

src/Database/Database.php
```diff
@@ -1221,146 +1348,604 @@ public function renameAttribute(string $collection, string $old, string $new): b
...
+                    );
+
+                    $this->deleteDocument(self::METADATA, $junction);
+                    break;
+                default:
+                    throw new Exception('Invalid relationship type.');
+            }
+        });
+
+        $deleted = $this->adapter->deleteRelationship(
```
christyjacob4, line 1786: Shouldn't we use the returned `true` or `false` to throw an exception accordingly ?

christyjacob4 (approved):
Looks good 👍

src/Database/Database.php
abnegate: Yep, will update

src/Database/Database.php
abnegate: Yep, will update

src/Database/Database.php
```diff
@@ -1496,15 +2397,276 @@ public function createDocument(string $collection, Document $document): Document
...
         }
 
+        if ($this->resolveRelationships) {
+            $this->silent(fn () => $this->createDocumentRelationships($collection, $document));
+        }
+
         $document = $this->adapter->createDocument($collection->getId(), $document);
 
+        if ($this->resolveRelationships) {
+            $this->silent(fn () => $this->getDocumentRelationships($collection, $document));
```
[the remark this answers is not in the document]
abnegate: This is mutating the `$document` in the `getDocumentRelationships` call, but I'll rename the function and use the return value to make it more transparent 👌

PR #251, opened 2023-03-27
Revert internal

fogelito:


abnegate:
Closing this as we will keep the same to ensure we avoid race conditions

PR #252, opened 2023-03-27
Swoole test

fogelito:


abnegate:
@fogelito Closing for now, we can re-open when we look at this idea again

PR #258, opened 2023-04-10
Empty values operators

fogelito:


src/Database/Query.php
```diff
@@ -482,9 +482,13 @@ protected static function getMethodFromAlias(string $method): string
      * @param string $attribute
      * @param array<mixed> $values
      * @return Query
+     * @throws Exception
      */
     public static function equal(string $attribute, array $values): self
     {
+        if (empty($values)) {
+            throw new Exception('Equal operator requires values');
```
abnegate, line 490:
```suggestion
            throw new Exception('Equal queries require at least one value.');
```

src/Database/Query.php
```diff
@@ -506,9 +510,13 @@ public static function notEqual(string $attribute, mixed $value): self
      * @param string $attribute
      * @param mixed $value
      * @return Query
+     * @throws Exception
      */
     public static function lessThan(string $attribute, mixed $value): self
     {
+        if ($value === null) {
+            throw new Exception("lessThan operator can't be null");
```
abnegate, line 518:
```suggestion
            throw new Exception('Less than queries cannot have a null value.');
```

abnegate (commented):
Let's format the messages like the suggestions

src/Database/Query.php
```diff
@@ -482,9 +482,13 @@ protected static function getMethodFromAlias(string $method): string
      * @param string $attribute
      * @param array<mixed> $values
      * @return Query
+     * @throws Exception
      */
     public static function equal(string $attribute, array $values): self
     {
+        if (empty($values)) {
+            throw new Exception('Equal queries require at least one value.');
```
Meldiron, line 490: Should we throw error, or continue without a query?
As Appwrite developer, if I do `[Query.equal('userId', ids])` and `ids` are empty, it would be annoying to get an error. It would mean I need to do exactly the same check in MANY places - everywhere, where I put array into query.

To keep SDK and HTTP consistent, I would avoid fixing this in SDK. I think this should be fixed here, or in Appwrite repo. I think empty array of values should return everything, instead of throwing error.

fogelito: Did not realize the SDK fixing.. We can do it with a validator if we want to throw an error 
The question is Should we throw errors, or select all? especially now that we removed limits?
@abnegate what do you think

abnegate: I think we should throw an error. In that way, the developer knows the query has an issue. They might not realize it's not doing anything if we just ignore it. Make more sense to do it in the validator though instead of here

fogelito:
Will be taken care of here...
#264

PR #262, opened 2023-04-18
feat: collection level permissions

TorstenDittmann:
- implements `documentSecurity` to collections
- respect `documentSecurity` on all CRUD operations

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2825,16 +2825,19 @@ class Database
             $document = $this->silent(fn () => $this->updateDocumentRelationships($collection, $old, $document));
         }
 
-        $document = $this->adapter->updateDocument($collection->getId(), $document);
+        $this->adapter->updateDocument($collection->getId(), $document);
 
-        if ($this->resolveRelationships) {
-            $document = $this->silent(fn () => $this->populateDocumentRelationships($collection, $document));
-        }
+        // TODO: `updateDocumentRelationships` removes the attributes, and for 1-1 relations, it also removes the id that is needed for 1-1 relationships
+        // if ($this->resolveRelationships) {
+        //     $document = $this->silent(fn () => $this->populateDocumentRelationships($collection, $document));
+        // }
 
-        $document = $this->decode($collection, $document);
+        // $document = $this->decode($collection, $document);
 
         $this->purgeRelatedDocuments($collection, $id);
         $this->cache->purge('cache-' . $this->getNamespace() . ':' . $collection->getId() . ':' . $id . ':*');
+
+        $document = $this->silent(fn () => $this->getDocument($collection->getId(), $document->getId()));
 
         $this->trigger(self::EVENT_DOCUMENT_UPDATE, $document);
 
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2827,17 +2827,14 @@ class Database
 
         $this->adapter->updateDocument($collection->getId(), $document);
 
-        // TODO: `updateDocumentRelationships` removes the attributes, and for 1-1 relations, it also removes the id that is needed for 1-1 relationships
-        // if ($this->resolveRelationships) {
-        //     $document = $this->silent(fn () => $this->populateDocumentRelationships($collection, $document));
-        // }
+        if ($this->resolveRelationships) {
+            $document = $this->silent(fn () => $this->populateDocumentRelationships($collection, $document));
+        }
 
-        // $document = $this->decode($collection, $document);
+        $document = $this->decode($collection, $document);
 
         $this->purgeRelatedDocuments($collection, $id);
         $this->cache->purge('cache-' . $this->getNamespace() . ':' . $collection->getId() . ':' . $id . ':*');
-
-        $document = $this->silent(fn () => $this->getDocument($collection->getId(), $document->getId()));
 
         $this->trigger(self::EVENT_DOCUMENT_UPDATE, $document);
 
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3,6 +3,7 @@
 namespace Utopia\Database;
 
 use Exception;
+use InvalidArgumentException;
 use Throwable;
 use Utopia\Cache\Cache;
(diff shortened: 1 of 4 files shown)


src/Database/Database.php
```diff
@@ -3558,6 +3604,13 @@ public function find(string $collection, array $queries = [], ?int $timeout = nu
 
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
+        if ($collection->getAttribute('documentSecurity', true) === false) {
+            $authorization = new Authorization(self::PERMISSION_READ);
+            if ($authorization->isValid($collection->getRead())) {
+                $skipAuth = true;
+            }
+        }
```
stnguyen90, line 3612: I don't think this check should be only for `documentSecurity == true`. What if `documentSecurity` is `true` AND there are collection-level permissions?

Same for the `count()` method.

stnguyen90 (changes requested):

src/Database/Database.php
```diff
@@ -2090,9 +2113,11 @@ public function getDocument(string $collection, string $id, array $queries = [])
         if ($cache = $this->cache->load($cacheKey, self::TTL)) {
             $document = new Document($cache);
 
-            if ($collection->getId() !== self::METADATA
-                && !$validator->isValid($document->getRead())) {
-                return new Document();
+            if ($collection->getId() !== self::METADATA) {
+                $isValid = $documentSecurity ? $validator->isValid($document->getRead()) : $validator->isValid($collection->getRead());
```
stnguyen90, line 2117: Would this ignore collection level permissions if `documentSecurity` was `true`?

stnguyen90 (changes requested):

src/Database/Database.php
```diff
@@ -3809,10 +3905,18 @@ public function count(string $collection, array $queries = [], ?int $max = null)
             throw new Exception("Collection not found");
         }
 
+        if ($collection->getAttribute('documentSecurity', true) === false) {
+            $authorization = new Authorization(self::PERMISSION_READ);
+            if ($authorization->isValid($collection->getRead())) {
+                $skipAuth = true;
+            }
+        }
```
stnguyen90, line 3913: The outer if check should be removed here, right?

TorstenDittmann: good catch, leftover

src/Database/Database.php
```diff
@@ -3455,21 +3542,21 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
             case Database::RELATION_ONE_TO_ONE:
                 $this->skipRelationships(fn () =>
-                    $this->deleteDocument(
-                        $relatedCollection->getId(),
-                        $value->getId()
-                    ));
+                $this->deleteDocument(
+                    $relatedCollection->getId(),
+                    $value->getId()
+                ));
```
stnguyen90, line 3548: I think this needs to check if `$value == null` before deleting or else you we get:

image

src/Database/Database.php
```diff
@@ -3455,21 +3542,21 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
                 foreach ($value as $relation) {
                     $this->skipRelationships(fn () =>
-                        $this->deleteDocument(
-                            $relatedCollection->getId(),
-                            $relation->getId()
-                        ));
+                    $this->deleteDocument(
+                        $relatedCollection->getId(),
+                        $relation->getId()
+                    ));
```
stnguyen90, line 3559: Should this always skip? If it does, we'll have a limitation where related documents will only delete 1 level deep. For example, if i have:

> Quiz -> Questions -> Answer Choices

and I delete the Quiz, the Questions will be deleted, but the Answer Choices won't be.

stnguyen90:
When setting a relationship, if a document with the specified ID doesn't exist, the API call will still succeed, but the relationship will be set to null.

src/Database/Database.php
stnguyen90: I've created a task for us to track this.

src/Database/Database.php
```diff
@@ -2410,6 +2470,13 @@ public function createDocument(string $collection, Document $document): Document
     {
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
+        if ($collection->getId() !== self::METADATA) {
+            $authorization = new Authorization(self::PERMISSION_CREATE);
+            if (!$authorization->isValid($collection->getCreate())) {
```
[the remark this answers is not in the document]
stnguyen90: @fogelito, I don't think we need to for `createDocument()`, right? That should only be on the collection level.

stnguyen90 (approved):

tests/Database/Base.php
```diff
@@ -9966,6 +9971,663 @@ public function testManyToManyRelationshipKeyWithSymbols(): void
...
+     * @return array<Document>
+     */
+    public function testCollectionPermissionsCreateWorks(Document $collection): array
+    {
+        Authorization::cleanRoles();
+        Authorization::setRole(Role::users()->toString());
+
+        $document = static::getDatabase()->createDocument($collection->getId(), new Document([
+            '$id' => ID::unique(),
+            '$permissions' => [
```
fogelito, line 10015: If documentSecurity is false for $collection , do we need to pass permissions on createDocument?

src/Database/Database.php
```diff
@@ -651,6 +674,36 @@ public function createCollection(string $id, array $attributes = [], array $inde
...
+        }
+
+        $collection = $this->silent(fn () => $this->getCollection($id));
+
+        $collection
+            ->setAttribute('$permissions', $permissions)
+            ->setAttribute('documentSecurity', $documentSecurity);
+
+        $collection = $this->silent(fn () => $this->updateDocument(self::METADATA, $collection->getId(), $collection));
+
```
fogelito, line 701: Do we need to clean cache?

fogelito: Can you add a test for the method?

TorstenDittmann: This should be covered by `$this->updateDocument`. Adding a test 👍🏻

PR #263, opened 2023-04-21
Don't exceed max depth for relationships

stnguyen90:
Before this, if a document had a 1 to 1, parent-sided many-to-one, or child-sided one-to-many relationship, the nested document's ID would still be included in the result even if the user didn't have access to the nested document. This clears the attribute so that the document ID is excluded from the result.

This PR depends on:

* https://github.com/utopia-php/database/pull/262

We should merge the dependant to main and then change this one to merge to main.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2369,7 +2369,12 @@ class Database
                     break;
                 case Database::RELATION_MANY_TO_ONE:
                     if ($side === Database::RELATION_SIDE_PARENT) {
-                        if (\is_null($value) || $skipFetch) {
+                        if ($skipFetch || $this->relationshipFetchDepth === Database::RELATION_MAX_DEPTH) {
+                            $document->removeAttribute($key);
+                            break;
+                        }
+
+                        if (\is_null($value)) {
                             break;
                         }
                         $this->relationshipFetchDepth++;
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2328,10 +2328,11 @@ class Database
                     break;
                 case Database::RELATION_ONE_TO_MANY:
                     if ($side === Database::RELATION_SIDE_CHILD) {
-                        if (!$twoWay) {
+                        if (!$twoWay || $this->relationshipFetchDepth === Database::RELATION_MAX_DEPTH || $skipFetch) {
                             $document->removeAttribute($key);
+                            break;
                         }
-                        if ($twoWay && !\is_null($value) && !$skipFetch) {
+                        if (!\is_null($value)) {
                             $this->relationshipFetchDepth++;
                             $this->relationshipFetchMap[] = $relationship;
 
@@ -2345,7 +2346,7 @@ class Database
                         break;
                     }
 
-                    if ($twoWay && ($this->relationshipFetchDepth === Database::RELATION_MAX_DEPTH || $skipFetch)) {
+                    if ($this->relationshipFetchDepth === Database::RELATION_MAX_DEPTH || $skipFetch) {
                         break;
                     }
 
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -2251,11 +2251,8 @@ private function populateDocumentRelationships(Document $collection, Document $d
                         break;
                     }
 
-                    if ($skipFetch) {
+                    if ($skipFetch || $twoWay && ($this->relationshipFetchDepth === Database::RELATION_MAX_DEPTH)) {
```
abnegate, line 2254: We may need a similar change for many-to-one relationships 🤔

stnguyen90: > We may need a similar change for many-to-one relationships 🤔

I literally remembered this 30 minutes ago 😅

stnguyen90: @abnegate, I've updated the logic for many to one (from parent) and one to many (from child).

TorstenDittmann (approved):

tests/Database/Base.php
```diff
@@ -9299,32 +9304,32 @@ public function testExceedMaxDepth(): void
...
-        $this->assertEquals('level2', $level1['level2'][0]->getId());
-        $this->assertEquals(1, count($level1['level2'][0]['level3']));
-        $this->assertEquals('level3', $level1['level2'][0]['level3'][0]->getId());
-        $this->assertArrayNotHasKey('level4', $level1['level2'][0]['level3'][0]);
+        $level1 = static::getDatabase()->getDocument($level1Collection, 'level1');
+        $this->assertEquals(1, count($level1[$level2Collection]));
+        $this->assertEquals('level2', $level1[$level2Collection][0]->getId());
+        $this->assertEquals(1, count($level1[$level2Collection][0][$level3Collection]));
+        $this->assertEquals('level3', $level1[$level2Collection][0][$level3Collection][0]->getId());
+        $this->assertArrayNotHasKey($level4Collection, $level1[$level2Collection][0][$level3Collection][0]);
```
christyjacob4, line 9319: Why do we allow the creation of 3 level nested documents, if the 3rd level cant be accessed anyways ?

christyjacob4 (approved):
Looks good. Please check the comment

tests/Database/Base.php
stnguyen90: Honestly, I'm not sure why you're allowed to create the level 4 document. While working on this, I wondered if it was a bug. Looking at the code now, I don't even see where `$relationshipCreateDepth` is incremented 👀 @abnegate is this a bug we overlooked?

christyjacob4: Exactly my confusion... 👀

stnguyen90: Ah I see. It's incremented in `relateDocuments()`. 

https://github.com/utopia-php/database/blob/b5ac84e0c77145bd0a7f38718ad915729c64fa93/src/Database/Database.php#L2585-L2626

However, perhaps this `>` check should be `>=`. What do you think, @abnegate?

https://github.com/utopia-php/database/blob/b5ac84e0c77145bd0a7f38718ad915729c64fa93/src/Database/Database.php#L2472

abnegate: This should be `>` as `$relationshipCreateDepth` starts from 1, so you can create and read up to the third level but not the fourth

christyjacob4: Hmm in that case the test case in question is not correct right?

abnegate: The test here is okay, this is checking that the fourth level is not returned in the response from creating the document as expected. The issue presents here: 

https://github.com/utopia-php/database/blob/fix-nested-document-id/tests/Database/Base.php#L9407-L9411

When fetching the third level, the fourth level _is_ present, which means it was created, even though it was not included in the response initially

PR #218, opened 2022-11-01
Add Base Exception class

everly-gif:
Following the changes in this [PR](https://github.com/appwrite/appwrite/pull/4575), we need to add some base Database Exception that we could use to distinguish between our errors and native errors. This PR implements a base exception class.

PR #268, opened 2023-04-28
Feat base exception

abnegate:
Continuation of https://github.com/utopia-php/database/pull/218


src/Database/Adapter/MariaDB.php
```diff
@@ -867,15 +868,15 @@ public function deleteDocument(string $collection, string $id): bool
...
 
         try {
-            $stmt->execute() || throw new Exception('Failed to delete document');
-            $stmtPermissions->execute() || throw new Exception('Failed to clean permissions');
+            $stmt->execute() || throw new DatabaseException('Failed to delete document');
+            $stmtPermissions->execute() || throw new DatabaseException('Failed to clean permissions');
         } catch (\Throwable $th) {
             $this->getPDO()->rollBack();
-            throw new Exception($th->getMessage());
+            throw new DatabaseException($th->getMessage());
```
fogelito, line 875: I know it was like this but..
should we Throw the caught Exception?

abnegate: This is intentional so that we can catch our specific exception type on the Appwrite side 👍

fogelito (approved):

PR #269, opened 2023-05-02
Fix cascading deletes across multiple levels

abnegate:
## What's changed
- Fix cascade deletes not accounting for nested relationships

## Test plan
- Added test to check multi-level cascade is working

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -276,12 +276,12 @@ class Database
     /**
      * @var array<Document>
      */
-    private array $relationshipFetchMap = [];
+    private array $relationshipFetchStack = [];
 
     /**
      * @var array<Document>
      */
-    private array $relationshipDeleteMap = [];
+    private array $relationshipDeleteStack = [];
 
     /**
      * @param Adapter $adapter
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -273,10 +274,15 @@ class Database
...
     /**
-     * @var array<array<string, mixed>>
+     * @var array<Document>
      */
     private array $relationshipFetchMap = [];
 
+    /**
+     * @var array<Document>
+     */
+    private array $relationshipDeleteMap = [];
```
stnguyen90, line 284: Since this isn't really a map, can we rename it to `$relationshipDeleteStack` to make that clear?

src/Database/Database.php
```diff
@@ -3396,6 +3400,13 @@ private function deleteDocumentRelationships(Document $collection, Document $doc
             $onDelete = $relationship['options']['onDelete'];
             $side = $relationship['options']['side'];
 
+            $relationship->setAttribute('collection', $collection->getId());
+            $relationship->setAttribute('document', $document->getId());
+
+            if ($this->relationshipDeleteDepth >= Database::RELATION_MAX_DEPTH) {
```
stnguyen90, line 3406: I'm not sure if we should have a max depth for delete...The attribute itself is configured cascade so I would expect it to cascade regardless of Appwrite's max depth configuration. If we don't cascade fully, there will be messy orphaned documents.

abnegate: Yep makes sense

PR #270, opened 2023-05-04
Feat: Added minSize property to string attribute.

singhbhaskar:
## Description

Added minSize for string attribute for Issue:
https://github.com/appwrite/appwrite/issues/5403

## Testing
Build appwrite docker image with the change to test
* Ensure that minSize is added for string attribute.
* Attribute with minSize is enforced on document when created.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -802,7 +802,7 @@ class Database
      * @throws StructureException
      * @throws Throwable
      */
-    public function createAttribute(string $collection, string $id, string $type, int $size, bool $required, mixed $default = null, bool $signed = true, bool $array = false, string $format = null, array $formatOptions = [], array $filters = []): bool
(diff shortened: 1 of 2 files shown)


fogelito:
Hey, @singhbhaskar thank you for this PR 👍 
Database attributes usually do not have an option, I say usually because there is an option for it on MariaDB using 
CHECK (LENGTH(attr) >= 10);
In this case, using $formatOptions on Appwrite's side can be a nice solution as we do in the Integer attribute, as more community members will vote for this feature.

singhbhaskar:
Sure @fogelito ,
I will check the implementation for $formatOptions and update here. Thanks for the suggestion

singhbhaskar:
Hi @fogelito ,
I was able to implement without making changes to Uthopia as per your suggestion of formatOptions. Hence closing this PR. Thanks for the suggestion.

PR #279, opened 2023-05-31
Fulltext fixes

fogelito:


bin/tasks/load.php
```diff
@@ -246,20 +252,20 @@ function addArticle($database, Generator $faker): void
...
+            Permission::read(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::create(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::create(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::create(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::update(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::update(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::update(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::delete(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::delete(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
+            Permission::delete(Role::user('6477a84d4cf9a381' . $faker->randomNumber(4))),
```
abnegate, line 268: Do we need this change? Can we just make the random number longer?

abnegate (approved):

bin/tasks/load.php
fogelito: Just made it bigger for now. Some times with a small number I get this error :
PHP Fatal error:  Uncaught InvalidArgumentException: Role "user" must have an ID value. in /usr/src/code/src/Database/Database.php:2537

Done 👍

PR #280, opened 2023-06-05
Add support for label permissions

stnguyen90:
A user can be given a particular label and access to resources can be restricted to only users with that label.

### DIFF
--- a/src/Database/Validator/Label.php
+++ b/src/Database/Validator/Label.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Utopia\Database\Validator;
+
+class Label extends Key
+{
+    protected string $message = 'Value must be a valid string between 1 and 36 chars containing only alphanumeric chars';
+
+    /**
+     * Is valid.
+     *
+     * Returns true if valid or false if not.
+     *
+     * @param $value
+     *
+     * @return bool
+     */
+    public function isValid($value): bool
+    {
+        if (!parent::isValid($value)) {
+            return false;
+        }
+
+        // Valid chars: A-Z, a-z, 0-9
+        if (\preg_match('/[^A-Za-z0-9]/', $value)) {
+            return false;
+        }
+
+        return true;
+    }
+}
--- a/src/Database/Helpers/Role.php
+++ b/src/Database/Helpers/Role.php
@@ -141,6 +141,17 @@ class Role
     }
 
     /**
+     * Create a label role from the given ID
+     *
+     * @param string $identifier
+     * @return Role
+     */
+    public static function label(string $identifier): self
+    {
+        return new self('label', $identifier, '');
+    }
+
+    /**
      * Create an any satisfy role
      *
      * @return Role
--- a/src/Database/Validator/Roles.php
+++ b/src/Database/Validator/Roles.php
@@ -14,6 +14,7 @@ class Roles extends Validator
     public const ROLE_USER = 'user';
     public const ROLE_TEAM = 'team';
     public const ROLE_MEMBER = 'member';
+    public const ROLE_LABEL = 'label';
 
     public const ROLES = [
         self::ROLE_ANY,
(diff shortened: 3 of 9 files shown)


src/Database/Validator/Alphanumeric.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Utopia\Database\Validator;
+
+use Utopia\Validator\Text;
+
+class Alphanumeric extends Text
+{
+    /**
```
eldadfux, line 9: Can we reuse some other validators from the framework or maybe the key validator for consistent logic?

src/Database/Validator/Label.php
```diff
@@ -0,0 +1,17 @@
...
+
+class Label extends Alphanumeric
+{
+    /**
+     * Label constructor.
+     *
+     * Validate text is a valid label
+     *
+     */
+    public function __construct()
```
eldadfux, line 13: If we do have a dedicated label validator, not sure the `Alphanumeric` will be reused any time soon. I would vote against pre optimization.

src/Database/Validator/Alphanumeric.php
stnguyen90: I can re-use the `Text` validator, but the error message is very messy:

```json
{
    "message": "Invalid labels: Value must a valid array and Value must be a valid string and at least 1 chars and no longer than 36 chars and only consist of '0, 1, 2, 3, 4, 5, 6, 7, 8, 9, A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X, Y, Z, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y, z' chars",
    "code": 400,
    "type": "general_argument_invalid",
    "version": "dev"
}
```

`Key` validator would require additional logic to check for special characters that it's not really worth it:

```php
    public function isValid($value): bool
    {
        if (!\is_string($value)) {
            return false;
        }

        /** NOT NEEDED

        // no leading special characters
        $leading = \mb_substr($value, 0, 1);
        if ($leading === '_' || $leading === '.' || $leading === '-') {
            return false;
        }

        $isInternal = $leading === '$';


        if ($isInternal && !$this->allowInternal) {
            return false;
        }

        if ($isInternal) {
            $allowList = [ '$id', '$createdAt', '$updatedAt' ];

            // If exact match, no need for any further checks
            return \in_array($value, $allowList);
        }
        
        **/
        
        // Valid chars: A-Z, a-z, 0-9, underscore, hyphen, period
        if (\preg_match('/[^A-Za-z0-9\_\-\.]/', $value)) {        // this regex can be simpler too
            return false;
        }

        if (\mb_strlen($value) > 36) {
            return false;
        }

        return true;
    }
```

stnguyen90: Here are the two approaches:

```php
class Label extends Key
{
    protected string $message = 'Value must be a valid string between 1 and 36 chars containing only alphanumeric chars';

    /**
     * Is valid.
     *
     * Returns true if valid or false if not.
     *
     * @param $value
     *
     * @return bool
     */
    public function isValid($value): bool
    {
        if (!parent::isValid($value)) {
            return false;
        }

        // Valid chars: A-Z, a-z, 0-9
        if (\preg_match('/[^A-Za-z0-9]/', $value)) {
            return false;
        }
        
        return true;
    }
}
```

```php
class Label extends Text
{
    /**
     * Label constructor.
     */
    public function __construct()
    {
        parent::__construct(36, 1, \array_merge(Text::ALPHABET_UPPER, Text::ALPHABET_LOWER, Text::NUMBERS));
    }

    /**
     * Get Description
     *
     * Returns validator description
     *
     * @return string
     */
    public function getDescription(): string
    {
        return 'Value must be a valid string between 1 and 36 chars containing only alphanumeric chars';
    }
}
```

eldadfux: I'm happy with either of them. If we need the framework as a dependency just for the Text validator, I would lean towards the first solution.

eldadfux (approved):

PR #281, opened 2023-06-08
Fix: $permissions returns as object rather than array

2002Bishwajeet:
## What does this do

Fixes [#5661](https://github.com/appwrite/appwrite/issues/5661)

## Testing 

Unit and E2E testing

### DIFF
--- a/src/Database/Document.php
+++ b/src/Database/Document.php
@@ -79,7 +79,7 @@ class Document extends ArrayObject
      */
     public function getPermissions(): array
     {
-        return \array_unique($this->getAttribute('$permissions', []));
+        return \array_values(\array_unique($this->getAttribute('$permissions', [])));
     }
 
     /**
--- a/tests/Database/PermissionTest.php
+++ b/tests/Database/PermissionTest.php
@@ -293,6 +293,7 @@ class PermissionTest extends TestCase
             'update("user:123")',
             'delete("user:123")'
         ];
+
         $parsed = Permission::aggregate($permissions, Database::PERMISSIONS);
         $this->assertEquals(['read("any")',
             'read("user:123")',
--- a/tests/Database/PermissionTest.php
+++ b/tests/Database/PermissionTest.php
@@ -284,5 +284,20 @@ class PermissionTest extends TestCase
 
         $parsed = Permission::aggregate($permissions, [Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE]);
         $this->assertEquals(['update("any")', 'delete("any")'], $parsed);
+
+        $permissions = [
+            'read("any")',
+            'read("user:123")',
+            'read("user:123")',
+            'write("user:123")',
+            'update("user:123")',
+            'delete("user:123")'
+        ];
+        $parsed = Permission::aggregate($permissions, Database::PERMISSIONS);
+        $this->assertEquals(['read("any")',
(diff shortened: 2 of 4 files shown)


src/Database/Document.php
```diff
@@ -79,7 +79,7 @@ public function getCollection(): string
      */
     public function getPermissions(): array
     {
-        return \array_unique($this->getAttribute('$permissions', []));
+        return \array_values(\array_unique($this->getAttribute('$permissions', [])));
```
stnguyen90, line 82: This is a good approach, but one thing we might want to do is make sure duplicates aren't inserted into the database. What do you think, @abnegate?

abnegate: Agree, we should be able to do that by wrapping [this line](https://github.com/utopia-php/database/blob/d821d1f18dfe83b0350ed208be3658764aec99d7/src/Database/Helpers/Permission.php#L173) with `\array_values(\array_unique())` too

tests/Database/PermissionTest.php
```diff
@@ -284,5 +284,21 @@ public function testAggregation(): void
...
+            'update("user:123")',
+            'delete("user:123")'
+        ];
+
+        $parsed = Permission::aggregate($permissions, Database::PERMISSIONS);
+        $this->assertEquals(['read("any")',
+            'read("user:123")',
+            'create("user:123")',
+            'update("user:123")',
+            'delete("user:123")'], $parsed);
```
abnegate, line 302: Let's break on the first and last array entry to help readability:

```suggestion
        $this->assertEquals([
            'read("any")',
            'read("user:123")',
            'create("user:123")',
            'update("user:123")',
            'delete("user:123")',
        ], $parsed);
```

tests/Database/Validator/PermissionsTest.php
```diff
@@ -297,4 +297,44 @@ public function testInvalidPermissions(): void
         $this->assertFalse($object->isValid($permissions));
         $this->assertEquals('You can only provide up to 100 permissions.', $object->getDescription());
     }
+
+    /*
+     *  Test for checking duplicate methods input. The getPermissions should return an a list array
+      */
+    public function testDuplicateMethods(): void
+    {
+        $object = new Permissions();
```
abnegate, line 306: Let's use a meaningful name like `$validator` here

tests/Database/Validator/PermissionsTest.php
```diff
@@ -297,4 +297,44 @@ public function testInvalidPermissions(): void
...
+            ]
+        ]);
+        $this->assertTrue($object->isValid($document->getPermissions()));
+        $permissions = $document->getPermissions();
+        $this->assertEquals(5, count($permissions));
+        $this->assertEquals('read("any")', $permissions[0]);
+        $this->assertEquals('read("user:' . $user . '")', $permissions[1]);
+        $this->assertEquals('write("user:' . $user . '")', $permissions[2]);
+        $this->assertEquals('update("user:' . $user . '")', $permissions[3]);
+        $this->assertEquals('delete("user:' . $user . '")', $permissions[4]);
```
abnegate, line 338: Let's assert the arrays are equal here instead of the individual elements

abnegate (commented):
Looks good just some style fixes :+1:

abnegate (approved):

PR #286, opened 2023-06-23
Allow custom createdAt and updatedAt on createDocument

### LINKED ISSUE #285: 🚀 Feature: Allow value for createdAt & updatedAt on create/update Document
### 🔖 Feature description

This fields [auto generate](https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L2530C48-L2530C48) on insert / [update](https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L2857C9-L2858)

It can be usefull to force them via `data` like

```json
{
...otherFields,
"$createdAt": "string or timestamp"
}
```

### 🎤 Pitch

In case of migration for others platform, without this we need to inject data directly on MariaDB, or we'll loose this information or if we simply need to change it for any other reason

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/utopia-php/database/blob/main/CODE_OF_CONDUCT.md)

fliitor:
Closes #285

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2527,8 +2527,8 @@ class Database
         $document
             ->setAttribute('$id', empty($document->getId()) ? ID::unique() : $document->getId())
             ->setAttribute('$collection', $collection->getId())
-            ->setAttribute('$createdAt', $time)
-            ->setAttribute('$updatedAt', $time);
+            ->setAttribute('$createdAt', empty($document->getCreatedAt()) ? $time : $document->getCreatedAt())
+            ->setAttribute('$updatedAt', empty($document->getUpdatedAt()) ? $time : $document->getUpdatedAt());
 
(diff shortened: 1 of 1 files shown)


fanatic75:
@fliitor, can you please add or update an appropriate test case

fliitor:
@fanatic75 I'm on vacation until September 15. If anyone else wants to do it, I'd love to.

fliitor:
@fanatic75 I've never created a test case in php, where do I start?
I've looked [here ](https://github.com/utopia-php/database/tree/main/tests/Database)and there doesn't seem to be an existing test for document creation.

DH-555:
> This can be close because it's already implemented

What do you mean by already implemented?

PR #287, opened 2023-06-28
Feat 5376 improve the select query

faisalill:


src/Database/Database.php
```diff
@@ -2207,14 +2207,14 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
-            }
-        }
+        // if ($collection->getId() !== self::METADATA) {
+        //     if (!$validator->isValid([
+        //         ...$collection->getRead(),
+        //         ...($documentSecurity ? $document->getRead() : [])
+        //     ])) {
+        //         return new Document();
+        //     }
+        // }
```
faisalill, line 2217: Commented this as of now because it was returning an empty document when '$id' was queried. Will fix it soon.

TorstenDittmann (commented):
starting CI now as well 👍🏻

faisalill:
Fixes https://github.com/appwrite/appwrite/issues/5376

src/Database/Adapter/Postgres.php
```diff
@@ -1215,6 +1224,12 @@ protected function getAttributeProjection(array $selections, string $prefix = ''
             }
         }
 
+        foreach ($selections as $key => $value) {
+            if ($value == "\"id\"" || $value == "\"internalId\"" || $value == "\"permissions\"" || $value == "\"createdAt\"" || $value == "\"updatedAt\"") {
```
fogelito, line 1228: I think you can use here 
if(in_array($value, ['id'...]))

src/Database/Adapter/Postgres.php
```diff
@@ -1056,17 +1056,26 @@ public function find(string $collection, array $queries = [], ?int $limit = 25,
...
+            }
+            if (isset($value['_createdAt'])) {
+                $results[$key]['$createdAt'] = $value['_createdAt'];
+                unset($results[$key]['_createdAt']);
+            }
+            if (isset($value['_updatedAt'])) {
+                $results[$key]['$updatedAt'] = $value['_updatedAt'];
+                unset($results[$key]['_updatedAt']);
+            }
+            if (isset($value['_permissions'])) {
```
fogelito, line 1075: Please double chec me.. if isset of when a value is null , will give you false, 
so maybe we need to check array_key_exists?

src/Database/Database.php
```diff
@@ -2263,6 +2263,26 @@ public function getDocument(string $collection, string $id, array $queries = [])
 
         $this->trigger(self::EVENT_DOCUMENT_READ, $document);
 
+        foreach ($queries as $query) {
+            if ($query->getMethod() == Query::TYPE_SELECT) {
+                $queriedValues = $query->getValues();
+                $defaultKeys = ['$id', '$internalId', '$permissions', '$createdAt', '$updatedAt', '$collection'];
```
fogelito, line 2269: Can we get or add this list as Static from Database::?

src/Database/Database.php
```diff
@@ -3978,6 +3998,21 @@ public function find(string $collection, array $queries = [], ?int $timeout = nu
 
         $this->trigger(self::EVENT_DOCUMENT_FIND, $results);
 
+        // remove default keys $id and $permissions from the results
+
+        foreach ($queries as $query) {
+            if ($query->getMethod() === Query::TYPE_SELECT) {
+                foreach ($results as $result) {
```
fogelito, line 4005: Can we reuse the static list of internal params from the previous comment here?

src/Database/Database.php
```diff
@@ -4471,6 +4512,12 @@ private function validateSelections(Document $collection, array $queries): array
             }
         }
 
+        $keys[] = '$id';
+        $keys[] = '$internalId';
+        $keys[] = '$createdAt';
+        $keys[] = '$updatedAt';
+        $keys[] = '$permissions';
```
fogelito, line 4519: Are we missing $collection?
Can use the static list here?

tests/Database/Base.php
```diff
@@ -4156,7 +4175,7 @@ public function testOneToOneOneWayRelationship(): void
         $this->assertArrayNotHasKey('area', $person->getAttribute('library'));
 
         $person = static::getDatabase()->getDocument('person', 'person1', [
-            Query::select(['*', 'library.name'])
+            Query::select(['*', 'library.name', '$id'])
         ]);
 
         $this->assertEquals('Library 1', $person->getAttribute('library')->getAttribute('name'));
```
fogelito, line 4181: Can you check if '$id' will be returned?

src/Database/Adapter/Postgres.php
faisalill: Switched it to array_key_exists.

src/Database/Adapter/Postgres.php
faisalill: Changed the whole thing to this:
```php
$selections = array_filter($selections, function ($value) {
            $excludedFields = ["\"id\"", "\"internalId\"", "\"permissions\"", "\"createdAt\"", "\"updatedAt\""];
            return !in_array($value, $excludedFields);
        });
```

src/Database/Database.php
faisalill: Done

tests/Database/Base.php
faisalill: Done

src/Database/Database.php
faisalill: Adding querying `$collection`.

src/Database/Database.php
```diff
@@ -4336,7 +4378,13 @@ public function decode(Document $collection, Document $document, array $selectio
             }
 
             if (empty($selections) || \in_array($key, $selections) || \in_array('*', $selections)) {
-                $document->setAttribute($key, ($array) ? $value : $value[0]);
+                if ($key === '$createdAt' && $value[0] === null) {
```
fogelito, line 4381: Can you combine $createdAt, $updatedAt to in_array please?

src/Database/Adapter/Postgres.php
```diff
@@ -1215,6 +1224,14 @@ protected function getAttributeProjection(array $selections, string $prefix = ''
             }
         }
 
+        $stringsToRemove = ["\"id\"", "\"internalId\"", "\"permissions\"", "\"createdAt\"", "\"updatedAt\"", "\"collection\""];
```
fogelito, line 1227: You can use the static internals ids array here, with string manipulation in the in_array later on?

src/Database/Database.php
```diff
@@ -4335,8 +4377,12 @@ public function decode(Document $collection, Document $document, array $selectio
                 }
             }
 
-            if (empty($selections) || \in_array($key, $selections) || \in_array('*', $selections)) {
-                $document->setAttribute($key, ($array) ? $value : $value[0]);
+            if (empty($selections) || \in_array($key, $selections) || \in_array('*', $selections) || \in_array($key, ['$createdAt', '$updatedAt'])) {
+                if (($key === '$createdAt' || $key === '$updatedAt') && $value[0] === null) {
```
fogelito, line 4381: if(in_array($key, ['$createdAt', '$updatedAt']) && $value[0] === null){

src/Database/Adapter/Postgres.php
fogelito: I had in mind only to change in the foreach ($selections as $key => $value) 
the $value  in the iteration to match the value of the internal ids. 
without the array_map loop . 👍

src/Database/Adapter/Postgres.php
```diff
@@ -1215,6 +1224,17 @@ protected function getAttributeProjection(array $selections, string $prefix = ''
...
+        $parameterNames =  \array_map(function ($value) {
+            return \ltrim($value, '$');
+        }, Database::INTERNAL_PARAMETERS);
+
+        foreach ($selections as $key => $value) {
+            $valueUnquoted = \trim(\stripslashes($value), "\"");
+            if (\in_array($valueUnquoted, $parameterNames)) {
+                unset($selections[$key]);
+            }
+        }
```
abnegate, line 1236: Let's be consistent with mariadb here and add the internal attributes when selected instead of removing when they are not

src/Database/Database.php
```diff
@@ -2286,6 +2296,26 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
+                }
+
+                foreach ($defaultKeys as $defaultKey) {
+                    if ($document->isSet($defaultKey)) {
+                        $document->removeAttribute($defaultKey);
+                    }
+                }
+            }
+        }
+
```
abnegate, line 2318: This should be done before the call to `trigger`

src/Database/Database.php
```diff
@@ -4001,6 +4031,18 @@ public function find(string $collection, array $queries = [], ?int $timeout = nu
...
+        foreach ($queries as $query) {
+            if ($query->getMethod() === Query::TYPE_SELECT) {
+                foreach ($results as $result) {
+                    foreach (Database::INTERNAL_PARAMETERS as $parameter) {
+                        $result->removeAttribute($parameter);
+                    }
+                }
+            }
+        }
+
```
abnegate, line 4045: This should be done before the call to `trigger`

src/Database/Database.php
```diff
@@ -4001,6 +4031,18 @@ public function find(string $collection, array $queries = [], ?int $timeout = nu
 
         $this->trigger(self::EVENT_DOCUMENT_FIND, $results);
 
+        // remove default keys $id and $permissions from the results
+
+        foreach ($queries as $query) {
+            if ($query->getMethod() === Query::TYPE_SELECT) {
+                foreach ($results as $result) {
+                    foreach (Database::INTERNAL_PARAMETERS as $parameter) {
```
abnegate, line 4039: This will remove internal attributes even if they were selected, we need to make sure they are only removed when not selected

Let's add a test to cover this case as well

src/Database/Database.php
```diff
@@ -4335,8 +4377,12 @@ public function decode(Document $collection, Document $document, array $selectio
                 }
             }
 
-            if (empty($selections) || \in_array($key, $selections) || \in_array('*', $selections)) {
-                $document->setAttribute($key, ($array) ? $value : $value[0]);
+            if (empty($selections) || \in_array($key, $selections) || \in_array('*', $selections) || \in_array($key, ['$createdAt', '$updatedAt'])) {
```
abnegate, line 4380: Let's break this into new lines to make it easier to read:

```suggestion
            if (
                empty($selections) 
                || \in_array($key, $selections) 
                || \in_array('*', $selections) 
                || \in_array($key, ['$createdAt', '$updatedAt'])
            ) {
```

tests/Database/Base.php
```diff
@@ -2539,18 +2558,19 @@ public function testFindSelect(): void
...
             $this->assertArrayNotHasKey('active', $document);
+            $this->assertArrayNotHasKey('director', $document);
+            $this->assertArrayNotHasKey('$id', $document);
+            $this->assertArrayNotHasKey('$internalId', $document);
+            $this->assertArrayNotHasKey('$collection', $document);
+            $this->assertArrayNotHasKey('$createdAt', $document);
+            $this->assertArrayNotHasKey('$updatedAt', $document);
+            $this->assertArrayNotHasKey('$permissions', $document);
         }
     }
```
abnegate, line 2575: Let's add a test for the `find` method when selecting internal attributes

src/Database/Adapter/Mongo.php
```diff
@@ -1335,6 +1335,26 @@ protected function getAttributeProjection(array $selections, string $prefix = ''
         $projection['_createdAt'] = 1;
         $projection['_updatedAt'] = 1;
```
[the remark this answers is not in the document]
faisalill: Except _collection and _id rest have to be set coz.
* no _permissions returns a null document (https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L2238)
* no _uid will not convert the data types of the document
* no _createdAt and _updatedAt are required here (https://github.com/utopia-php/database/blob/main/src/Database/Adapter/Mongo.php#L971)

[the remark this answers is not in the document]
abnegate: Can we make it so that they are not required at this point? I think we should be able to if we can do it for MariaDB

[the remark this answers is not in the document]
faisalill: For all adapters _uid and _permissions are a must. Because of this:
* no _permissions returns a null document (https://github.com/utopia-php/database/blob/main/src/Database/Database.php#L2238)
* no _uid will not convert the data types of the document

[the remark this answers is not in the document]
faisalill: For Postgres I am querying all the internal attributes and if any internal attribute is in the query I am removing it from selection then before the document is returned I check which internal attributes were in the query and remove the internal attributes from the document. It's a mess ik. I am thinking of closing this PR and starting from scratch.

src/Database/Database.php
faisalill: https://github.com/faisalill/database/blob/feat-5376-improve-the-select-query/tests/Database/Base.php#L1042

Tests here check whether only the queried internal attribute is returned.

faisalill:
I will close this PR start from scratch and update u guys on the same.

PR #288, opened 2023-07-04
Primary key permission

fogelito:
We would like to changes in all _perms collection from using Varchar DocumentId => integer document id
For better performance and shrinking the index resources.
Since this is internal it would be relatively easier changes.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,7 +7,7 @@
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
     processIsolation="false"
-    stopOnFailure="true">
+    stopOnFailure="false">
     <testsuites>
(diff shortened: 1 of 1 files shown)


abnegate:
@fogelito What's the benefit of this change? Please add a description to the PR :pray:

fogelito:
We have a Benchmark for this on Asana

abnegate:
@fogelito Awesome let's add a migration PR in Appwrite for this as well

abnegate (approved):

fogelito:
Still need to make changes here as in the update for example we need to pass the internal id after read the original doc.
Need to think the affect on relations as well.
Currently original values saved as well.

abnegate:
@fogelito Cool will mark this as draft for now 👌

PR #290, opened 2023-07-17
Fix permission test

fogelito:


src/Database/Helpers/Permission.php
```diff
@@ -89,6 +89,10 @@ public static function parse(string $permission): self
         }
 
         $permission = $permissionParts[0];
+        if(!in_array($permission, Database::PERMISSIONS)){
+            throw new DatabaseException('Invalid permission Value');
```
abnegate, line 93:
```suggestion
            throw new DatabaseException('Invalid permission type: "' . $permission . '".');
```

tests/Database/PermissionTest.php
```diff
@@ -260,33 +260,58 @@ public function testInputFromRoles(): void
...
+        $this->assertEquals(['create("any")'], $parsed);
 
         $parsed = Permission::aggregate($permissions, [Database::PERMISSION_UPDATE, Database::PERMISSION_DELETE]);
-        $this->assertEquals(['update("any")', 'delete("any")'], $parsed);
+
+        // TODO looks like something here is not working proper
+        var_dump($parsed);
+        die;
+
+        $this->assertEquals(['create("any")', 'update("any")', 'delete("any")'], $parsed);
```
abnegate, line 314: I don't think we need to change this test, it is supposed to check that any `write` permission becomes create, update and delete instead (if create is allowed by the resource type)

tests/Database/PermissionTest.php
```diff
@@ -260,20 +260,40 @@ public function testInputFromRoles(): void
...
+        }
+
+        try {
+            Permission::parse('read("users/")');
+            $this->fail('Failed to throw Exception');
+        } catch (\Exception $e) {
+            $this->assertEquals('Dimension must not be empty', $e->getMessage());
+        }
+
+        try {
```
fogelito, line 291: @stnguyen90 I have restructured this test function.
The label test is passing while I guess it should fail?

abnegate: Why do we have a different format for role identifiers vs the rest 🤔

stnguyen90: @abnegate, are you asking why labels must be alphanumeric only? This is to allow for searching users using the role. If we allow non-alphanumeric characters, it may break up the word and cause problems with the full-text index.

tests/Database/PermissionTest.php
```diff
@@ -260,20 +260,40 @@ public function testInputFromRoles(): void
...
+            Permission::parse('read("label:alphanumeric-only")');
+            $this->fail('Failed to throw Exception');
+        } catch (\Exception $e) {
+            $this->assertEquals('Do we want this to fail?', $e->getMessage());
```
stnguyen90, line 295: yes, this should fail because the dash is not allowed.

fogelito: The test is currently passing not failing.
We can see it now after restructuring the test block.

Is there a validation somewhere to make it fail?

tests/Database/PermissionTest.php
abnegate: @stnguyen90 Do we have the same issues now searching for ID's that contain any of `.`, `_` or `-`?

stnguyen90: @abnegate, yes, except possibly _. mysql doesn't consider _ a stop character. We had a bug with how we treated _ but i think that should be resolved now.

tests/Database/PermissionTest.php
stnguyen90: Looking at the code, it only fails during validation, not parsing. We can remove this parse test.

FYI, here's the PR that added it: https://github.com/utopia-php/database/pull/280

stnguyen90 (commented):
See https://github.com/utopia-php/database/pull/290#discussion_r1290736404

tests/Database/PermissionTest.php
abnegate: That makes sense, will remove :+1:

tests/Database/PermissionTest.php
abnegate: @stnguyen90 Ah yes I remember. Please make a backlog task to investigate/discuss this further, would be good to have consistency between the two (and potentially fix the URI format issue for OAuth too) :+1:

src/Database/Helpers/Permission.php
```diff
@@ -89,6 +89,10 @@ public static function parse(string $permission): self
         }
 
         $permission = $permissionParts[0];
+        if (!in_array($permission, array_merge(Database::PERMISSIONS, [Database::PERMISSION_WRITE]))) {
+            throw new DatabaseException('Invalid permission type: "' . $permission . '".');
+        }
+
```
abnegate, line 95:
```suggestion

        if (!\in_array($permission, array_merge(Database::PERMISSIONS, [Database::PERMISSION_WRITE]))) {
            throw new DatabaseException('Invalid permission type: "' . $permission . '".');
        }
```

abnegate (approved):

PR #291, opened 2023-07-17
Fix database filters

christyjacob4:


abnegate:
Closing as we have ruled out this approach

PR #295, opened 2023-07-18
No Exception when no change in update document function

fanatic75:
We should not throw authorization exception when there is no change in update document coming document.
This is required for Relationships as we need to allow users to update document having related collection with no permission.

Part fix https://github.com/appwrite/appwrite/issues/5404

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2791,7 +2791,6 @@ class Database
         }
 
         $time = DateTime::now();
-        $document->setAttribute('$updatedAt', $time);
 
         $old = Authorization::skip(fn () => $this->silent(fn () => $this->getDocument($collection, $id))); // Skip ensures user does not need read permission for this
         $collection = $this->silent(fn () => $this->getCollection($collection));
@@ -2801,13 +2800,25 @@ class Database
         if ($collection->getId() !== self::METADATA) {
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
 
-            if (!$validator->isValid([
+            $skipPermission = true;
+            foreach ($document as $key=>$value) {
+                if ($old->getAttribute($key) instanceof Document) {
+                    continue;
+                }
+                if ($old->getAttribute($key) !== $value) {
+                    $skipPermission = false;
+                    break;
+                }
+            }
+            if (!$skipPermission && !$validator->isValid([
                 ...$collection->getUpdate(),
                 ...($documentSecurity ? $old->getUpdate() : [])
             ])) {
                 throw new AuthorizationException($validator->getDescription());
             }
         }
+
+        $document->setAttribute('$updatedAt', $time);
 
         // Check if document was updated after the request timestamp
         $oldUpdatedAt = new \DateTime($old->getUpdatedAt());
@@ -3010,16 +3021,16 @@ class Database
                             $removedDocuments = \array_diff($oldIds, $newIds);
 
                             foreach ($removedDocuments as $relation) {
-                                $relation = $this->skipRelationships(fn () => $this->getDocument(
+                                $relation = $this->skipRelationships(fn () => Authorization::skip(fn () => $this->getDocument(
                                     $relatedCollection->getId(),
                                     $relation
-                                ));
+                                )));
 
-                                $this->skipRelationships(fn () => $this->updateDocument(
+                                $this->skipRelationships(fn () => Authorization::skip(fn () =>$this->updateDocument(
                                     $relatedCollection->getId(),
                                     $relation->getId(),
                                     $relation->setAttribute($twoWayKey, null)
-                                ));
+                                ))); //removing relationship from 10 id document in collection B by updating it.
                             }
 
                             foreach ($value as $relation) {
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2801,9 +2801,9 @@ class Database
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
 
             $skipPermission = true;
-            //compare if the document any changes
+            // Compare if the document has any changes
             foreach ($document as $key=>$value) {
-                //skip the nested documents as they will be checked later in recursions.
+                // Skip the nested documents as they will be checked later in recursions.
                 if ($old->getAttribute($key) instanceof Document) {
                     continue;
                 }
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2796,7 +2796,7 @@ class Database
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
         $validator = new Authorization(self::PERMISSION_UPDATE);
-        $skipPermission = true;
+        $shouldUpdate = false;
 
         if ($collection->getId() !== self::METADATA) {
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
@@ -2808,11 +2808,11 @@ class Database
                     continue;
                 }
                 if ($old->getAttribute($key) !== $value) {
-                    $skipPermission = false;
+                    $shouldUpdate = true;
                     break;
                 }
             }
-            if (!$skipPermission && !$validator->isValid([
+            if ($shouldUpdate && !$validator->isValid([
                 ...$collection->getUpdate(),
                 ...($documentSecurity ? $old->getUpdate() : [])
             ])) {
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -2791,24 +2791,37 @@ public function updateDocument(string $collection, string $id, Document $documen
...
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
         $validator = new Authorization(self::PERMISSION_UPDATE);
 
         if ($collection->getId() !== self::METADATA) {
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
 
-            if (!$validator->isValid([
+            $skipPermission = true;
+            //compare if the document any changes
```
stnguyen90, line 2804: Let's be consistent with the style:

```suggestion
            // Compare if the document has any changes
```

src/Database/Database.php
```diff
@@ -2791,24 +2791,37 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+            foreach ($document as $key=>$value) {
+                //skip the nested documents as they will be checked later in recursions.
```
stnguyen90, line 2806:
```suggestion
                // Skip the nested documents as they will be checked later in recursions.
```

tests/Database/Base.php
```diff
@@ -2728,13 +2728,54 @@ public function testWritePermissionsUpdateFailure(Document $document): Document
...
+        $document = static::getDatabase()->createDocument('documents', new Document([
             'string' => 'text📝',
             'integer' => 5,
             'bigint' => 8589934592, // 2^33
             'float' => 5.55,
             'boolean' => true,
             'colors' => ['pink', 'green', 'blue'],
         ]));
+        Authorization::cleanRoles();
+        //no changes in document
```
stnguyen90, line 2761:
```suggestion
        // No changes in document
```

tests/Database/Base.php
```diff
@@ -2728,13 +2728,54 @@ public function testWritePermissionsUpdateFailure(Document $document): Document
...
+            '$id' => ID::custom($document->getId()),
+            'string' => 'text📝',
+            'integer' => 5,
+            'bigint' => 8589934592, // 2^33
+            'float' => 5.55,
+            'boolean' => true,
+            'colors' => ['pink', 'green', 'blue'],
+            '$collection' => 'documents',
+        ]);
+        $documentToUpdate->setAttribute('$createdAt', $document->getAttribute('$createdAt'));
```
stnguyen90, line 2772: `$createdAt` shouldn't change, right?

tests/Database/Base.php
```diff
@@ -2728,13 +2728,54 @@ public function testWritePermissionsUpdateFailure(Document $document): Document
...
+        $documentToUpdate->setAttribute('$updatedAt', $document->getAttribute('$updatedAt'));
```
stnguyen90, line 2773: It seems odd to allow someone without access to update`$updatedAt` when they don't have access, just by issuing an update with the same values 🧐

stnguyen90 (changes requested):

tests/Database/Base.php
fanatic75: Since we are creating a new document using the Document constructor, I am purposefully setting  createdAt and updatedAt as document constructor doesn't add those attributes. 
Update document requires both of those attributes for MongoDB in the coming document.

tests/Database/Base.php
fanatic75: We are not updating updateAt. We are just adding those attributes in the document that needs to be updated which is required for MongoDB.

fanatic75: Okay, due to relationships, there can be a change in a child document, so we cannot skip updating the document even if there is no change and have no permission to that collection

fanatic75: Skipping updating the time when there is no change in document

stnguyen90 (approved):
LGTM. I'll defer to @abnegate for another set of eyes.

tests/Database/Base.php
```diff
@@ -3485,6 +3527,7 @@ public function testCreatedAtUpdatedAtAssert(): void
         $document = static::getDatabase()->getDocument('created_at', 'uid123');
         $this->assertEquals(true, !$document->isEmpty());
         sleep(1);
```
fogelito, line 3529: No need for sleep on getDocument

fanatic75: That's not an addition in this PR. If you think we should drop it, we can remove that from a testcase in a refactor PR?

src/Database/Database.php
```diff
@@ -2791,24 +2791,39 @@ public function updateDocument(string $collection, string $id, Document $documen
...
 
         $time = DateTime::now();
-        $document->setAttribute('$updatedAt', $time);
-
         $old = Authorization::skip(fn () => $this->silent(fn () => $this->getDocument($collection, $id))); // Skip ensures user does not need read permission for this
+
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
         $validator = new Authorization(self::PERMISSION_UPDATE);
+        $skipPermission = true;
```
abnegate, line 2799: Let's rename this from `skipPermission` to `shouldUpdate`, it will make the usage more clear

src/Database/Database.php
```diff
@@ -2791,24 +2791,39 @@ public function updateDocument(string $collection, string $id, Document $documen
...
 
         if ($collection->getId() !== self::METADATA) {
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
 
-            if (!$validator->isValid([
+            // Compare if the document has any changes
+            foreach ($document as $key=>$value) {
+                // Skip the nested documents as they will be checked later in recursions.
+                if ($old->getAttribute($key) instanceof Document) {
```
abnegate, line 2807: Let's cache the old value so we only have to lookup the key once for the best case

tests/Database/Base.php
```diff
@@ -2728,13 +2729,54 @@ public function testWritePermissionsUpdateFailure(Document $document): Document
...
+
+
+    /**
+     * @depends testCreateDocument
+     */
+    public function testNoChangeUpdateDocumentWithoutPermission(Document $document): Document
+    {
+        Authorization::cleanRoles();
+        Authorization::setRole(Role::any()->toString());
+
```
abnegate, line 2751: Let's add only one blank line between code blocks

tests/Database/Base.php
```diff
@@ -2728,13 +2729,54 @@ public function testWritePermissionsUpdateFailure(Document $document): Document
...
+        $document = static::getDatabase()->createDocument('documents', new Document([
             'string' => 'text📝',
             'integer' => 5,
             'bigint' => 8589934592, // 2^33
             'float' => 5.55,
             'boolean' => true,
             'colors' => ['pink', 'green', 'blue'],
         ]));
+        Authorization::cleanRoles();
```
abnegate, line 2761: We don't need this both here and at the start of the function

tests/Database/Base.php
```diff
@@ -2728,13 +2729,54 @@ public function testWritePermissionsUpdateFailure(Document $document): Document
...
+            'integer' => 5,
+            'bigint' => 8589934592, // 2^33
+            'float' => 5.55,
+            'boolean' => true,
+            'colors' => ['pink', 'green', 'blue'],
+            '$collection' => 'documents',
+        ]);
+        $documentToUpdate->setAttribute('$createdAt', $document->getAttribute('$createdAt'));
+        $documentToUpdate->setAttribute('$updatedAt', $document->getAttribute('$updatedAt'));
+        $updatedDocument = static::getDatabase()->updateDocument('documents', $document->getId(), $documentToUpdate);
```
abnegate, line 2775: Instead of recreating the document with the same values, you can just try passing the `$document` returned from create directly to the update function

src/Database/Database.php
fanatic75: Done this

src/Database/Database.php
fanatic75: Here old is coming from the get document; it should already be cached; also, in every recursion call, $old changes.

tests/Database/Base.php
fanatic75: removed it from beginning

tests/Database/Base.php
fanatic75: passed $document directly.

fanatic75:
@abnegate  can you please re-review this

src/Database/Database.php
fanatic75: done this, added a variable for holding oldAttributeValue

abnegate (approved):

PR #296, opened 2023-07-19
Add method for creating a timeout for all find, sum, count queries in timeout supported dbs

fanatic75:
Asana Task : https://app.asana.com/0/1204806686347431/1205080260273691 

This feature creates a way to add a timeout in a static manner

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -22,6 +22,8 @@ abstract class Adapter
      */
     protected array $debug = [];
 
+    protected static ?int $timeout = null;
+
     /**
      * @param string $key
      * @param mixed $value
@@ -641,4 +643,39 @@ abstract class Adapter
      * @return int
      */
     abstract public function getMaxIndexLength(): int;
+
+
+    /**
+     * Set a global timeout for database queries in milliseconds.
+     *
+     * This function allows you to set a maximum execution time for all database
+     * queries executed using the library. Once this timeout is set, any database
+     * query that takes longer than the specified time will be automatically
+     * terminated by the library, and an appropriate error or exception will be
+     * raised to handle the timeout condition.
+     *
+     * @param int $milliseconds The timeout value in milliseconds for database queries.
+     * @return void
+     *
+     * @throws \Exception The provided timeout value must be greater than or equal to 0.
+    */
+    public static function setTimeoutForQueries(int $milliseconds): void
+    {
+        if ($milliseconds < 0) {
+            throw new Exception('The provided timeout value must be greater than or equal to 0.');
+        }
+        self::$timeout = $milliseconds;
+    }
+
+
+    /**
+     * Clears a global timeout for database queries.
+     *
+     * @return void
+     *
+    */
+    public static function clearTimeoutForQueries(): void
+    {
+        self::$timeout = null;
+    }
 }
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -1005,8 +1005,8 @@ class MariaDB extends SQL
             {$sqlLimit};
         ";
 
-        if ($timeout) {
-            $sql = $this->setTimeout($sql, $timeout);
+        if ($timeout || static::$timeout) {
+            $sql = $this->setTimeout($sql, $timeout ? $timeout : static::$timeout);
         }
 
         $stmt = $this->getPDO()->prepare($sql);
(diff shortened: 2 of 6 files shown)


src/Database/Adapter.php
```diff
@@ -641,4 +643,39 @@ abstract public function increaseDocumentAttribute(string $collection, string $i
...
+     * @throws \Exception The provided timeout value must be greater than or equal to 0.
+    */
+    public static function setTimeoutForQueries(int $milliseconds): void
+    {
+        if ($milliseconds < 0) {
+            throw new Exception('The provided timeout value must be greater than or equal to 0.');
+        }
+        self::$timeout = $milliseconds;
+    }
+
```
abnegate, line 669: Let's add just one blank line between functions

src/Database/Adapter.php
```diff
@@ -641,4 +643,39 @@ abstract public function increaseDocumentAttribute(string $collection, string $i
...
+     * @param int $milliseconds The timeout value in milliseconds for database queries.
+     * @return void
+     *
+     * @throws \Exception The provided timeout value must be greater than or equal to 0.
+    */
+    public static function setTimeoutForQueries(int $milliseconds): void
+    {
+        if ($milliseconds < 0) {
+            throw new Exception('The provided timeout value must be greater than or equal to 0.');
+        }
```
abnegate, line 666: Let's keep the logic and message consistent:

```suggestion
        if ($milliseconds <= 0) {
            throw new Exception('Timeout must be greater than 0');
        }
```

src/Database/Adapter/Postgres.php
```diff
@@ -1159,13 +1164,19 @@ public function sum(string $collection, string $attribute, array $queries = [],
...
-                WHERE {$permissions} AND " . implode(' AND ', $where) . "
-                {$limit}
-            ) table_count");
+        $sql = "SELECT SUM({$attribute}) as sum
+        FROM (
+            SELECT {$attribute}
+            FROM {$this->getSQLTable($name)} table_main
+            WHERE {$permissions} AND " . implode(' AND ', $where) . "
+            {$limit}
+        ) table_count";
```
abnegate, line 1173: Let's fix the indentation here

fanatic75:
@abnegate  I have made changes according to your comments. can you review again?

src/Database/Adapter/MariaDB.php
```diff
@@ -1146,16 +1150,20 @@ public function sum(string $collection, string $attribute, array $queries = [],
...
         $sqlWhere = !empty($where) ? 'where ' . implode(' AND ', $where) : '';
+        $sql = "
+        SELECT SUM({$attribute}) as sum
+        FROM (
+            SELECT {$attribute}
+            FROM {$this->getSQLTable($name)} table_main
+             " . $sqlWhere . "
+            {$limit}
+        ) table_count
+    ";
```
abnegate, line 1161: Let's fix the indentation here

src/Database/Adapter.php
```diff
@@ -22,6 +22,8 @@ abstract class Adapter
      */
     protected array $debug = [];
 
+    protected static ?int $timeout = null;
```
abnegate, line 25: Does this need to be static or could we use an instance var?

src/Database/Adapter/Postgres.php
```diff
@@ -1159,13 +1164,20 @@ public function sum(string $collection, string $attribute, array $queries = [],
...
-                {$limit}
-            ) table_count");
+        $sql = "SELECT SUM({$attribute}) as sum
+                FROM (
+                        SELECT {$attribute}
+                        FROM {$this->getSQLTable($name)} table_main
+                        WHERE {$permissions} AND " . implode(' AND ', $where) . "
+                        {$limit}
+                     )  
+                table_count";
```
abnegate, line 1174: Let's fix the indentation here too

src/Database/Adapter.php
fanatic75: We need to keep this static as the idea here is to set timeout once for the whole single worker and runs al queries with the timeout in that worker.

src/Database/Adapter/Postgres.php
fanatic75: fixed this

src/Database/Adapter/MariaDB.php
fanatic75: fixed this

abnegate (approved):

PR #297, opened 2023-07-26
Add support for batch creates and updates

christyjacob4:
* add supports for createDocuments, and updateDocuments


christyjacob4:
Reference : Batch updates using inserts
https://stackoverflow.com/questions/20255138/sql-update-multiple-records-in-one-query

src/Database/Adapter/MariaDB.php
```diff
@@ -616,6 +711,213 @@ public function updateDocument(string $collection, Document $document): Document
...
+                    if (!empty($updateClause)) {
+                        $updateClause .= ', ';
+                    }
+                    $updateClause .= "{$column} = VALUES({$column})";
+                }
+
+                $stmt = $this->getPDO()->prepare("
+                    INSERT INTO {$this->getSQLTable($name)} (" . \implode(", ", $columns) . ") 
+                    VALUES " . \implode(', ', $batchKeys) . "
+                    ON DUPLICATE KEY UPDATE $updateClause
```
fogelito, line 869: ON DUPLICATE KEY UPDATE 
It can make a problem if uid will be changed (like we wanted to do once in the attributes table in Appwrite?)

src/Database/Database.php
```diff
@@ -1446,6 +1450,58 @@ public function createDocument(string $collection, Document $document): Document
...
+
+        foreach ($documents as $key => $document) {
+            $document
+                ->setAttribute('$id', empty($document->getId()) ? ID::unique() : $document->getId())
+                ->setAttribute('$collection', $collection->getId())
+                ->setAttribute('$createdAt', $time)
+                ->setAttribute('$updatedAt', $time);
+
+            $document = $this->encode($collection, $document);
+
```
fogelito, line 1485: Do we care about relational attributes?

src/Database/Database.php
```diff
@@ -1446,6 +1450,58 @@ public function createDocument(string $collection, Document $document): Document
...
+            return [];
+        }
+
+        $collection = $this->silent(fn() => $this->getCollection($collection));
+
+
+
+        $time = DateTime::now();
+
+        foreach ($documents as $key => $document) {
```
fogelito, line 1477: Is it possible to call self::createDocument() and accumulate the docs?

src/Database/Adapter/MariaDB.php
abnegate: What will be the problem? I thought it would update because we have `_uid = VALUES(_uid)` in the `$updateClause?

src/Database/Adapter/MariaDB.php
```diff
@@ -1016,7 +1318,7 @@ protected function getSQLCondition(Query $query): string
                  * Prepend wildcard by default on the back.
                  */
                 $value = $this->getSQLValue($query->getMethod(), $value);
-                return 'MATCH('.$attribute.') AGAINST ('.$this->getPDO()->quote($value).' IN BOOLEAN MODE)';
+                return 'MATCH(' . $attribute . ') AGAINST (' . $this->getPDO()->quote($value) . ' IN BOOLEAN MODE)';
```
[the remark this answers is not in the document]
abnegate: I think we don't need that change since this is based of an old version (0.30.0) instead of main

src/Database/Database.php
```diff
@@ -1446,6 +1450,58 @@ public function createDocument(string $collection, Document $document): Document
...
+     */
+    public function createDocuments(string $collection, array $documents, int $batchSize = self::INSERT_BATCH_SIZE): array
+    {
+        if (empty($documents)) {
+            return [];
+        }
+
+        $collection = $this->silent(fn() => $this->getCollection($collection));
+
+
```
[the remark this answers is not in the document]
abnegate: - Removed
- We don't have the METADATA checks in this version

src/Database/Database.php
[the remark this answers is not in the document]
abnegate: Not for create document :+1:

abnegate: There are no relations in this version, if/when we merge this up to main, we'll have to add support here :+1:

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
fogelito: yep 👍

src/Database/Database.php
fogelito: yep, old version again 👍

src/Database/Database.php
fogelito: ok 👍

fogelito (approved):

PR #298, opened 2023-07-27
Fix 5871 relationships with non existing documents

fanatic75:
This PR fixes bug in multiple relationships 
- One to One One way
- One to One Two way (Bug was not reproducible but for consistency added code as well)
- One to Many One Way and Two way
- Many to One one way and two way 
- Many to many one way and two way
Fixes https://github.com/appwrite/appwrite/issues/5871

Previously we were getting 500 for one to one two way as the non existing doc needs to be updated in One to One two-way relationship type and non existing doc does not have key.
 Now we are keeping it consistent with other relationship types and skipping non existing doc

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2995,7 +2995,14 @@ class Database
                 switch ($relationType) {
                     case Database::RELATION_ONE_TO_ONE:
                         if (!$twoWay) {
-                            if ($value instanceof Document) {
+                            if (\is_string($value)) {
+                                $related = $this->getDocument($relatedCollection->getId(), $value);
+                                if ($related->isEmpty()) {
+                                    //If no such document exists in related collection
+                                    //For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                    $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
+                                }
+                            } elseif ($value instanceof Document) {
                                 $relationId = $this->relateDocuments(
                                     $collection,
                                     $relatedCollection,
@@ -3133,6 +3140,10 @@ class Database
                                         $this->getDocument($relatedCollection->getId(), $relation)
                                     );
 
+                                    if ($related->isEmpty()) {
+                                        continue;
+                                    }
+
                                     $this->skipRelationships(fn () => $this->updateDocument(
                                         $relatedCollection->getId(),
                                         $related->getId(),
@@ -3233,7 +3244,7 @@ class Database
 
                         foreach ($value as $relation) {
                             if (\is_string($relation)) {
-                                if (\in_array($relation, $oldIds)) {
+                                if (\in_array($relation, $oldIds) || $this->getDocument($relatedCollection->getId(), $relation)->isEmpty()) {
                                     continue;
                                 }
                             } elseif ($relation instanceof Document) {
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2999,8 +2999,8 @@ class Database
                                 $related = $this->getDocument($relatedCollection->getId(), $value);
                                 if ($related->isEmpty()) {
                                     // If no such document exists in related collection
-                                    // For one-one we need to update the related key to old value, either null if no relation exists or old related document id
-                                    $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
+                                    // For one-one we need to update the related key to null if no relation exists
+                                    $document->setAttribute($key, null);
                                 }
                             } elseif ($value instanceof Document) {
                                 $relationId = $this->relateDocuments(
@@ -3024,8 +3024,8 @@ class Database
                                 $related = $this->skipRelationships(fn () => $this->getDocument($relatedCollection->getId(), $value));
                                 if ($related->isEmpty()) {
                                     // If no such document exists in related collection
-                                    // For one-one we need to update the related key to old value, either null if no relation exists or old related document id
-                                    $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
+                                    // For one-one we need to update the related key to null if no relation exists
+                                    $document->setAttribute($key, null);
                                     break;
                                 }
                                 if (
@@ -3184,9 +3184,9 @@ class Database
                         if (\is_string($value)) {
                             $related = $this->getDocument($relatedCollection->getId(), $value);
                             if ($related->isEmpty()) {
-                                //If no such document exists in related collection
-                                //For one-one we need to update the related key to old value, either null if no relation exists or old related document id
-                                $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
+                                // If no such document exists in related collection
+                                // For many-one we need to update the related key to null if no relation exists
+                                $document->setAttribute($key, null);
                             }
                             $this->deleteCachedDocument($relatedCollection->getId(), $value);
                         } elseif ($value instanceof Document) {
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3177,6 +3177,12 @@ class Database
                         }
 
                         if (\is_string($value)) {
+                            $related = $this->getDocument($relatedCollection->getId(), $value);
+                            if ($related->isEmpty()) {
+                                //If no such document exists in related collection
+                                //For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
+                            }
                             $this->deleteCachedDocument($relatedCollection->getId(), $value);
                         } elseif ($value instanceof Document) {
                             $related = $this->getDocument($relatedCollection->getId(), $value->getId());
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3022,7 +3022,12 @@ class Database
                         switch (\gettype($value)) {
                             case 'string':
                                 $related = $this->skipRelationships(fn () => $this->getDocument($relatedCollection->getId(), $value));
-
+                                if ($related->isEmpty()) {
+                                    // If no such document exists in related collection
+                                    // For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                    $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
+                                    break;
+                                }
                                 if (
                                     $oldValue?->getId() !== $value
                                     && $this->skipRelationships(fn () => $this->findOne($relatedCollection->getId(), [
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2998,8 +2998,8 @@ class Database
                             if (\is_string($value)) {
                                 $related = $this->getDocument($relatedCollection->getId(), $value);
                                 if ($related->isEmpty()) {
-                                    //If no such document exists in related collection
-                                    //For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                    // If no such document exists in related collection
+                                    // For one-one we need to update the related key to old value, either null if no relation exists or old related document id
                                     $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
                                 }
                             } elseif ($value instanceof Document) {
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -2995,7 +2995,14 @@ private function updateDocumentRelationships(Document $collection, Document $old
                 switch ($relationType) {
                     case Database::RELATION_ONE_TO_ONE:
                         if (!$twoWay) {
-                            if ($value instanceof Document) {
+                            if (\is_string($value)) {
```
fanatic75, line 2998: This is for one to one, one way

src/Database/Database.php
```diff
@@ -3133,6 +3140,10 @@ private function updateDocumentRelationships(Document $collection, Document $old
                                         $this->getDocument($relatedCollection->getId(), $relation)
                                     );
 
+                                    if ($related->isEmpty()) {
```
fanatic75, line 3143: This is for one to many, one way and two way both.

src/Database/Database.php
```diff
@@ -3166,6 +3177,12 @@ private function updateDocumentRelationships(Document $collection, Document $old
                         }
 
                         if (\is_string($value)) {
+                            $related = $this->getDocument($relatedCollection->getId(), $value);
```
fanatic75, line 3180: This is for many to one, one way and two way both

src/Database/Database.php
```diff
@@ -3233,7 +3250,7 @@ private function updateDocumentRelationships(Document $collection, Document $old
 
                         foreach ($value as $relation) {
                             if (\is_string($relation)) {
-                                if (\in_array($relation, $oldIds)) {
+                                if (\in_array($relation, $oldIds) || $this->getDocument($relatedCollection->getId(), $relation)->isEmpty()) {
```
fanatic75, line 3253: This is for many to many, one way and two way both

src/Database/Database.php
```diff
@@ -2995,7 +2995,14 @@ private function updateDocumentRelationships(Document $collection, Document $old
                 switch ($relationType) {
                     case Database::RELATION_ONE_TO_ONE:
                         if (!$twoWay) {
-                            if ($value instanceof Document) {
+                            if (\is_string($value)) {
+                                $related = $this->getDocument($relatedCollection->getId(), $value);
+                                if ($related->isEmpty()) {
+                                    //If no such document exists in related collection
+                                    //For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                    $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
```
stnguyen90, line 3003: @abnegate, do we want to restore it or leave it as is?

@fanatic75, 

1. What if `$oldValue` isn't a `Document` (Can it ever just be an ID)? 
2. What if `$oldValue` is empty because the user can't access the related document?
3. What if `$oldValue` is empty because we hit the max depth?

fanatic75: We need to restore whatever was earlier for one to one as one to one updates the value directly in the parent table otherwise the bug remains the same, it gets updated with a non existing document as the document is returned from this function and goes to adapter call. 

1. Old value will always be a document. The only case when it's not a document is when the relation doesn't exist prior and then we set null itself as you can see in the condition. 

If old value is ever empty we set null again. We are adding null at the end of ternary condition.

fanatic75: Also for this bug there cannot be a higher depth of relationship than 1 because we are doing something like this 

Database.updateDocument('db','collection','doc', { b : 'id_value'});

To have a higher degree than 1 we need to pass an object with $id and if $id is present it creates a new document so that document won't be non existing anymore

stnguyen90 (commented):
https://github.com/utopia-php/database/pull/298/files#r1278173401

src/Database/Database.php
fanatic75: I think I am not able to follow you. Can you share some more context, maybe a way to reproduce what you are saying so I can test it out on my side? 
Technically for one-to-one and many-to-one, the parent side will store the related document id in the table and if it does, it cannot be null in the database, so we will just restore the same value.  
But some reproducing steps would really help for your case.

stnguyen90: Create 5 collections with two-way one to one relationships and then create a document in each and link them.  When you fetch, the API will only return the 1st 3 levels.

stnguyen90 (commented):
https://github.com/utopia-php/database/pull/298#discussion_r1281223984

src/Database/Database.php
fanatic75: This bug is not reproducible in two-way one-to-one because, right now, two-way one-to-one cannot be updated directly. 
```
databases.updateDocument(
  'test', 'a-collection', 'a-doc', {
    b:'new-b',
  })
```

But I did test your scenario on Many to One two-way with five levels. It's fine because it will only fetch the first three levels. They will only be updated. So there will be no update of the old value above max depth.

fanatic75: I have metioned in PR description that one to one two way does not need any change and the reason why

tests/Database/Base.php
```diff
@@ -4095,6 +4095,23 @@ public function testOneToOneOneWayRelationship(): void
...
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Person 1',
+            '$collection' => 'person',
+            'library' => 'no-library'
+        ]));
+
+        $person1Document = static::getDatabase()->getDocument('person', 'person1');
+        // Assert document does not contain non existing relation document.
+        $this->assertEquals('library1', $person1Document->getAttribute('library')->getAttribute('$id'));
```
abnegate, line 4113: `$person1Document->getAttribute('library')` should equal `null`

tests/Database/Base.php
```diff
@@ -4510,6 +4527,23 @@ public function testOneToOneTwoWayRelationship(): void
...
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'England',
+            '$collection' => 'country',
+            'city' => 'no-city'
+        ]));
+
+        $country1Document = static::getDatabase()->getDocument('country', 'country1');
+        // Assert document does not contain non existing relation document.
+        $this->assertEquals('city1', $country1Document->getAttribute('city')->getAttribute('$id'));
```
abnegate, line 4545: `$person1Document->getAttribute('city')` should equal `null`

tests/Database/Base.php
```diff
@@ -4510,6 +4527,23 @@ public function testOneToOneTwoWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('country', 'country1', new Document([
+            '$id' => 'country1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'England',
+            '$collection' => 'country',
```
abnegate, line 4539: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

tests/Database/Base.php
```diff
@@ -5135,6 +5169,26 @@ public function testOneToManyOneWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('artist', 'artist1', new Document([
+            '$id' => 'artist1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Artist 1',
+            '$collection' => 'artist',
```
abnegate, line 5181: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

tests/Database/Base.php
```diff
@@ -5513,6 +5567,27 @@ public function testOneToManyTwoWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('customer', 'customer1', new Document([
+            '$id' => 'customer1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Customer 1',
+            '$collection' => 'customer',
```
abnegate, line 5580: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

tests/Database/Base.php
```diff
@@ -5954,6 +6029,23 @@ public function testManyToOneOneWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('review', 'review1', new Document([
+            '$id' => 'review1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Review 1',
+            '$collection' => 'review',
```
abnegate, line 6041: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

tests/Database/Base.php
```diff
@@ -6281,6 +6373,23 @@ public function testManyToOneTwoWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('product', 'product1', new Document([
+            '$id' => 'product1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Product 1',
+            '$collection' => 'product',
```
abnegate, line 6385: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

tests/Database/Base.php
```diff
@@ -6741,6 +6850,26 @@ public function testManyToManyOneWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('playlist', 'playlist1', new Document([
+            '$id' => 'playlist1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Playlist 1',
+            '$collection' => 'playlist',
```
abnegate, line 6862: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

tests/Database/Base.php
```diff
@@ -7032,6 +7161,26 @@ public function testManyToManyTwoWayRelationship(): void
...
+        // Update a document with non existing related document. It should not get added to the list.
+        static::getDatabase()->updateDocument('students', 'student1', new Document([
+            '$id' => 'student1',
+            '$permissions' => [
+                Permission::read(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'name' => 'Student 1',
+            '$collection' => 'students',
```
abnegate, line 7173: We can remove the `$collection` from here, it will be implicitly set in `updateDocument`

src/Database/Database.php
```diff
@@ -3015,7 +3022,12 @@ private function updateDocumentRelationships(Document $collection, Document $old
                         switch (\gettype($value)) {
                             case 'string':
                                 $related = $this->skipRelationships(fn () => $this->getDocument($relatedCollection->getId(), $value));
-
+                                if ($related->isEmpty()) {
+                                    // If no such document exists in related collection
+                                    // For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                    $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
```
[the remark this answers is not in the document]
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

fanatic75: done this

src/Database/Database.php
```diff
@@ -3166,6 +3182,12 @@ private function updateDocumentRelationships(Document $collection, Document $old
                         }
 
                         if (\is_string($value)) {
+                            $related = $this->getDocument($relatedCollection->getId(), $value);
+                            if ($related->isEmpty()) {
+                                //If no such document exists in related collection
+                                //For one-one we need to update the related key to old value, either null if no relation exists or old related document id
+                                $document->setAttribute($key, ($oldValue instanceof Document && !($oldValue->isEmpty()) ? $oldValue->getId() : null));
```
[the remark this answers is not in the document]
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

tests/Database/Base.php
fanatic75: done this

abnegate (approved):

PR #300, opened 2023-07-31
Fix crud operations permission in update relationships

fanatic75:
- Skipping authorization for crud operations which are integral parts of logic but not done by the user
- Added a test case.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2885,10 +2885,6 @@ class Database
             return $attribute['type'] === Database::VAR_RELATIONSHIP;
         });
 
-        $relationships = \array_filter($collection->getAttribute('attributes', []), function ($attribute) {
-            return $attribute['type'] === Database::VAR_RELATIONSHIP;
-        });
-
         $validator = new Authorization(self::PERMISSION_UPDATE);
         $shouldUpdate = false;
 
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2911,7 +2911,7 @@ class Database
                     // Skip the nested documents as they will be checked later in recursions.
                     if ($oldAttributeValue instanceof Document) {
                         continue;
-                    } elseif (\is_array($oldAttributeValue) && \is_array($value) && !($value instanceof Document)) {
+                    } elseif (\is_array($oldAttributeValue) && \is_array($value)) {
                         if ($compareChanges($value, $oldAttributeValue)) {
                             $isDifferent = true;
                         }
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3285,12 +3285,11 @@ class Database
                                             $relation->setAttribute($twoWayKey, $document->getId())
                                         );
                                     } else {
-                                        // We need to skip the relationship when updating document so that comparison check does not fail
-                                        $this->skipRelationships(fn () => $this->updateDocument(
+                                        $this->updateDocument(
                                             $relatedCollection->getId(),
                                             $related->getId(),
                                             $relation->setAttribute($twoWayKey, $document->getId())
-                                        ));
+                                        );
                                     }
                                 } else {
                                     throw new DatabaseException('Invalid value for relationship');
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2793,7 +2793,7 @@ class Database
         }
 
         // Try to get the related document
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -2995,7 +3011,14 @@ private function updateDocumentRelationships(Document $collection, Document $old
                 switch ($relationType) {
                     case Database::RELATION_ONE_TO_ONE:
                         if (!$twoWay) {
-                            if ($value instanceof Document) {
+                            if (\is_string($value)) {
+                                $related = Authorization::skip(fn () => $this->getDocument($relatedCollection->getId(), $value));
```
stnguyen90, line 3015: What if the user doesn't have read access to the related document? 🧐

stnguyen90 (changes requested):

src/Database/Database.php
fanatic75: Why should the user have read access? The user is only performing an update operation. 

For eg ,
If a user has update permission on collection B,  
On a relationship A > B, and they are updating B via A, they should still be able to update B without having read permission on it.

src/Database/Database.php
```diff
@@ -2881,20 +2881,34 @@ public function updateDocument(string $collection, string $id, Document $documen
         $old = Authorization::skip(fn () => $this->silent(fn () => $this->getDocument($collection, $id))); // Skip ensures user does not need read permission for this
 
         $collection = $this->silent(fn () => $this->getCollection($collection));
+        $relationships = \array_filter($collection->getAttribute('attributes', []), function ($attribute) {
+            return $attribute['type'] === Database::VAR_RELATIONSHIP;
+        });
+
+        $relationships = \array_filter($collection->getAttribute('attributes', []), function ($attribute) {
+            return $attribute['type'] === Database::VAR_RELATIONSHIP;
+        });
```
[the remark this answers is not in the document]
fanatic75: Yes, this has been fixed. It came due to merging a branch.

src/Database/Database.php
abnegate: Now that we are setting the related document to null here if it does not exist or they don't have permission, we need to make sure auth is enabled here, so that when they read it, they still get null instead of an empty document

stnguyen90 (commented):
Is this ready for review? Also, it looks like there are merge conflicts.

abnegate:
Closing as for now we don't want to allow relating to documents without read permission

fanatic75:
@abnegate  
I have removed the ability to skip authentication when fetching related documents. However, we still need to perform certain operations, such as updating removed documents from the relation without authentication. 

Let's consider this scenario: Level 1 has a one-to-many relationship with Level 2. If we create new documents in Level 2 using Level 1, we should not need update permission in Level 2. However, due to updating old documents, we are unable to create new documents without update permission. This pull request resolves this issue.

src/Database/Database.php
```diff
@@ -3285,11 +3285,12 @@ private function updateDocumentRelationships(Document $collection, Document $old
                                             $relation->setAttribute($twoWayKey, $document->getId())
                                         );
                                     } else {
-                                        $this->updateDocument(
+                                        // We need to skip the relationship when updating document so that comparison check does not fail
+                                        $this->skipRelationships(fn () => $this->updateDocument(
```
abnegate, line 3289: Can you explain this a bit more? Why does having the relationships populated here fail the comparison check?

tests/Database/Base.php
```diff
@@ -11843,6 +11843,67 @@ public function testCollectionPermissionsRelationshipsDeleteWorks(array $data):
...
+            [
+                '$id' => 'child1',
+            ],
+            [
+                '$id' => 'child2',
+            ],
+        ]);
+        $updatedParent = static::getDatabase()->updateDocument('parentRelationTest', 'parent1', $parent);
+
+        $this->assertEquals(2, \count($updatedParent['children']));
```
abnegate, line 11901: This test would pass without the changes since we are only adding an item to the relationship, let's add a case where we remove an item from the relationship as well

src/Database/Database.php
fanatic75: When updating the document with a two-way key, it's important to note that the key will not appear in the child document when the old version of the document is fetched. The key is simply used for the parent document to reference the child document.

fanatic75: I have removed this for now.

abnegate (approved):

stnguyen90 (approved):

PR #301, opened 2023-08-01
fix comparison check in updateDocument

fanatic75:
Previously the comparison check between old document and new document in updateDocument was not able to check if the key is a relationship key or not so it was creating issues when updating document with relationships.

Added test case

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2902,11 +2902,8 @@ class Database
                 if (array_key_exists($key, $relationshipKeys)) {
                     continue;
                 }
-                // Skip the nested documents as they will be checked later in recursions.
+
                 $oldAttributeValue = $old->getAttribute($key);
-                if ($oldAttributeValue instanceof Document) {
-                    continue;
-                }
                 // If values are not equal we need to update document.
                 if ($oldAttributeValue !== $value) {
                     $shouldUpdate = true;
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -2881,20 +2881,33 @@ public function updateDocument(string $collection, string $id, Document $documen
...
             foreach ($document as $key=>$value) {
+                // Skip the nested documents as they will be checked later in recursions.
+                if (array_key_exists($key, $relationshipKeys)) {
+                    continue;
+                }
                 // Skip the nested documents as they will be checked later in recursions.
                 $oldAttributeValue = $old->getAttribute($key);
                 if ($oldAttributeValue instanceof Document) {
                     continue;
                 }
```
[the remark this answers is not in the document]
fanatic75: Done this.

src/Database/Database.php
```diff
@@ -2881,20 +2881,30 @@ public function updateDocument(string $collection, string $id, Document $documen
...
         $validator = new Authorization(self::PERMISSION_UPDATE);
         $shouldUpdate = false;
 
         if ($collection->getId() !== self::METADATA) {
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
 
+            $relationshipKeys = [];
+            foreach ($relationships as $relationship) {
+                $relationshipKey = $relationship->getAttribute('key');
+                $relationshipKeys[$relationshipKey] =   $relationshipKey;
```
abnegate, line 2897: Let's fix the spacing after the equals here

abnegate (approved):

src/Database/Database.php
fanatic75: fixed this

PR #302, opened 2023-08-03
Fallback to $id  when validating select queries

fanatic75:
Metadata tables for indexes and attributes does not have key property so select queries cannot be validated. We need to fallback on  when key property is not present in such cases. This commit fixes that. 
Here's a discussion with Jake on this 
https://github.com/appwrite/appwrite/pull/5885#discussion_r1282509448


lohanidamodar (approved):
Approving this based on @abnegate 's comment on
https://github.com/appwrite/appwrite/pull/5885#discussion_r1282509448

PR #304, opened 2023-08-09
add checking of nested docs id in comparison check when updating document

fanatic75:
This PR checks the nested docs id to the old doc nested docs. 
For example:

If document A is related to document B but we do not have permission to access it, we can still establish a relationship between document A and a different version of document B (let's call it B1). This is because we compare B to its previous version, and B1 to its previous version as well. If there are no changes in B1, then we do not need permission to relate it to document A.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2930,7 +2930,7 @@ class Database
                     $relationType = (string) $relationships[$key]['options']['relationType'];
                     $side = (string) $relationships[$key]['options']['side'];
                     switch($relationType) {
-                        case DATABASE::RELATION_ONE_TO_ONE: {
+                        case Database::RELATION_ONE_TO_ONE: {
                             $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
                             if (
                                 (\is_null($old->getAttribute($key)) && !\is_null($value)) ||
@@ -2952,8 +2952,8 @@ class Database
                         case Database::RELATION_ONE_TO_MANY:
                             {
                                 if (
-                                    ($side === DATABASE::RELATION_SIDE_PARENT && $relationType === Database::RELATION_MANY_TO_ONE) ||
-                                    ($side === DATABASE::RELATION_SIDE_CHILD && $relationType === Database::RELATION_ONE_TO_MANY)
+                                    ($side === Database::RELATION_SIDE_PARENT && $relationType === Database::RELATION_MANY_TO_ONE) ||
+                                    ($side === Database::RELATION_SIDE_CHILD && $relationType === Database::RELATION_ONE_TO_MANY)
                                 ) {
                                     if (
                                         (\is_null($old->getAttribute($key)) && !\is_null($value)) ||
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2920,82 +2920,82 @@ class Database
             $documentSecurity = $collection->getAttribute('documentSecurity', false);
 
             foreach ($relationships as $relationship) {
-                $relationshipKey = $relationship->getAttribute('key');
-                $relationships[$relationshipKey] = $relationship;
+                $relationships[$relationship->getAttribute('key')] = $relationship;
             }
+
             // Compare if the document has any changes
-            foreach ($document as $key=>$value) {
+            foreach ($document as $key => $value) {
                 // Skip the nested documents as they will be checked later in recursions.
-                if (array_key_exists($key, $relationships)) {
+                if (\array_key_exists($key, $relationships)) {
                     $relationType = (string) $relationships[$key]['options']['relationType'];
                     $side = (string) $relationships[$key]['options']['side'];
+
                     switch($relationType) {
-                        case Database::RELATION_ONE_TO_ONE: {
-                            $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
-                            if (
-                                (\is_null($oldValue) && !\is_null($value)) ||
-                                (!is_null($oldValue) && \is_null($value)) ||
-                                (\is_string($value) && $oldValue !== $value) ||
-                                ($value instanceof Document && $oldValue !== $value->getId())
-                            ) {
+                        case Database::RELATION_ONE_TO_ONE:
+                            $oldValue = $old->getAttribute($key) instanceof Document
+                                ? $old->getAttribute($key)->getId()
+                                : $old->getAttribute($key);
+
+                            if ((\is_null($value) !== \is_null($oldValue))
+                            || (\is_string($value) && $value !== $oldValue)
+                            || ($value instanceof Document && $value->getId() !== $oldValue)) {
                                 $shouldUpdate = true;
-                                break;
                             }
                             break;
-                        }
-                        case Database::RELATION_MANY_TO_ONE:
                         case Database::RELATION_ONE_TO_MANY:
-                        case Database::RELATION_MANY_TO_MANY: {
+                        case Database::RELATION_MANY_TO_ONE:
+                        case Database::RELATION_MANY_TO_MANY:
                             if (
                                 ($relationType === Database::RELATION_MANY_TO_ONE && $side === Database::RELATION_SIDE_PARENT) ||
                                 ($relationType === Database::RELATION_ONE_TO_MANY && $side === Database::RELATION_SIDE_CHILD)
                             ) {
-                                $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
-                                if (
-                                    (\is_null($oldValue) && !\is_null($value)) ||
-                                    (!is_null($oldValue) && \is_null($value)) ||
-                                    (\is_string($value) && $oldValue !== $value) ||
-                                    ($value instanceof Document &&  $oldValue !== $value->getId())
-                                ) {
+                                $oldValue = $old->getAttribute($key) instanceof Document
+                                    ? $old->getAttribute($key)->getId()
+                                    : $old->getAttribute($key);
+
+                                if ((\is_null($value) !== \is_null($oldValue))
+                                || (\is_string($value) && $value !== $oldValue)
+                                || ($value instanceof Document &&  $value->getId() !== $oldValue)) {
+                                    $shouldUpdate = true;
+                                }
+                                break;
+                            }
+
+                            if ((\is_null($old->getAttribute($key)) !== \is_null($value))
+                            || \count($old->getAttribute($key)) !== \count($value)) {
+                                $shouldUpdate = true;
+                                break;
+                            }
+                            foreach ($value as $index => $relation) {
+                                $oldValue = $old->getAttribute($key)[$index] instanceof Document
+                                    ? $old->getAttribute($key)[$index]->getId()
+                                    : $old->getAttribute($key)[$index];
+
+                                if ((\is_string($relation) && $relation !== $oldValue)
+                                || ($relation instanceof Document && $relation->getId() !== $oldValue)) {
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -2919,15 +2919,113 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                        case Database::RELATION_MANY_TO_ONE:
+                        case Database::RELATION_ONE_TO_MANY:
+                            {
+                                if (
+                                    ($side === Database::RELATION_SIDE_PARENT && $relationType === Database::RELATION_MANY_TO_ONE) ||
+                                    ($side === Database::RELATION_SIDE_CHILD && $relationType === Database::RELATION_ONE_TO_MANY)
+                                ) {
+                                    if (
+                                        (\is_null($old->getAttribute($key)) && !\is_null($value)) ||
+                                        (!is_null($old->getAttribute($key)) && \is_null($value))) {
```
abnegate, line 2960: Let's be consistent with global space prefix usage (and we should use it, it's faster and easier to understand), this check can be simplified too based on other comments

src/Database/Database.php
```diff
@@ -2919,15 +2919,113 @@ public function updateDocument(string $collection, string $id, Document $documen
...
             }
             // Compare if the document has any changes
             foreach ($document as $key=>$value) {
                 // Skip the nested documents as they will be checked later in recursions.
-                if (array_key_exists($key, $relationshipKeys)) {
+                if (array_key_exists($key, $relationships)) {
+                    $relationType = (string) $relationships[$key]['options']['relationType'];
+                    $side = (string) $relationships[$key]['options']['side'];
+                    switch($relationType) {
+                        case Database::RELATION_ONE_TO_ONE: {
```
abnegate, line 2933: We don't need the curly braces around the switch cases

src/Database/Database.php
```diff
@@ -2919,15 +2919,113 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                                break;
+                            } elseif (\is_string($value)) {
+                                if ($oldValue !== $value) {
+                                    $shouldUpdate = true;
+                                    break;
+                                }
+                            } elseif ($oldValue !== $value->getId()) {
+                                $shouldUpdate = true;
+                                break;
+                            }
```
abnegate, line 2948: We can simplify this:

```suggestion
                            if (
                                (\is_null($value) !== \is_null($oldValue)) ||
                                (\is_string($value) && $oldValue !== $value) ||
                                ($oldValue !== $value->getId())
                            ) {
                                $shouldUpdate = true;
                                break;
                            }
```

src/Database/Database.php
```diff
@@ -2919,15 +2919,113 @@ public function updateDocument(string $collection, string $id, Document $documen
...
             // Compare if the document has any changes
             foreach ($document as $key=>$value) {
                 // Skip the nested documents as they will be checked later in recursions.
-                if (array_key_exists($key, $relationshipKeys)) {
+                if (array_key_exists($key, $relationships)) {
+                    $relationType = (string) $relationships[$key]['options']['relationType'];
+                    $side = (string) $relationships[$key]['options']['side'];
+                    switch($relationType) {
+                        case Database::RELATION_ONE_TO_ONE: {
+                            $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
```
abnegate, line 2934: `$old->getAttribute($key)` will always be a `?Document` here, since it was fetched with `getDocument`, which will have populated the relationships

src/Database/Database.php
```diff
@@ -2919,15 +2919,113 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                                break;
+                            }
+                            break;
+                        }
+                        case Database::RELATION_MANY_TO_ONE:
+                        case Database::RELATION_ONE_TO_MANY:
+                            {
+                                if (
+                                    ($side === Database::RELATION_SIDE_PARENT && $relationType === Database::RELATION_MANY_TO_ONE) ||
+                                    ($side === Database::RELATION_SIDE_CHILD && $relationType === Database::RELATION_ONE_TO_MANY)
```
abnegate, line 2956: Let's check `$relationType` first then `$side` to be consistent with other checks

abnegate (changes requested):
Let's see if we can simplify some of the conditions to make it a bit easier to read

src/Database/Database.php
fanatic75: done this

src/Database/Database.php
fanatic75: can we keep them? It allows me to minimize the whole case at once without curly brackets, it's not possible.

src/Database/Database.php
fanatic75: no in nested call of updateDocument, 
I found that $old->getAttribute($key) was not always a document. It was just a related document id as well.

src/Database/Database.php
fanatic75: I simplified the if conditions. just check once it's okay or not

src/Database/Database.php
fanatic75: updated the if conditions

src/Database/Database.php
abnegate: Let's remove them, they will be inconsistent with other switches, and we shouldn't change code style just for IDE features. 

If you're using VS Code, you could change the `folding strategy` to allow minimizing the whole case without changing code

src/Database/Database.php
abnegate: Ah okay that makes sense since the nested call use `skipRelationships` :+1:

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
-                if (array_key_exists($key, $relationshipKeys)) {
+                if (array_key_exists($key, $relationships)) {
+                    $relationType = (string) $relationships[$key]['options']['relationType'];
+                    $side = (string) $relationships[$key]['options']['side'];
+                    switch($relationType) {
+                        case Database::RELATION_ONE_TO_ONE: {
+                            $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
+                            if (
+                                (\is_null($oldValue) && !\is_null($value)) ||
+                                (!is_null($oldValue) && \is_null($value)) ||
```
abnegate, line 2937: These two can just be `(\is_null($value) !== \is_null($oldValue)) `

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                                if (
+                                    (\is_null($oldValue) && !\is_null($value)) ||
+                                    (!is_null($oldValue) && \is_null($value)) ||
+                                    (\is_string($value) && $oldValue !== $value) ||
+                                    ($value instanceof Document &&  $oldValue !== $value->getId())
+                                ) {
+                                    $shouldUpdate = true;
+                                    break;
+                                }
+                            } else {
```
abnegate, line 2963: Let's break from the if instead of nesting in an else

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
             // Compare if the document has any changes
             foreach ($document as $key=>$value) {
                 // Skip the nested documents as they will be checked later in recursions.
-                if (array_key_exists($key, $relationshipKeys)) {
+                if (array_key_exists($key, $relationships)) {
+                    $relationType = (string) $relationships[$key]['options']['relationType'];
+                    $side = (string) $relationships[$key]['options']['side'];
+                    switch($relationType) {
+                        case Database::RELATION_ONE_TO_ONE: {
+                            $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
```
abnegate, line 2934: Let's line break to make ternarys easier to read:

```suggestion
                            $oldValue = $old->getAttribute($key) instanceof Document 
                                ? $old->getAttribute($key)->getId()
                                : $old->getAttribute($key);
```

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                                if (
+                                    (\is_null($old->getAttribute($key)) && !\is_null($value)) ||
+                                    (!is_null($old->getAttribute($key)) && \is_null($value)) ||
+                                    \count($old->getAttribute($key)) !== \count($value)
+                                ) {
+                                    $shouldUpdate = true;
+                                    break;
+                                }
+                                foreach ($value as $index=>$relation) {
+                                    $oldValue = $old->getAttribute($key)[$index] instanceof Document ? $old->getAttribute($key)[$index]->getId() : $old->getAttribute($key)[$index];
```
abnegate, line 2973: Let's line break this too

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                                    if (
+                                        (\is_string($relation) && $oldValue!==$relation) ||
```
abnegate, line 2975:
```suggestion
                                        (\is_string($relation) && $oldValue !== $relation) ||
```

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                        case Database::RELATION_ONE_TO_MANY:
+                        case Database::RELATION_MANY_TO_MANY: {
+                            if (
+                                ($relationType === Database::RELATION_MANY_TO_ONE && $side === Database::RELATION_SIDE_PARENT) ||
+                                ($relationType === Database::RELATION_ONE_TO_MANY && $side === Database::RELATION_SIDE_CHILD)
+                            ) {
+                                $oldValue = $old->getAttribute($key) instanceof Document ? $old->getAttribute($key)->getId() : $old->getAttribute($key);
+                                if (
+                                    (\is_null($oldValue) && !\is_null($value)) ||
+                                    (!is_null($oldValue) && \is_null($value)) ||
```
abnegate, line 2956: These two can just be `(\is_null($value) !== \is_null($oldValue))`

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                                    (\is_string($value) && $oldValue !== $value) ||
+                                    ($value instanceof Document &&  $oldValue !== $value->getId())
+                                ) {
+                                    $shouldUpdate = true;
+                                    break;
+                                }
+                            } else {
+                                if (
+                                    (\is_null($old->getAttribute($key)) && !\is_null($value)) ||
+                                    (!is_null($old->getAttribute($key)) && \is_null($value)) ||
```
abnegate, line 2966: These two can just be `(\is_null($value) !== \is_null($oldValue))`

src/Database/Database.php
```diff
@@ -2919,15 +2919,73 @@ public function updateDocument(string $collection, string $id, Document $documen
...
             foreach ($relationships as $relationship) {
                 $relationshipKey = $relationship->getAttribute('key');
-                $relationshipKeys[$relationshipKey] = $relationshipKey;
+                $relationships[$relationshipKey] = $relationship;
             }
             // Compare if the document has any changes
             foreach ($document as $key=>$value) {
                 // Skip the nested documents as they will be checked later in recursions.
-                if (array_key_exists($key, $relationshipKeys)) {
+                if (array_key_exists($key, $relationships)) {
```
abnegate, line 2929:
```suggestion
                if (\array_key_exists($key, $relationships)) {
```

abnegate (commented):
Nice looks much better, just a few more style changes :+1:

abnegate (approved):

PR #306, opened 2023-08-11
fixes edge case in comparison check

fanatic75:
This PR fixes the edge case of comparing nested doc ID only if the nested level is less than MAX DEPTH allowed.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2940,6 +2940,10 @@ class Database
             foreach ($document as $key => $value) {
                 // Skip the nested documents as they will be checked later in recursions.
                 if (\array_key_exists($key, $relationships)) {
+                    // No need to compare nested documents more than max depth.
+                    if (count($this->relationshipWriteStack) >= Database::RELATION_MAX_DEPTH - 1) {
+                        continue;
+                    }
                     $relationType = (string) $relationships[$key]['options']['relationType'];
                     $side = (string) $relationships[$key]['options']['side'];
 
--- a/tests/Database/Base.php
+++ b/tests/Database/Base.php
@@ -3369,7 +3369,7 @@ abstract class Base extends TestCase
                 'level3' => [
                     '$id' => 'level3',
                     '$permissions' => [],
-                    'name' => 'Level haha',
+                    'name' => 'Level 3',
                     'level4' => [
                         '$id' => 'level4',
                         '$permissions' => [],
--- a/tests/Database/Base.php
+++ b/tests/Database/Base.php
@@ -3337,13 +3337,13 @@ abstract class Base extends TestCase
             'filters' => [],
         ]);
 
-        $permission = [
+        $permissions = [
             Permission::read(Role::any()),
             Permission::create(Role::any()),
             Permission::delete(Role::any()),
         ];
         for ($i=1; $i < 6; $i++) {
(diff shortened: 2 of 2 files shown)


src/Database/Database.php
```diff
@@ -2940,6 +2940,10 @@ public function updateDocument(string $collection, string $id, Document $documen
             foreach ($document as $key => $value) {
                 // Skip the nested documents as they will be checked later in recursions.
                 if (\array_key_exists($key, $relationships)) {
+                    // No need to compare nested documents more than max depth.
+                    if (count($this->relationshipWriteStack) >= Database::RELATION_MAX_DEPTH - 1) {
```
abnegate, line 2944: Let's keep this simple:

```suggestion
                    if (\count($this->relationshipWriteStack) > Database::RELATION_MAX_DEPTH) {
```

abnegate: Ignore this for now, let's add a test for updating a 3rd level document and ensure the changes are made

tests/Database/Base.php
```diff
@@ -3326,49 +3326,87 @@ public function testNoChangeUpdateDocumentWithRelationWithoutPermission(): void
...
-        $parent->setAttribute('children', ['child1']);
+        $updateLevel1 = static::getDatabase()->updateDocument('level1', 'level1', $level1);
 
-        $updatedParent = static::getDatabase()->updateDocument('parentRelationTest', 'parent1', $parent);
+        $this->assertEquals($level1, $updateLevel1);
 
-        $this->assertEquals($updatedParent->getUpdatedAt(), $parent->getUpdatedAt());
-        $this->assertEquals($updatedParent->getAttribute('children')[0]->getUpdatedAt(), $parent->getAttribute('children')[0]->getUpdatedAt());
+        $this->expectException(AuthorizationException::class);
+        static::getDatabase()->updateDocument('level1', 'level1', $level1->setAttribute('name', 'haha'));
```
abnegate, line 3405: If you want to run code after this, you need to use a try/catch and instanceof on the exception, or fail the test if there was no exception thrown

tests/Database/Base.php
```diff
@@ -3326,49 +3326,87 @@ public function testNoChangeUpdateDocumentWithRelationWithoutPermission(): void
...
 
-        static::getDatabase()->deleteCollection('parentRelationTest');
```
abnegate, line 3370: Let's assert that we can successfully update a third level document

fanatic75: Added assertion

tests/Database/Base.php
fanatic75: fixed this

tests/Database/Base.php
```diff
@@ -3326,49 +3326,93 @@ public function testNoChangeUpdateDocumentWithRelationWithoutPermission(): void
...
+        } catch(Exception $e) {
+            $this->assertInstanceOf(AuthorizationException::class, $e);
+        }
+        $level1->setAttribute('name', 'Level 1');
+        static::getDatabase()->updateCollection('level3', [
+            Permission::read(Role::any()),
+            Permission::create(Role::any()),
+            Permission::update(Role::any()),
+            Permission::delete(Role::any()),
+        ], false);
```
abnegate, line 3402: You shouldn't be able to do this without update permission unless using `Auth::skip` 🤔 I'll address in a separate PR

abnegate (approved):

PR #309, opened 2023-08-17
Fix update calls not returning relationships that weren't updated

abnegate:
Make sure not to unset relationships values from the document when values are the same when updating, so that the document returned to the user has the relationships


fanatic75:
@abnegate , The tests are failing because of this change. Can you check if this is due to returning the document since the collection is _metadata or due to not removing the relationship key?

stnguyen90 (changes requested):
merge conflict

PR #310, opened 2023-08-21
Fix throwing global exception instead of database exception

abnegate:
- Fix throwing global exception instead of database exception in some cases
- Add query exception so that query specific exceptions can be caught since validation happens in the library now

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -2180,7 +2180,7 @@ class Database
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
         if ($collection->isEmpty()) {
-            throw new DatabaseException("Collection not found");
+            throw new DatabaseException('Collection not found');
         }
 
         $attributes = $collection->getAttribute('attributes', []);
@@ -4242,7 +4242,6 @@ class Database
 
         $queries = Query::groupByType($queries)['filters'];
         $queries = self::convertQueries($collection, $queries);
-
 
         $getCount = fn () => $this->adapter->count($collection->getId(), $queries, $max);
         $count = $skipAuth ?? false ? Authorization::skip($getCount) : $getCount();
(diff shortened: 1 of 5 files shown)


stnguyen90 (approved):

src/Database/Database.php
```diff
@@ -2177,6 +2179,17 @@ public function getDocument(string $collection, string $id, array $queries = [])
 
         $collection = $this->silent(fn () => $this->getCollection($collection));
 
+        if ($collection->isEmpty()) {
+            throw new DatabaseException("Collection not found");
```
abnegate, line 2183:
```suggestion
            throw new DatabaseException('Collection not found');
```

src/Database/Database.php
```diff
@@ -4327,6 +4243,7 @@ public function count(string $collection, array $queries = [], ?int $max = null)
         $queries = Query::groupByType($queries)['filters'];
         $queries = self::convertQueries($collection, $queries);
 
+
```
abnegate, line 4246:
```suggestion
```

src/Database/Query.php
```diff
@@ -200,7 +200,7 @@ public static function parse(string $filter): self
         $paramsStart = mb_strpos($filter, static::CHAR_PARENTHESES_START);
 
         if ($paramsStart === false) {
-            throw new DatabaseException("Invalid query");
+            throw new QueryException("Invalid query");
```
abnegate, line 203:
```suggestion
            throw new QueryException('Invalid query');
```

src/Database/Query.php
```diff
@@ -211,7 +211,7 @@ public static function parse(string $filter): self
 
         // Check for deprecated query syntax
         if (\str_contains($method, '.')) {
-            throw new DatabaseException("Invalid query method");
+            throw new QueryException("Invalid query method");
```
abnegate, line 214:
```suggestion
            throw new QueryException('Invalid query method');
```

PR #316, opened 2023-08-27
Skip structure validation for now in createDocuments

lohanidamodar:
Also includes changes to use internalID on createDocuments

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1407,10 +1407,11 @@ class Database
 
             $document = $this->encode($collection, $document);
 
-            $validator = new Structure($collection);
-            if (!$validator->isValid($document)) {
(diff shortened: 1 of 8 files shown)


src/Database/Database.php
```diff
@@ -236,11 +238,12 @@ class Database
      * @param Adapter $adapter
      * @param Cache $cache
      */
-    public function __construct(Adapter $adapter, Cache $cache, array $filters = [])
+    public function __construct(Adapter $adapter, Cache $cache, array $filters = [], String $name = 'default')
```
eldadfux, line 241: String should be lowercase. It would make more sense to have nam as the first param later.

eldadfux (changes requested):

PR #319, opened 2023-09-01
bump mongo version

fanatic75:


composer.json
```diff
@@ -35,13 +35,13 @@
...
         "utopia-php/cache": "0.8.*",
-        "utopia-php/mongo": "0.2.*"
+        "utopia-php/mongo": "0.3.*"
     },
     "require-dev": {
         "fakerphp/faker": "^1.14",
         "phpunit/phpunit": "^9.4",
         "pcov/clobber": "^2.0",
-        "mongodb/mongodb": "1.8.0",
+        "mongodb/mongodb": "1.10.0",
```
abnegate, line 44: I think we can just remove this since it's a transitive dependency of utopia mongo

fanatic75: done this

abnegate (approved):

PR #320, opened 2023-09-04
fixes missing keys when updating doc

fanatic75:
This PR fixes the missing keys when updating doc via relationship. 

We currently delete related keys if the old value is the same as the new value for nested documents. However, this creates a problem when populating relationships. We need the key to have the ID of the related document so that we can fetch it and populate it once the document update process is complete.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3111,6 +3111,14 @@ class Database
             $side = (string) $relationship['options']['side'];
 
             if ($oldValue == $value) {
+                if (
+                    ($relationType === Database::RELATION_ONE_TO_ONE  || 
+                    ($relationType === DATABASE::RELATION_MANY_TO_ONE && $side === DATABASE::RELATION_SIDE_PARENT)) && 
+                    $value instanceof Document
+                ) {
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -3111,6 +3111,14 @@ private function updateDocumentRelationships(Document $collection, Document $old
             $side = (string) $relationship['options']['side'];
 
             if ($oldValue == $value) {
+                if (
+                    ($relationType === Database::RELATION_ONE_TO_ONE  ||
+                    ($relationType === Database::RELATION_MANY_TO_ONE && $side === Database::RELATION_SIDE_PARENT)) &&
```
stnguyen90, line 3116: do we need to consider `$relationType === Database::RELATION_ONE_TO_MANY && $side === Database::RELATION_SIDE_CHILD`?

stnguyen90: From Prateek:

> No. It's not a problem.  My test case covers this scenario. I'm updating parent via child in one to many relationship.

stnguyen90 (approved):

abnegate (approved):

PR #322, opened 2023-09-13
Cursors

fogelito:
Fixing Cursor length limitation for migrations

https://github.com/appwrite/appwrite/issues/6196

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -839,16 +839,16 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "1.10.32",
+            "version": "1.10.33",
             "source": {
                 "type": "git",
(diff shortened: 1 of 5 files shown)


stnguyen90 (changes requested):
It would be best to remove the composer.lock change if no version bump was required in this PR.

Otherwise, looks good.

fogelito:
@abnegate Are we going with a solution for Cursor Validation? or increasing the length for UID validation?

abnegate:
> @abnegate Are we going with a solution for Cursor Validation? or increasing the length for UID validation?

@fogelito We'll go with a different approach and allow enabling/disabling validation per instance, this way we can disable it during migrations only

PR #323, opened 2023-09-26
Non alpha sort

fogelito:


abnegate (approved):

src/Database/Adapter/MariaDB.php
```diff
@@ -31,7 +31,7 @@ public function create(string $name): bool
             return true;
         }
 
-        $sql = "CREATE DATABASE `{$name}` /*!40100 DEFAULT CHARACTER SET utf8mb4 */;";
+        $sql = "CREATE DATABASE IF NOT EXISTS `{$name}` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;";
```
fogelito, line 34: I think  COLLATE is different between Maria & MySQL

fogelito:
Do we want to support accents / case sensitive?

PR #327, opened 2023-10-01
Same twoWayKey

fogelito:


tests/Database/Base.php
```diff
@@ -3429,6 +3429,47 @@ public function testNoChangeUpdateDocumentWithoutPermission(Document $document):
         return $document;
     }
 
+    public function testRelationSameKey(): void
+    {
+    if (!static::getDatabase()->getAdapter()->getSupportForRelationships()) {
```
abnegate, line 3434:
```suggestion
        if (!static::getDatabase()->getAdapter()->getSupportForRelationships()) {
```

tests/Database/Base.php
```diff
@@ -3429,6 +3429,47 @@ public function testNoChangeUpdateDocumentWithoutPermission(Document $document):
...
+            return;
+        }
+
+        $permissions = [
+            Permission::read(Role::any()),
+            Permission::create(Role::any()),
+            Permission::delete(Role::any()),
+        ];
+
+        static::getDatabase()->createCollection("c1", [], [], $permissions);
```
abnegate, line 3445: Let's use single quotes throughout the test

src/Database/Database.php
```diff
@@ -1528,7 +1528,7 @@ public function createRelationship(
 
         $id ??= $relatedCollection->getId();
 
-        $twoWayKey ??= $collection->getId();
+        $twoWayKey ??= $collection->getId() . '_' . uniqid();
```
abnegate, line 1531: Maybe we could do this only after we get a duplicate exception? Then if there's no duplicate we still get a nice key

tests/Database/Base.php
fogelito: Done!

src/Database/Database.php
```diff
@@ -1537,11 +1537,15 @@ public function createRelationship(
...
+            try {
+                if ($attribute->getAttribute('type') === self::VAR_RELATIONSHIP
+                    && \strtolower($attribute->getAttribute('options')['twoWayKey']) === \strtolower($twoWayKey)
+                    && $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
+                ) {
+                    throw new DuplicateException('Related attribute already exists');
+                }
+            } catch (DuplicateException $e) {
+                $twoWayKey ??= $collection->getId() . '_' . uniqid();
             }
```
abnegate, line 1549: I don't think we need a try/catch here, we can use use the if:

```suggestion
            if ($attribute->getAttribute('type') === self::VAR_RELATIONSHIP
                && \strtolower($attribute->getAttribute('options')['twoWayKey']) === \strtolower($twoWayKey)
                && $attribute->getAttribute('options')['relatedCollection'] === $relatedCollection->getId()
            ) {
                $twoWayKey ??= $collection->getId() . '_' . uniqid();
            }
```

fogelito:
We tool care of this on Appwrite side

PR #329, opened 2023-10-03
Check database exist before create

fogelito:


tests/Database/Adapter/MariaDBTest.php
```diff
@@ -49,11 +49,6 @@ public static function getDatabase(): Database
         $database = new Database(new MariaDB($pdo), $cache);
         $database->setDefaultDatabase('utopiaTests');
         $database->setNamespace('myapp_'.uniqid());
-
-        if ($database->exists('utopiaTests')) {
```
fogelito, line 53: I remove this,  since it is Done in Base.php Tests

abnegate:
Do we need this change? Shouldn't it be okay as is with `CREATE DATABASE IF NOT EXISTS`?

fogelito:
We had some lock incidents with the create schema, it is triggering a DDL operation,
https://stackoverflow.com/questions/6169263/mysqls-create-database-if-not-exists-misbehaves-when-database-has-live-connec
@stnguyen90 What do you think about it? you saw this live

stnguyen90 (approved):

abnegate:
> We had some lock incidents with the create schema, it is triggering a DDL operation, https://stackoverflow.com/questions/6169263/mysqls-create-database-if-not-exists-misbehaves-when-database-has-live-connec @stnguyen90 What do you think about it? you saw this live

@fogelito I don't think this query was the one that held the lock, but the one that was blocked when trying to acquire it

stnguyen90:
> @fogelito I don't think this query was the one that held the lock, but the one that was blocked when trying to acquire it

@abnegate I think this can make the problem worse because once this query starts, a full database lock is created and can block other queries after.

tests/Database/Adapter/MariaDBTest.php
abnegate: Can you link it? I don't see a call to delete in Base.php

fogelito: https://github.com/utopia-php/database/blob/107a6332c01da5c08bc2b31a158cf08acc0ac792/tests/Database/Base.php#L72

I guess I can revert this to be more consistent for a fresh start, not to depend on this exist method 👍

eldadfux (approved):

src/Database/Adapter/MariaDB.php
```diff
@@ -27,6 +27,10 @@ public function create(string $name): bool
     {
         $name = $this->filter($name);
 
+        if ($this->exists($name)) {
+            return true;
+        }
+
         $sql = "CREATE DATABASE IF NOT EXISTS `{$name}` /*!40100 DEFAULT CHARACTER SET utf8mb4 */;";
```
eldadfux, line 34:
```suggestion
        $sql = "CREATE DATABASE `{$name}` /*!40100 DEFAULT CHARACTER SET utf8mb4 */;";
```

With the new check we don't really need this.

eldadfux: Also, do we have the same in other adapters?

abnegate (approved):

PR #330, opened 2023-10-04
fetchAll getDocument

fogelito:
Cannot execute queries while other unbuffered queries are active. Consider using PDOStatement::fetchAll(). Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -1182,8 +1182,8 @@ class MariaDB extends SQL
         $stmt->execute();
 
         $result = $stmt->fetchall();
-        if (!empty($document)) {
-            $result = $document[0];
+        if (!empty($result)) {
+            $result = $result[0];
         }
 
         return $result['sum'] ?? 0;
@@ -1242,8 +1242,8 @@ class MariaDB extends SQL
         $stmt->execute();
 
         $result = $stmt->fetchall();
-        if (!empty($document)) {
-            $result = $document[0];
+        if (!empty($result)) {
+            $result = $result[0];
         }
 
         return $result['sum'] ?? 0;
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -1181,7 +1181,10 @@ class MariaDB extends SQL
 
         $stmt->execute();
 
-        $result = $stmt->fetch();
+        $result = $stmt->fetchall();
+        if (!empty($document)) {
+            $result = $document[0];
+        }
 
         return $result['sum'] ?? 0;
     }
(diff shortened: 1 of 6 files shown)


src/Database/Adapter/SQL.php
```diff
@@ -121,12 +121,14 @@ public function getDocument(string $collection, string $id, array $queries = [])
         $stmt->bindValue(':_uid', $id);
         $stmt->execute();
 
-        $document = $stmt->fetch();
+        $document = $stmt->fetchAll();
```
abnegate, line 124: Is there any other places we need to change to `fetchAll`?

fogelito: I added one more place 👍 
For the count (aggregation functions) I do not think we need to do so.

fogelito: The error is on line 122 in sentry , not 144 🤔

abnegate: > I added one more place 👍 For the count (aggregation functions) I do not think we need to do so.

Let's do it for all `fetch` calls, we can't tell which ones could cause concurrency issues so if we do it for all we can be sure it's not the issue 👌

fogelito: Done 👍

src/Database/Adapter/MariaDB.php
```diff
@@ -1181,7 +1183,11 @@ public function count(string $collection, array $queries = [], ?int $max = null,
 
         $stmt->execute();
 
-        $result = $stmt->fetch();
+        $result = $stmt->fetchall();
```
abnegate, line 1186: Let's make sure we use proper casing:

```suggestion
        $result = $stmt->fetchAll();
```

src/Database/Adapter/MariaDB.php
```diff
@@ -1238,7 +1244,11 @@ public function sum(string $collection, string $attribute, array $queries = [],
...
```
abnegate, line 1247: Let's make sure we use proper casing:

```suggestion
        $result = $stmt->fetchAll();
```

src/Database/Adapter/SQL.php
```diff
@@ -81,7 +81,11 @@ public function exists(string $database, ?string $collection): bool
 
         $stmt->execute();
 
-        $document = $stmt->fetch();
+        $document = $stmt->fetchall();
```
abnegate, line 84: Let's make sure we use proper casing:

```suggestion
        $result = $stmt->fetchAll();
```

src/Database/Adapter/SQLite.php
```diff
@@ -53,7 +53,11 @@ public function exists(string $database, ?string $collection): bool
 
         $stmt->execute();
 
-        $document = $stmt->fetch();
+        $document = $stmt->fetchall();
```
abnegate, line 56: Let's make sure we use proper casing:

```suggestion
        $result = $stmt->fetchAll();
```

abnegate (approved):

PR #335, opened 2023-10-09
Feat array contains

abnegate:
- Allow querying array contents using `Query::contains`


src/Database/Adapter/MariaDB.php
```diff
@@ -218,7 +218,7 @@ public function createAttribute(string $collection, string $id, string $type, in
         $type = $this->getSQLType($type, $size, $signed);
 
         if ($array) {
-            $type = 'LONGTEXT';
+            $type = 'JSON';
```
fogelito, line 221: We will have to migrate this on Appwrite

src/Database/Adapter/MariaDB.php
```diff
@@ -246,7 +246,7 @@ public function updateAttribute(string $collection, string $id, string $type, in
         $type = $this->getSQLType($type, $size, $signed);
 
         if ($array) {
```
fogelito, line 248: Should it be possible to switch between array attributes and nonarray attributes?

tests/Database/Base.php
```diff
@@ -1845,7 +1846,7 @@ public function testFindContains(): void
          * Array contains OR condition
          */
         $documents = static::getDatabase()->find('movies', [
-            Query::contains('generes', ['comics', 'kids']),
+            Query::contains('genres', ['comics', 'kids']),
```
fogelito, line 1849: Will Query::contains on nonarray attribute should throw an Exception?

src/Database/Adapter/MariaDB.php
abnegate: In future we can allow it, it won't work for now even in main. It will throw an error when you try change from `LONGTEXT` to `TEXT` without manually truncating the contents if they're too long

tests/Database/Base.php
abnegate: Not sure about this one 🤔 I think it should, the other option could be treating it as `LIKE` on a non-array attribute?

src/Database/Adapter/MariaDB.php
fogelito: OK
BTW On MariaDB when trying to create a Json Attribute it does this under the hood:
`genres` longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL CHECK (json_valid(`genres`))
While in Mysql it is really a JSON field

src/Database/Adapter/MariaDB.php
fogelito: OK
Just saw on MariaDB when trying to create a JSON attribute, It creates this:
genres **longtext** CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL CHECK (json_valid(genres))
While in Mysql it is really a JSON field

src/Database/Adapter/MariaDB.php
fogelito: ok!

tests/Database/Base.php
fogelito: I think we should add a test for :)

fanatic75 (approved):
LGTM!

tests/Database/Base.php
fogelito: I tested a query on 
$documents = static::getDatabase()->find('movies', [
                Query::contains('name', ['captain']),
            ]);
            No errors but no results

src/Database/Adapter/MariaDB.php
abnegate: Hmm so we won't need a migration for MariaDB but will for MySQL.. this is a new problem 😅

fogelito: Maybe we can save it as Text + create a key / value Collection for search?

tests/Database/Base.php
abnegate: Fixed :+1:

PR #337, opened 2023-10-10
Feat query metadata

abnegate:
- Allow transforming queries per event
- Add method to set metadata on the database which will be added to queries as comments

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -22,6 +22,13 @@ abstract class Adapter
      */
     protected array $debug = [];
 
+    /**
+     * @var array<string, array<callable>>
+     */
+    protected array $transformations = [
+        '*' => [],
+    ];
+
     protected static ?int $timeout = null;
 
     /**
@@ -133,6 +140,37 @@ abstract class Adapter
         }
 
         return $this->defaultDatabase;
+    }
+
+
+    /**
+     * Apply a transformation to a query before an event occurs
+     *
+     * @param string $event
+     * @param string $name
+     * @param ?callable $callback
+     * @return self
+     */
+    public function before(string $event, string $name = '', ?callable $callback = null): self
+    {
+        if (!isset($this->transformations[$event])) {
+            $this->transformations[$event] = [];
+        }
+        $this->transformations[$event][$name] = $callback;
+
+        return $this;
+    }
+
+    protected function trigger(string $event, mixed $query): mixed
+    {
+        foreach ($this->transformations[Database::EVENT_ALL] as $callback) {
+            $query = $callback($query);
+        }
+        foreach (($this->transformations[$event] ?? []) as $callback) {
+            $query = $callback($query);
+        }
+
+        return $query;
     }
 
     /**
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -153,10 +153,7 @@ abstract class Adapter
 
         $output = '';
         foreach ($this->metadata as $key => $value) {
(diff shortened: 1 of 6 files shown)


src/Database/Adapter.php
```diff
@@ -135,6 +147,62 @@ public function getDefaultDatabase(): string
...
+    protected function trigger(string $event, mixed $query): mixed
+    {
+        foreach ($this->transformations[Database::EVENT_ALL] as $callback) {
+            $query = $callback($query);
+        }
+        foreach (($this->transformations[$event] ?? []) as $callback) {
+            $query = $callback($query);
+        }
+
+        return $query;
```
fogelito, line 203: This will return only the last $query, Should appended and send an array?

tests/Database/Base.php
```diff
@@ -12414,6 +12423,31 @@ public function testEmptyOperatorValues(): void
...
+                'required' => true,
+            ])
+        ]);
+
+        static::getDatabase()->createDocument('docs', new Document([
+            '$id' => 'doc1',
+            'name' => 'value1',
+        ]));
+
+        static::getDatabase()->before(Database::EVENT_DOCUMENT_READ, 'test', function (string $query) {
```
fogelito, line 12442: They are 2 places EVENT_DOCUMENT_READ in getDocument  Is this ok?

src/Database/Adapter.php
```diff
@@ -135,6 +147,62 @@ public function getDefaultDatabase(): string
...
     }
 
+    public function setMetadata(string $key, mixed $value): self
+    {
+        $this->metadata[$key] = $value;
+
+        $output = '';
+        foreach ($this->metadata as $key => $value) {
+            $output .= "-- {$key}: {$value}";
+            if (\array_key_last($this->metadata) !== $key) {
```
fogelito, line 157: In tests I see :
-- key: valueCREATE UNIQUE 
Without \n 
Not sure if this is expected

tests/Database/Base.php
```diff
@@ -12331,6 +12331,15 @@ public function testLabels(): void
         $this->assertCount(1, $documents);
     }
 
+    public function testMetadata(): void
+    {
+        static::getDatabase()->setMetadata('key', 'value');
+
+        static::getDatabase()->createCollection('testers');
+
+        $this->expectNotToPerformAssertions();
```
fogelito, line 12340: Should clearMetadata after test?

src/Database/Adapter.php
fogelito: var_dump $query I get:
-- key: valueCREATE INDEX 
Without \n Is this expected?

fogelito:
Tests fail :)

eldadfux (approved):

src/Database/Adapter.php
abnegate: > This will return only the last $query, Should appended and send an array?

Each callback will overwrite the original query with the modified one, so the changes are kept between each call

>var_dump $query I get:
-- key: valueCREATE INDEX
Without \n Is this expected?

This one is is not expected 🤔 will take a look

tests/Database/Base.php
abnegate: Yep one is for firing the event and the other is for firing the query transformation :+1:

src/Database/Adapter.php
abnegate: We were removing the newline from the last entry added, I think because some queries start with a new line. Now we always include it either way :+1:

PR #338, opened 2023-10-16
Feat: DB Proxy Adapter

Meldiron:
DB proxy is dedicated HTTP server that maintains all database collections.
Using this proxy instead of separating maximum connection limit between containers allows improvements to scalability. It also prevents many idling connections when they could be used better by another containers.


composer.json
```diff
@@ -52,5 +53,11 @@
...
         "ext-pdo": "Needed to support MariaDB, MySQL or SQLite Database Adapter",
         "mongodb/mongodb": "Needed to support MongoDB Database Adapter"
-    }
+    },
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/utopia-php/fetch.git"
+        }
+    ]
```
Meldiron, line 62: Remove once fetch is released

src/Database/Adapter/Proxy.php
```diff
@@ -0,0 +1,561 @@
...
+
+            try {
+                $exception = new $error['type']($error['message'], $error['code']);
+                /**
+                 * @var \Utopia\Database\Exception $exception
+                 */
+
+                $exception->setFile($error['file']);
+                $exception->setLine($error['line']);
+                // TODO: If possible in PHP, set trace too for better error reporting
```
Meldiron, line 71: Ask someone if this TODO is relevant. Seems like trace cannot be set in PHP

src/Database/Adapter/ProxyMariaDB.php
```diff
@@ -0,0 +1,471 @@
+<?php
+
+namespace Utopia\Database\Adapter;
+
+use Utopia\Database\Document;
+
+class ProxyMariaDB extends Proxy
```
abnegate, line 7: This sounds better to me as `MariaDBProxy`, what do you think?

src/Database/Exception.php
```diff
@@ -4,4 +4,12 @@
...
 class Exception extends \Exception
 {
+    public function setFile(string $file): void
+    {
+        $this->file = $file;
+    }
+    public function setLine(int $line): void
+    {
+        $this->line = $line;
+    }
```
abnegate, line 14: Can we set these in the constructor instead?

src/Database/Exception/Conflict.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Utopia\Database\Exception;
 
-class Conflict extends \Exception
+use Utopia\Database\Exception;
+
+class Conflict extends Exception
```
abnegate, line 7: Let's make sure we check if there's anywhere we need to update the order of catch blocks because of this change (if any sub-types are caught, those catches need to now come _before_ the `DatabaseException` catch)

tests/Database/Base.php
```diff
@@ -5088,7 +5088,7 @@ public function testOneToOneTwoWayRelationship(): void
             ],
         ]);
 
-        static::getDatabase()->createDocument('country', $doc);
+        static::getDatabase()->createDocument('country', new Document($doc->getArrayCopy()));
```
abnegate, line 5091: Why do we need this change?

src/Database/Adapter/ProxyMariaDB.php
eldadfux: I think we can skip this entire class and just use a generic proxy adapter to implement all of these methods and return the values from a "regular" adapter that is being passed in the constructor. It will save a lot of code duplication and maintenance.

src/Database/Adapter/Proxy.php
```diff
@@ -0,0 +1,561 @@
...
+
+        if ($response->getStatusCode() >= 400) {
+            if (empty($response->getBody())) {
+                throw new Exception('Internal ' . $response->getStatusCode() . ' HTTP error in database proxy');
+            }
+
+            $error = \json_decode($response->getBody(), true);
+
+            try {
+                $exception = new $error['type']($error['message'], $error['code']);
```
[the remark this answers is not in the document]
Meldiron: Yes. Im pretty confident about this try/catch, as we expect many different exception types in the tests

src/Database/Adapter/ProxyMariaDB.php
Meldiron: I remember there were some methods I couldn't get done over proxy.. I think some of the static ones. Also some didn't make sense to proxy, as they were just a simple number.
Thats why I made 1 adapter for 1 DB type - proxy all important methods, but simple static ones are kept on adapter

Meldiron: Also I renamed it to follow Data API naming convention

phpunit.xml
```diff
@@ -7,7 +7,7 @@
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
     processIsolation="false"
-    stopOnFailure="false">
+    stopOnFailure="true">
```
Meldiron, line 10:
```suggestion
    stopOnFailure="false">
```

src/Database/Query.php
```diff
@@ -2,10 +2,11 @@
 
 namespace Utopia\Database;
 
+use JsonSerializable;
 use JsonException;
 use Utopia\Database\Exception\Query as QueryException;
 
-class Query
+class Query implements JsonSerializable
```
Meldiron, line 9: Find better way. Like custom Serializable interface

src/Database/Adapter/ProxyMariaDB.php
Meldiron: Renamed to MariaDBProxy

src/Database/Adapter/Proxy.php
```diff
@@ -0,0 +1,746 @@
...
+    /**
+     * Create Database
+     *
+     * @param string $name
+     *
+     * @return bool
+     */
+    public function create(string $name): bool
+    {
+        return $this->query('create', [$name]);
```
eldadfux, line 141: Should we consider something like this for getting all arguments?
https://www.php.net/manual/en/function.func-get-args.php

Meldiron: We could, but I wanted to keep args in the definition so it's copy&paste of adapter abstract function.

Meldiron: Your call

eldadfux: We can consider it for later.

eldadfux (approved):

PR #339, opened 2023-10-17
Invalid Utf chars

fogelito:


phpunit.xml
```diff
@@ -7,7 +7,7 @@
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
     processIsolation="false"
-    stopOnFailure="false">
+    stopOnFailure="true">
```
abnegate, line 10: Let's revert this

tests/Database/Base.php
```diff
@@ -1082,6 +1081,43 @@ public function testCreateDocumentDefaults(): void
         static::getDatabase()->deleteCollection('defaults');
     }
 
+    public function testInvalidUtfCharacters(): void
```
abnegate, line 1084: Can you add some comments about what's being tested through the function?

fogelito: Done 👍 
I did not make any global changes in Utopia for this fix, but we should know there is a problem we should be aware when trying to insert non-utf chars , for next time this issue pops up we can suggest a fix.

abnegate: What's the exception in this case?

fogelito: I just wanted them to pass, to be aware of this issue, to alert it, and maybe later on add it in Utopia and not in the test

fogelito: Ohh :)
The exception message varies between adapters I update the message in the tests, but with mariadb I can not since it includes namespaces which change, so I added the status code. 
Good catch, did not realize Sqlite was passing, not sure what to do about it.

phpunit.xml
fogelito: Done 👍

PR #340, opened 2023-10-18
adds past date validation

fanatic75:


src/Database/Validator/Datetime.php
```diff
@@ -11,8 +11,16 @@ class Datetime extends Validator
      */
     protected string $message = 'Date is not valid';
 
-    public function __construct()
+    /**
+     * @var bool
+     */
+    protected bool $allowOldDates = true;
```
abnegate, line 17: Let's invert this to `$requireDateInFuture = false`, the term `old` doesn't make much sense here without context

src/Database/Validator/Datetime.php
```diff
@@ -37,7 +45,13 @@ public function isValid($value): bool
         }
 
         try {
-            new \DateTime($value);
+            $date = new \DateTime($value);
+            $now = new \DateTime();
+
+            if ($this->allowOldDates === false && $date < $now) {
+                $this->message = 'Date is in the past';
```
abnegate, line 52: Let's update this too:

```suggestion
                $this->message = 'Date must be in the future';
```

src/Database/Validator/Datetime.php
fanatic75: done this

src/Database/Validator/Datetime.php
fanatic75: done this

abnegate (approved):

PR #341, opened 2023-10-19
Global timeouts using before query hooks

abnegate:
## What's Changed

Allow using instance level query hooks to set timeout per database event (or all)

## Test Plan

Existing tests, updated timeout test

### DIFF
--- a/src/Database/Adapter/Mongo.php
+++ b/src/Database/Adapter/Mongo.php
@@ -42,6 +42,8 @@ class Mongo extends Adapter
 
     protected Client $client;
 
+    protected ?int $timeout = null;
+
     /**
      * Constructor.
      *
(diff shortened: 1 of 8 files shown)


tests/Database/Base.php
```diff
@@ -221,13 +221,14 @@ public function testCreatedAtUpdatedAt(): void
...
-    public function testQueryTimeoutUsingStaticTimeout(): void
+
+    public function testQueryTimeout(): void
     {
         if ($this->getDatabase()->getAdapter()->getSupportForTimeouts()) {
             static::getDatabase()->createCollection('global-timeouts');
             $this->assertEquals(true, static::getDatabase()->createAttribute('global-timeouts', 'longtext', Database::VAR_STRING, 100000000, true));
 
-            for ($i = 0 ; $i <= 5 ; $i++) {
+            for ($i = 0 ; $i <= 20 ; $i++) {
```
abnegate, line 231: Had to change this otherwise tests were failing, not sure if it's local to me, seems to be a performance win so not worried.

Context: loop iterations determine how many huge documents to insert to force a timeout when querying them

fogelito: We can make the text file bigger

fogelito (approved):

tests/Database/Base.php
fogelito: We can make the text in the file little longer, maybe will help to get a timeout

src/Database/Adapter/Mongo.php
```diff
@@ -1707,4 +1702,16 @@ public function getMaxIndexLength(): int
     {
         return 0;
     }
+
+    public function setTimeout(int $milliseconds, string $event = Database::EVENT_ALL): void
+    {
+        $this->timeout = $milliseconds;
```
christyjacob4, line 1708: dont we need to check $this->getSupportForTimeouts() ?

christyjacob4 (changes requested):
Tiny comment

christyjacob4 (approved):

PR #343, opened 2023-10-19
Feat: Make authorization per-instance

Meldiron:
Makes auth validator instance-aware, to make it work fine with corutines server

- [x] Update tests


src/Database/Database.php
```diff
@@ -306,13 +306,25 @@ class Database
...
+     */
+    private Authorization $authorization;
+
     /**
      * @param Adapter $adapter
      * @param Cache $cache
      * @param array<string, array{encode: callable, decode: callable}> $filters
      */
-    public function __construct(Adapter $adapter, Cache $cache, array $filters = [])
+    public function __construct(Adapter $adapter, Cache $cache, array $filters = [], Authorization $authorization = null)
```
abnegate, line 319: Can we have a `setAuthorization` function in this class as well instead of the constructor param?

src/Database/Database.php
```diff
@@ -2235,7 +2247,7 @@ public function getDocument(string $collection, string $id, array $queries = [])
 
         $queries = \array_values($queries);
 
-        $validator = new Authorization(self::PERMISSION_READ);
+        $validator = $this->authorization->setAction(self::PERMISSION_READ);
```
abnegate, line 2250: We should reset the message to the default in the `setAction` function to make sure no state carries over between validations

src/Database/Adapter/MariaDB.php
```diff
@@ -1144,15 +1143,15 @@ public function find(string $collection, array $queries = [], ?int $limit = 25,
...
+        $roles = $this->authorization->getRoles();
         $where = [];
         $limit = \is_null($max) ? '' : 'LIMIT :max';
 
         foreach ($queries as $query) {
             $where[] = $this->getSQLCondition($query);
         }
 
-        if (Authorization::$status) {
+        if ($this->authorization->status) {
```
fogelito, line 1154: Do we want to use a function getStatus and make Stutus private?

src/Database/Database.php
```diff
@@ -306,13 +306,25 @@ class Database
      */
     private array $relationshipDeleteStack = [];
 
+    /**
+     * @var Authorization
+     */
+    private Authorization $authorization;
```
fogelito, line 312: Can default be null?

src/Database/Database.php
Meldiron: We always need instance of validator, and it would be weird to create database, constructor makes new auth, and then I setAuth.

Ill keep this option (from constructor), and Ill add setAuth method.

src/Database/Database.php
Meldiron: Default is now new Authorization, as defined in constructor.

src/Database/Adapter/MariaDB.php
Meldiron: updated to getStatus

src/Database/Database.php
Meldiron: Implement using validate method which throws message, and I pass action directly into this method. That way there is no way it could bleed like this.

fogelito (approved):

src/Database/Validator/Authorization.php
```diff
@@ -43,109 +31,143 @@ public function getDescription(): string
...
+        $this->roles[$role] = true;
     }
 
     /**
      * @param string $role
      *
      * @return void
      */
-    public static function unsetRole(string $role): void
+    public function unsetRole(string $role): void
```
abnegate, line 94: While we're making changes here, let's change these to add/remove role to better reflect what they do. set/unset is for a singular item

src/Database/Database.php
```diff
@@ -2249,10 +2269,12 @@ public function getDocument(string $collection, string $id, array $queries = [])
...
-                if (!$validator->isValid([
-                    ...$collection->getRead(),
-                    ...($documentSecurity ? $document->getRead() : [])
-                ])) {
+                try {
+                    $this->authorization->validate(self::PERMISSION_READ, [
+                        ...$collection->getRead(),
+                        ...($documentSecurity ? $document->getRead() : [])
+                    ]);
+                } catch(Validation $err) {
```
abnegate, line 2277: Why do we need to try/catch? Let's just return false from the validator as before

src/Database/Validator/Authorization.php
```diff
@@ -43,109 +31,143 @@ public function getDescription(): string
...
+     * @throws Validation
      */
-    public function isValid($permissions): bool
+    public function validate(string $action, mixed $permissions): void
     {
-        if (!self::$status) {
-            return true;
+        if (!$this->status) {
+            return;
         }
```
abnegate, line 63: We shouldn't break the interface for just one validator, let's trying find a solution where we can still use the `isValid` method (which will also require less work for library consumers).

src/Database/Validator/Authorization.php
```diff
@@ -43,109 +31,143 @@ public function getDescription(): string
...
+    protected bool $status = true;
 
     /**
      * Default value in case we need
      *  to reset Authorization status
      *
      * @var bool
      */
-    public static bool $statusDefault = true;
+    protected bool $statusDefault = true;
```
abnegate, line 136: Let's move all properties to the start of the class

src/Database/Validator/Authorization.php
```diff
@@ -43,109 +31,143 @@ public function getDescription(): string
...
 
     /**
      * Change default status.
      * This will be used for the
-     *  value set on the self::reset() method
+     *  value set on the $this->reset() method
```
abnegate, line 141: We can remove the line break in the doc here, not sure why it's there

abnegate (commented):
Can't comment on the `Base.php` diff since it's so big 😅 but do we need both the static and instance `Authorization` instances there?

src/Database/Database.php
Meldiron: It was hard decision me and Torsten did.
Her is flow of auth validator:

1. isValid()
2. It sets $message
3. getDescription
4. It reads $message

Considering that Validator is designed to be scoped per-request (as you add auth permissions of a request), message should be too. Here are some ideas we had:

1. Have Validator and ValidatorResult
2. Add more params to isValid
3. Have Make isValid return error message directly
4. Make isValid throw exception

src/Database/Validator/Authorization.php
Meldiron: We can discuss, but I couldnt find better way. More info in above suggestion comment

Meldiron:
@abnegate 

> but do we need both the static and instance Authorization instances there?

`Authorization::` is no more. You always must `new Authroization()`. No backwards compatibility. Breaking change.

src/Database/Database.php
abnegate: Exceptions shouldn't be used as control flow; they're for when something exceptional occurs. In my opinion we should return arrays or objects (), and accept arrays or objects for the `isValid` method

abnegate:
> @abnegate
> 
> > but do we need both the static and instance Authorization instances there?
> 
> `Authorization::` is no more. You always must `new Authroization()`. No backwards compatibility. Breaking change.

I mean in the test base class we have an instance var `authorization` and a static var `authorizationSingleton`, I think we only need one or the other

Meldiron:
> authorizationSingleton

I cant improve this without renaming var in ~100 lines. The setup that is there still make sense - it's not Utopia-spepcfiic, but PHPUnit-specific.
static-var is there so state of it is shared across all tests.
instnace-var is there so we don't need to call static method in all tests.

src/Database/Database.php
Meldiron: No more try/catches

src/Database/Validator/Authorization.php
Meldiron: We are now back to normal isValid

tests/Database/Base.php
```diff
@@ -41,38 +51,44 @@ abstract protected static function getAdapterName(): string;
 
     public function setUp(): void
     {
-        Authorization::setRole('any');
+        if (is_null(self::$authorizationSingleton)) {
+            self::$authorizationSingleton = new Authorization();
+            self::$authorizationSingleton->addRole('any');
+        }
+
+        $this->authorization = self::$authorizationSingleton;
```
abnegate, line 59: Still not sure we need both :thinking: can't we just the instance level one?

src/Database/Validator/Authorization.php
```diff
@@ -43,109 +45,122 @@ public function getDescription(): string
...
-     *
-     * @return bool
-     */
-    public function isValid($permissions): bool
+    */
+    public function isValid(mixed $input): bool // any, CREATE
     {
-        if (!self::$status) {
+        if(!($input instanceof Input)) {
+            throw new Validation('Authorization validator requires Input class to validate.');
```
abnegate, line 56: Let's set the message `Invalid input provided` and return false here instead of throw, otherwise we'll need to add a catch block for all calls

abnegate (approved):

PR #349, opened 2023-10-27
Allow enabling/disabling validation

abnegate:


src/Database/Database.php
```diff
@@ -2691,9 +2728,11 @@ public function createDocument(string $collection, Document $document): Document
 
         $document = $this->encode($collection, $document);
 
-        $validator = new Permissions();
-        if (!$validator->isValid($document->getPermissions())) {
-            throw new InvalidArgumentException($validator->getDescription());
+        if ($this->validate) {
+            $validator = new Permissions();
```
fogelito, line 2732: Should Permissions be shut down as well?

fogelito (approved):

PR #351, opened 2023-11-02
Fix type error

abnegate:
- Force return an empty string when $id, $internalId, or $collection are explicitly set to null to avoid a type error

### DIFF
--- a/src/Database/Document.php
+++ b/src/Database/Document.php
@@ -55,7 +55,7 @@ class Document extends ArrayObject
      */
     public function getId(): string
     {
-        return $this->getAttribute('$id', '');
+        return $this->getAttribute('$id', '') ?? '';
     }
 
     /**
@@ -63,7 +63,7 @@ class Document extends ArrayObject
      */
     public function getInternalId(): string
     {
-        return $this->getAttribute('$internalId', '');
+        return $this->getAttribute('$internalId', '') ?? '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


eldadfux:
Isn't the getAttribute `$default` is designed to do exactly that?

```php
/**
     * Get Attribute.
     *
     * Method for getting a specific fields attribute. If $name is not found $default value will be returned.
     *
     * @param string $name
     * @param mixed $default
     *
     * @return mixed
     */
    public function getAttribute(string $name, mixed $default = null): mixed
    {
        if (isset($this[$name])) {
            return $this[$name];
        }

        return $default;
    }
 ```

abnegate:
> Isn't the getAttribute `$default` is designed to do exactly that?
> 
> ```
> /**
>      * Get Attribute.
>      *
>      * Method for getting a specific fields attribute. If $name is not found $default value will be returned.
>      *
>      * @param string $name
>      * @param mixed $default
>      *
>      * @return mixed
>      */
>     public function getAttribute(string $name, mixed $default = null): mixed
>     {
>         if (isset($this[$name])) {
>             return $this[$name];
>         }
> 
>         return $default;
>     }
> ```

Yeah this was a band-aid for the `getId()` on null error, somewhere the value for `$collection` must be explicitly set to `null` `$document->setAttribute('$collection', null);`, that's the only possible way you could get null here. Needs more investigation to find the root cause though

PR #354, opened 2023-11-05
 Many to One issue - invalid relation type 

fogelito:
Many to One issue - invalid relation type
From Sentry, with no specific issue, this is supposed to solve people getting 500 errors on update relations


stnguyen90:
@fogelito is there any related github issue for this?

fogelito:
Can't point to an issue since we fish it from Sentry

fogelito:
Closing this PR it looks like the input for unset a relation (object) should be null, not an empty array in this case

PR #358, opened 2023-11-17
Feat isolation mode

abnegate:
Allow setting different isolation modes by using different setters

Schema per project:
```php
$database
    ->setDatabase($projectId)
    ->create();
```

Single schema, unique tables (current arch):
```php
$database
    ->setDatabase('appwrite')
    ->setNamespace($projectId)
    ->create();
```

Single schema, shared tables:
```php
$database
    ->setDatabase('appwrite')
    ->setShareTables(true)
    ->setTenant($projectId)
    ->create();
```


src/Database/Adapter/MariaDB.php
```diff
@@ -116,18 +114,21 @@ public function createCollection(string $name, array $attributes = [], array $in
...
 
         $sql = "
-			CREATE TABLE IF NOT EXISTS `{$database}`.`{$namespace}_{$id}` (
-				`_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+			CREATE TABLE IF NOT EXISTS {$this->getSQLTable($id)} (
+				`_id` INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
 				`_uid` VARCHAR(255) NOT NULL,
-				`_createdAt` datetime(3) DEFAULT NULL,
-				`_updatedAt` datetime(3) DEFAULT NULL,
+				`_tenant` VARCHAR(36) DEFAULT NULL,
```
fogelito, line 120: If shareTables is false, do we want to create this attribute? can a shared table become a non-shared and the other way around? 
If we do not use _tenant , I am not sure we want to create

fogelito, line 120: If _tenant will be the project Internal id in Appwrite , can it be an integer? I know it is less generic but better for us

src/Database/Adapter/MariaDB.php
```diff
@@ -116,18 +114,21 @@ public function createCollection(string $name, array $attributes = [], array $in
...
+				`_createdAt` DATETIME(3) DEFAULT NULL,
+				`_updatedAt` DATETIME(3) DEFAULT NULL,
 				`_permissions` MEDIUMTEXT DEFAULT NULL,
 				" . \implode(' ', $attributeStrings) . "
 				PRIMARY KEY (`_id`),
 				" . \implode(' ', $indexStrings) . "
 				UNIQUE KEY `_uid` (`_uid`),
-				KEY `_created_at` (`_createdAt`),
-				KEY `_updated_at` (`_updatedAt`)
+				KEY `_tenant` (`_tenant`),
```
fogelito, line 128: Same question again about shareTables , if shareTables = true, it will always be added to the filter,
So could be we need to create the indexes with an if case.
For example if shareTables = false , We do not need this index:
KEY `_tenant` (`_tenant`)
and we will not create compound indexes as 
KEY `_created_at` (`_tenant`,`_createdAt`)

src/Database/Database.php
```diff
@@ -2126,6 +2284,10 @@ public function renameIndex(string $collection, string $old, string $new): bool
      */
     public function createIndex(string $collection, string $id, string $type, array $attributes, array $lengths = [], array $orders = []): bool
     {
+        if ($this->adapter->getShareTables() && empty($this->adapter->getTenant())) {
+            throw new DatabaseException('Missing tenant. Tenant must be set when table sharing is enabled.');
+        }
+
```
fogelito, line 2290: If sharedTables is True , Do we want the add Tenant as first part of the index?

src/Database/Database.php
```diff
@@ -643,6 +644,36 @@ public function clearTimeout(string $event = Database::EVENT_ALL): void
...
 
+    /**
+     * Set Share Tables
+     *
+     * Set whether to share tables between tenants
+     *
+     * @param bool $share
+     * @return self
+     */
+    public function setShareTables(bool $share): self
```
fogelito, line 655: In case of $share = true, should we force to $tenant as a second function param?
So we do not need to check in functions this:
 if ($this->adapter->getShareTables() && empty($this->adapter->getTenant()))

src/Database/Database.php
```diff
@@ -643,6 +644,36 @@ public function clearTimeout(string $event = Database::EVENT_ALL): void
...
+     * Set Tenant
+     *
+     * Set tenant to use if tables are shared
+     *
+     * @param string $tenant
+     * @return self
+     */
+    public function setTenant(string $tenant): self
+    {
+        $this->adapter->setTenant($tenant);
```
fogelito, line 672: Can we set setTenant while shared if false?

src/Database/Adapter/MariaDB.php
abnegate: It could be anything, I have been using $project->getId(), but could certainly use internal ID since it's not exposed on the Appwrite side, that will much better for query performance, will update :+1:

abnegate: We are just adding it to all tables to keep things simple, shouldn't be much impact if it's just unused right?

src/Database/Database.php
abnegate: I think it's better to keep them separate, then from the Appwrite side, we only need to call setShareTables once when instantiating the DB, and then for every request, we call setTenant with the project ID

src/Database/Database.php
abnegate: Yep, it will just not do anything

src/Database/Database.php
```diff
@@ -827,20 +883,30 @@ public function createCollection(string $id, array $attributes = [], array $inde
...
+            throw new DatabaseException($validator->getDescription());
         }
 
         $collection = $this->silent(fn () => $this->getCollection($id));
 
+        if ($collection->isEmpty()) {
+            throw new DatabaseException('Collection not found');
+        }
+
+        if ($this->adapter->getShareTables() && $collection->getAttribute('$tenant') !== $this->adapter->getTenant()) {
```
eldadfux, line 906: This could be another indication why having only tenant can be sufficient.

src/Database/Database.php
```diff
@@ -3782,7 +3992,7 @@ public function decreaseDocumentAttribute(string $collection, string $id, string
 
         $min = $min ? $min + $value : null;
         $result = $this->adapter->increaseDocumentAttribute($collection->getId(), $id, $attribute, $value * -1, $min);
-        $this->cache->purge('cache-' . $this->getNamespace() . ':' . $collection->getId() . ':' . $id . ':*');
+        $this->deleteCachedDocument($collection->getId(), $id);
```
eldadfux, line 3995: Calling this purgeDocument might help us have better separation and avoid confusion with the delete methods.

src/Database/Adapter/SQL.php
```diff
@@ -98,16 +107,26 @@ public function list(): array
     public function getDocument(string $collection, string $id, array $queries = []): Document
     {
         $name = $this->filter($collection);
-
         $selections = $this->getAttributeSelections($queries);
 
-        $stmt = $this->getPDO()->prepare("
-            SELECT {$this->getAttributeProjection($selections)} 
+        $sql = "
+			SELECT {$this->getAttributeProjection($selections)}
```
eldadfux, line 113: Indentation is off.

src/Database/Adapter/SQL.php
```diff
@@ -259,7 +282,7 @@ public function getCountOfIndexes(Document $collection): int
      */
     public static function getCountOfDefaultAttributes(): int
     {
-        return 4;
+        return 5;
```
eldadfux, line 285: Is there any way for us to count this? Can we use the attributes constant?

src/Database/Adapter/SQL.php
```diff
@@ -269,7 +292,7 @@ public static function getCountOfDefaultAttributes(): int
      */
     public static function getCountOfDefaultIndexes(): int
     {
-        return 5;
+        return 6;
```
eldadfux, line 295: Is there any way for us to count this? Can we use the attributes constant?

src/Database/Adapter/SQL.php
```diff
@@ -304,34 +327,13 @@ public function getAttributeWidth(Document $collection): int
...
-                            $total += ($attribute['size'] * 4) + 2;
-                            break;
-
-                        default:
-                            // $size = $size * 4; // utf8mb4 up to 4 bytes per char
-                            // 8 bytes length + 1 bytes for VARCHAR(<=255)
-                            $total += ($attribute['size'] * 4) + 1;
-                            break;
-                    }
+                    $total += match (true) {
```
eldadfux, line 330: Cool, but let's keep the comments.

src/Database/Adapter/Mongo.php
```diff
@@ -1609,7 +1638,7 @@ public static function getCountOfDefaultAttributes(): int
      */
     public static function getCountOfDefaultIndexes(): int
     {
-        return 5;
+        return 6;
```
eldadfux, line 1641: Is there any way for us to count this? Can we use the attributes constant?

src/Database/Adapter/MariaDB.php
eldadfux: I think we can always keep the index for simplicity and just set the tenant value to something default if not used.

abnegate: @fogelito What would the impact on performance be if querying on only `_createdAt` (for existing projects)? The DB will still use the composite index even though it only matches one column, right?

Wondering about which order we should use as well because `(_tenant,_createdAt)` will perform much better for shared tables since tenant will massively reduce the rows, but `(_createdAt,_tenant)` would be much better for not-shared tables since tenant will always be null. 

Maybe it is worth conditionally creating them depending on whether shared tables is set or not

abnegate: Changed this so we conditionally create the indexes optimized for performance

src/Database/Database.php
abnegate: Done

src/Database/Adapter.php
```diff
@@ -133,16 +117,72 @@ public function setDefaultDatabase(string $name, bool $reset = false): bool
...
+    /**
+     * Set Share Tables.
+     *
+     * Set whether to share tables between tenants
+     *
+     * @param bool $shareTables
+     *
+     * @return bool
+     */
+    public function setShareTables(bool $shareTables): bool
```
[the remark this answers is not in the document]
abnegate: If we assume that, we also have to assume if it's not set, we're not using shared tables, which might not be the case (and will allow access to other tenants' data)

src/Database/Adapter/SQL.php
abnegate: Tabs vs spaces 😄

src/Database/Adapter/MariaDB.php
```diff
@@ -116,21 +114,34 @@ public function createCollection(string $name, array $attributes = [], array $in
...
 				" . \implode(' ', $indexStrings) . "
-				UNIQUE KEY `_uid` (`_uid`),
-				KEY `_created_at` (`_createdAt`),
-				KEY `_updated_at` (`_updatedAt`)
-			)
 		";
 
+        if ($this->shareTables) {
+            $sql .= "
+				UNIQUE KEY _uid_tenant (_tenant, _uid),
```
fogelito, line 131: Since _uid is much less in numbers than _tenant , and we have _tenant index in first position let's make this index 
UNIQUE KEY _uid_tenant (_uid, _tenant).

src/Database/Adapter/MariaDB.php
```diff
@@ -116,21 +114,34 @@ public function createCollection(string $name, array $attributes = [], array $in
...
+				KEY _created_at (_tenant, _createdAt),
```
fogelito, line 132: Do we have Workers / Maintance that can fetch data by _createdAt for example?
From a glance I tool in the Workers it always go by projected, we need to double check...

tests/e2e/Adapter/Base.php
```diff
@@ -12545,6 +12548,127 @@ public function testEmptyOperatorValues(): void
...
+        $database->createDocument('people', new Document([
+            '$id' => $docId,
+            '$permissions' => [
+                Permission::read(Role::any()),
+            ],
+            'name' => 'Spiderman',
+        ]));
+
+        $doc = $database->getDocument('people', $docId);
+        $this->assertEquals('Spiderman', $doc['name']);
```
fogelito, line 12636: Can we add an assertion that of $tenant (fetched) ===  $tenant1?

src/Database/Adapter/MariaDB.php
abnegate: Makes sense, will update :+1:

src/Database/Adapter/MariaDB.php
abnegate: Everything must be by project ID, since there was no other option previously (everything is in either project or console namespace)

tests/e2e/Adapter/Base.php
abnegate: Yep done

src/Database/Adapter/MariaDB.php
```diff
@@ -139,17 +150,29 @@ public function createCollection(string $name, array $attributes = [], array $in
...
+					_type VARCHAR(12) NOT NULL,
+					_permission VARCHAR(255) NOT NULL,
+					_document VARCHAR(255) NOT NULL,
+					PRIMARY KEY (_id),
 			";
 
+            if ($this->shareTables) {
+                $sql .= "
+					UNIQUE INDEX _index1 (_document, _tenant, _type, _permission),
+					INDEX _permission (_permission, _type, _document, _tenant)
```
fogelito, line 165: Hmm this Index supposed to cover :
SELECT _document
FROM `sharedTables`.`_people_perms`
WHERE _permission IN ('user:a', 'any', 'label:reader')
AND _type = 'read'
AND _tenant = :_tenant

INDEX _permission (_tenant, _type, _permission) 
We do not need _document in this Index, since we do not care about order by.

abnegate: Updated 👌

src/Database/Adapter/MariaDB.php
```diff
@@ -139,17 +150,29 @@ public function createCollection(string $name, array $attributes = [], array $in
...
+				";
+            } else {
+                $sql .= "
+					UNIQUE INDEX _index1 (_document, _type, _permission),
+					INDEX _permission (_permission, _type, _document)
```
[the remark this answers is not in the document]
abnegate: Done

christyjacob4:
Let's setup a test bench to simulate the kind of loads we receive on cloud.

src/Database/Adapter/MariaDB.php
```diff
@@ -139,17 +150,29 @@ public function createCollection(string $name, array $attributes = [], array $in
...
+					_type VARCHAR(12) NOT NULL,
+					_permission VARCHAR(255) NOT NULL,
+					_document VARCHAR(255) NOT NULL,
+					PRIMARY KEY (_id),
 			";
 
+            if ($this->shareTables) {
+                $sql .= "
+					UNIQUE INDEX _index1 (_document, _tenant, _type, _permission),
+					INDEX _permission (_tenant, _permission, _type)
```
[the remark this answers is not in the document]
fanatic75: I think the correct order is _permission, _type only.

src/Database/Adapter/MariaDB.php
```diff
@@ -949,19 +997,27 @@ public function updateDocument(string $collection, Document $document): Document
         }
         if (!empty($removeQuery)) {
             $removeQuery .= ')';
-            $removeQuery = "
+            $sql = "
```
fanatic75, line 1000: I think the remove query and insert query should run in a single transaction, this can cause problems.

fanatic75 (changes requested):

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
fogelito: We are not filtering or order by _document so we do not need it.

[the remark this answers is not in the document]
fogelito: We covers this query: 
SELECT `table_main`.*
            FROM `sharedTables`.`_people` as table_main
            WHERE table_main._uid IN (
                    SELECT _document
                    FROM `sharedTables`.`_people_perms`
                    WHERE _permission IN ('user:a', 'any', 'label:reader')
                      AND _type = 'read'
                      AND _tenant = :_tenant
                ) AND table_main._tenant = :_tenant
            ORDER BY table_main._id ASC
            LIMIT :limit OFFSET :offset;

src/Database/Adapter/MariaDB.php
```diff
@@ -139,17 +150,29 @@ public function createCollection(string $name, array $attributes = [], array $in
...
 
+            if ($this->shareTables) {
+                $sql .= "
+					UNIQUE INDEX _index1 (_document, _tenant, _type, _permission),
+					INDEX _permission (_tenant, _permission, _type)
+				";
+            } else {
+                $sql .= "
+					UNIQUE INDEX _index1 (_document, _type, _permission),
+					INDEX _permission (_permission, _type)
```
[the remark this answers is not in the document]
fogelito: We are not filtering or order by _document so we do not need it.

src/Database/Database.php
```diff
@@ -2160,6 +2320,10 @@ public function renameIndex(string $collection, string $old, string $new): bool
      */
     public function createIndex(string $collection, string $id, string $type, array $attributes, array $lengths = [], array $orders = []): bool
     {
+        if ($this->adapter->getShareTables() && empty($this->adapter->getTenant())) {
+            throw new DatabaseException('Missing tenant. Tenant must be set when table sharing is enabled.');
+        }
+
```
fogelito, line 2326: On fulltext index we can't add _tenant

src/Database/Adapter/MariaDB.php
```diff
@@ -116,21 +114,34 @@ public function createCollection(string $name, array $attributes = [], array $in
...
 				" . \implode(' ', $indexStrings) . "
-				UNIQUE KEY `_uid` (`_uid`),
-				KEY `_created_at` (`_createdAt`),
-				KEY `_updated_at` (`_updatedAt`)
-			)
 		";
 
+        if ($this->shareTables) {
+            $sql .= "
+				UNIQUE KEY _uid_tenant (_uid, _tenant),
```
fogelito, line 131: We need to benchmark: do we need  an index (_tenant, _id) for find queries
 with no filters as `select * from table where _tenant  = 1 order by _id` 
 Or does_created_at index can cover it since (_tenant) is in the first position in Index tree 🤔

fogelito: @abnegate Perhaps we can add it, It will be easier to drop than create it later on..

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
abnegate: I agree here, _permission will make the traversal shorter than _type in this position right @fogelito?

src/Database/Adapter/MariaDB.php
```diff
@@ -711,19 +746,20 @@ public function createDocument(string $collection, Document $document): Document
         foreach (Database::PERMISSIONS as $type) {
             foreach ($document->getPermissionsByType($type) as $permission) {
                 $permission = \str_replace('"', '', $permission);
-                $permissions[] = "('{$type}', '{$permission}', '{$document->getId()}')";
+                $permissions[] = "('{$type}', '{$permission}', '{$document->getId()}', :_tenant)";
```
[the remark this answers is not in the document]
abnegate: It's better to bind here, otherwise, we'll insert an empty string when `$this-tenant` is null

src/Database/Adapter/MariaDB.php
```diff
@@ -651,9 +687,10 @@ public function deleteIndex(string $collection, string $id): bool
     public function createDocument(string $collection, Document $document): Document
     {
         $attributes = $document->getAttributes();
+        $attributes['_tenant'] = $this->tenant;
```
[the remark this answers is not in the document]
abnegate: It will be inserted as `null`

src/Database/Adapter/MariaDB.php
abnegate: All 3 (update data, remove permission, insert permissions) are executed in a transaction [later on](https://github.com/utopia-php/database/blob/main/src/Database/Adapter/MariaDB.php#L1038-L1049), we are just building the queries here

src/Database/Database.php
abnegate: Fixed and added test case :+1:

src/Database/Adapter/MariaDB.php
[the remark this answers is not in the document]
fogelito: On this 
SELECT _document from PERMS
WHERE 
_permission IN ('user:a', 'any', 'label:reader') 
AND _type = 'read' 
AND _tenant = :_tenant;

So the MySQL engine will do this:
SELECT _document PERMS
WHERE _permission = 'user:a' AND _type = 'read' AND _tenant = :_tenant

SELECT _document PERMS
WHERE _permission = 'any' AND _type = 'read'  AND _tenant = :_tenant

SELECT _document PERMS
WHERE _permission ='label:reader'AND _type = 'read'  AND _tenant = :_tenant

So the question is what he will like better from the 3 options
If _permission would be a specific user as 'user:a' , he would like _permission first , but if _permission will be 'any' which has a lot of rows, So I would say last. 
So (without a benchmark is hard to know)

because we have Equal operation on _tenant, _type , that Is why I put them first,
What do you think?

fogelito (approved):

PR #359, opened 2023-11-21
Add check for length in Queries Validator

vermakhushboo:
There are some scenarios where we need both length and queries validation.

### DIFF
--- a/src/Database/Validator/Queries.php
+++ b/src/Database/Validator/Queries.php
@@ -19,13 +19,19 @@ class Queries extends Validator
     protected array $validators;
 
     /**
+     * @var int
+     */
+    protected int $length;
+
+    /**
      * Queries constructor
      *
      * @param array<Base> $validators
      */
-    public function __construct(array $validators = [])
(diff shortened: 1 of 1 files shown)


src/Database/Validator/Queries.php
```diff
@@ -51,6 +57,10 @@ public function isValid($value): bool
             return false;
         }
 
+        if ($this->length && \count($value) > $this->length) {
```
fogelito, line 60: Not sure we need the first since it is constructed with 0?

fogelito: besides this looks good 
@abnegate Do you want to add something?

fogelito:
Besides small comment looks ok 👍 
@abnegate?

src/Database/Validator/Queries.php
abnegate: @fogelito It looks like we are using 0 to mean unlimited, so when 0, the \count($value) > $this->length won't be checked

abnegate (approved):

PR #361, opened 2023-11-23
Feat: preserve dates

Meldiron:
In some scenarios (migration) you want to create/update a document, but don't want the $createdAt or $updatedAt to change.

I added a parameter to allow control of this. Default stays the same as it is, to prevent breaking changes.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -513,16 +513,16 @@
         },
         {
             "name": "laravel/pint",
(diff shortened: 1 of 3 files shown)


abnegate (commented):
Let's use an instance variable instead of adding new parameters. Looks like tests are failing too

PR #365, opened 2023-12-05
Feat json queries

abnegate:
- Use JSON for query syntax


src/Database/Query.php
```diff
@@ -188,292 +201,29 @@ public static function isMethod(string $value): bool
...
-        return match ($method) {
-            'lt' => self::TYPE_LESSER,
-            'lte' => self::TYPE_LESSEREQUAL,
-            'gt' => self::TYPE_GREATER,
-            'gte' => self::TYPE_GREATEREQUAL,
-            'eq' => self::TYPE_EQUAL,
-            default => $method
-        };
-        */
+        return new self($method, $attribute, $values);
```
fogelito, line 226: In the case of a nested query, I think we need to iterate through all queries and make Objects out of them recursively. or should it be done in the OR queries branch?

src/Database/Query.php
```diff
@@ -188,292 +201,29 @@ public static function isMethod(string $value): bool
...
-            case self::TYPE_STARTS_WITH:
-            case self::TYPE_ENDS_WITH:
-                $attribute = $parsedParams[0] ?? '';
-                if (count($parsedParams) < 2) {
-                    return new self($method, $attribute);
-                }
-                return new self($method, $attribute, \is_array($parsedParams[1]) ? $parsedParams[1] : [$parsedParams[1]]);
-
-            case self::TYPE_BETWEEN:
-                return new self($method, $parsedParams[0], [$parsedParams[1], $parsedParams[2]]);
```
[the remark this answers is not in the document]
abnegate: We do this in the validator instead of parser

src/Database/Query.php
abnegate: Makes sense, will update here

PR #366, opened 2023-12-05
Parsing nested

fogelito:


src/Database/Adapter/SQL.php
```diff
@@ -911,4 +918,38 @@ public function getMaxIndexLength(): int
...
+     */
+    abstract protected function getSQLCondition(Query $query): string;
+
+    /**
+     * @param array<Query> $queries
+     * @param string $separator
+     * @return string
+     * @throws Exception
+     */
+    public function getSQLConditions(array $queries = [], string $separator = 'and'): string
```
abnegate, line 935: Let's use capitals for SQL keywords

```suggestion
public function getSQLConditions(array $queries = [], string $separator = 'AND'): string
```

src/Database/Query.php
```diff
@@ -694,6 +519,24 @@ public static function endsWith(string $attribute, string $value): self
         return new self(self::TYPE_ENDS_WITH, $attribute, [$value]);
     }
 
+    /**
+     * @param array<mixed> $queries
```
abnegate, line 523: Let's use `array<Query>` here

src/Database/Query.php
```diff
@@ -694,6 +519,24 @@ public static function endsWith(string $attribute, string $value): self
...
+     * @return Query
+     */
+    public static function or(array $queries): self
+    {
+        return new self(self::TYPE_OR, '', $queries);
+    }
+
+    /**
+     * @param array<mixed> $queries
```
abnegate, line 532: Let's use `array<Query>` here

src/Database/Query.php
```diff
@@ -808,24 +654,18 @@ public static function groupByType(array $queries): array
     }
 
     /**
-     * Iterate over $queries attempting to parse each
+     * Is isNested
      *
-     * @param array<string> $queries
+     * Function will return true if nested method
```
abnegate, line 659:
```suggestion
     * Is this query able to contain other queries
```

src/Database/Validator/Queries.php
```diff
@@ -92,7 +98,9 @@ public function isValid($value): bool
                 Query::TYPE_BETWEEN,
                 Query::TYPE_STARTS_WITH,
                 Query::TYPE_CONTAINS,
-                Query::TYPE_ENDS_WITH => Base::METHOD_TYPE_FILTER,
+                Query::TYPE_ENDS_WITH,
+                Query::TYPE_AND,
+                Query::TYPE_OR => Base::METHOD_TYPE_FILTER,
```
abnegate, line 103: Let's add a `METHOD_TYPE_LOGICAL` for and/or queries

src/Database/Validator/Query/Filter.php
```diff
@@ -172,6 +172,22 @@ public function isValid($value): bool
             case Query::TYPE_IS_NOT_NULL:
                 return $this->isValidAttributeAndValues($attribute, $value->getValues());
 
+            case Query::TYPE_OR:
+            case Query::TYPE_AND:
+                $filters = Query::groupByType($value->getValues())['filters'];
+
+                if(count($value->getValues()) !== count($filters)) {
+                    $this->message = \ucfirst($method) . ' queries requires only filters';
```
abnegate, line 180:
```suggestion
                    $this->message = \ucfirst($method) . ' queries can only contain filter queries';
```

src/Database/Query.php
fogelito: Done!

src/Database/Query.php
fogelito: Done!

src/Database/Query.php
fogelito: Done!

src/Database/Validator/Queries.php
fogelito: I think we will never this hit this case here, since always will be pass TYPE_AND / TYPE_OR

src/Database/Query.php
```diff
@@ -577,10 +414,10 @@ public static function between(string $attribute, mixed $start, mixed $end): sel
      * Helper method to create Query with search method
      *
      * @param string $attribute
-     * @param mixed $value
+     * @param string|int|float|bool $value
      * @return Query
      */
-    public static function search(string $attribute, mixed $value): self
+    public static function search(string $attribute, string|int|float|bool $value): self
```
fogelito, line 420: @abnegate can the term be something else than a String?

abnegate: I think so, but it would be treated as a string by the database anyway right? So I could pass int 100 and it would search for string '100'

abnegate (approved):

src/Database/Query.php
abnegate: Hmm the more I think about it; we should be explicit and force a string up front anyway, rather than relying on the database to convert for us. Will update :+1:

PR #370, opened 2023-12-12
Fix-#3914-invalid-queries causes 500 error

jaysomani:
#### Test Invalid API Call

Added a PHPUnit test case to ensure proper validation of an invalid API call. This test checks if the `isValid` method in the `Documents` class correctly handles the scenario where the provided queries are not an array. The expected behavior is that the method returns `false` and sets an error message indicating that "Queries must be an array."

#### Enhance isArray Method

Improved the `isArray` method in the same class (`Documents`) by fixing a parameter mismatch. The method now correctly checks if the provided value is an array using the `is_array` function, enhancing the reliability of the isArray function.

These changes aim to strengthen the validation logic and improve the overall robustness of the codebase.

### DIFF
--- a/src/Database/Validator/Queries.php
+++ b/src/Database/Validator/Queries.php
@@ -127,7 +127,7 @@ class Queries extends Validator
      */
     public function isArray(): bool
     {
-        return true;
+        return is_array($value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Database/Validator/Queries.php
```diff
@@ -127,7 +127,7 @@ public function isValid($value): bool
      */
     public function isArray(): bool
     {
-        return true;
+        return is_array($value);
```
stnguyen90, line 130: I don't think this is used, but it still shouldn't be changed so that it's consistent with others.

PR #373, opened 2023-12-28
fix relation structure

fogelito:


abnegate (commented):
Looks good, let's add a test to validate

abnegate (approved):

PR #375, opened 2024-01-03
Fix: setAction method

Meldiron:
Attempt to remove setAction, to prevent bleed of error messages between validation calls


Meldiron:
All tests pass, and implementation allignts with Torsten's vision of it. Merging now.

Meldiron:
Next PR down the tree: https://github.com/utopia-php/database/pull/343

PR #377, opened 2024-01-07
Contains  query v2

fogelito:
Injecting the collections attributes to get if an attribute is an array


src/Database/Validator/Index.php
```diff
@@ -113,6 +112,42 @@ public function checkFulltextIndexNonString(Document $index): bool
...
+        $orders = $index->getAttribute('orders', []);
+
+        $arrayAttributes = [];
+        foreach ($attributes as $key => $attribute) {
+            $attribute = $this->attributes[\strtolower($attribute)] ?? new Document();
+
+            if($attribute->getAttribute('array') === true) {
+                // Database::INDEX_UNIQUE Is not allowed! since mariaDB VS MySQL makes the unique Different on values
+                if(!in_array($index->getAttribute('type'), [Database::INDEX_KEY])) {
+                    $this->message = ucfirst($index->getAttribute('type')) . '" index is forbidden on array attributes';
```
abnegate, line 131:
```suggestion
                    $this->message = '"' . ucfirst($index->getAttribute('type')) . '" index is forbidden on array attributes';
```

src/Database/Validator/Index.php
```diff
@@ -113,6 +112,42 @@ public function checkFulltextIndexNonString(Document $index): bool
...
+                    return false;
+                }
+
+                $arrayAttributes[] = $attribute->getAttribute('key', '');
+                if(count($arrayAttributes) > 1) {
+                    $this->message = 'Only a single index can be created on array attributes found "' . implode(',', $arrayAttributes) . '"';
```
abnegate, line 137:
```suggestion
                   $this->message = 'An index may only contain one array attribute';
```

src/Database/Validator/Index.php
```diff
@@ -113,6 +112,42 @@ public function checkFulltextIndexNonString(Document $index): bool
...
+                    return false;
+                }
+
+                $direction = $orders[$key] ?? '';
+                if(!empty($direction)) {
+                    $this->message = 'Invalid index order "' . $direction . '" on array attribute "'. $attribute->getAttribute('key', '') .'"';
```
abnegate, line 143:
```suggestion
                    $this->message = 'An array attribute can not have an index order';
```

src/Database/Validator/Index.php
```diff
@@ -129,6 +164,18 @@ public function checkIndexLength(Document $index): bool
         foreach ($index->getAttribute('attributes', []) as $attributePosition => $attributeName) {
             $attribute = $this->attributes[\strtolower($attributeName)];
 
+            $isArray = $attribute->getAttribute('array', false);
+
+            if($isArray && empty($lengths[$attributePosition])) {
+                $this->message = 'Index length for array not specified';
```
abnegate, line 170: Is this required? Why do we only need length for array attributes?

src/Database/Validator/Index.php
```diff
@@ -129,6 +164,18 @@ public function checkIndexLength(Document $index): bool
...
+                return false;
+            }
+
+            if(!$isArray && $attribute->getAttribute('type') !== Database::VAR_STRING && !empty($lengths[$attributePosition])) {
+                $this->message = 'Key part length are forbidden on "' . $attribute->getAttribute('type') . '" data-type for "' . $attributeName . '"';
```
abnegate, line 175:
```suggestion
                $this->message = 'Cannot set a length on "'. $attribute->getAttribute('type') . '" attributes';
```

src/Database/Validator/Index.php
```diff
@@ -129,6 +164,18 @@ public function checkIndexLength(Document $index): bool
         foreach ($index->getAttribute('attributes', []) as $attributePosition => $attributeName) {
             $attribute = $this->attributes[\strtolower($attributeName)];
 
+            $isArray = $attribute->getAttribute('array', false);
+
+            if($isArray && empty($lengths[$attributePosition])) {
+                $this->message = 'Index length for array not specified';
+                return false;
+            }
```
abnegate, line 172: Let's do this in the `checkArrayIndex` function

src/Database/Validator/Query/Filter.php
```diff
@@ -101,6 +101,18 @@ protected function isValidAttributeAndValues(string $attribute, array $values):
             }
         }
 
+        if(isset($attributeSchema['array'])) {
+            if($attributeSchema['array'] === false && Query::TYPE_CONTAINS === $method) {
+                $this->message = 'Cannot query contains on attribute "' . $attribute . '" because it is not an array.';
+                return false;
```
abnegate, line 107: If the attribute is a string, we need to treat it as a `LIKE` query

docker-compose.yml
```diff
@@ -38,7 +38,7 @@ services:
       POSTGRES_PASSWORD: password
 
   mariadb:
-    image: mariadb:10.7
+    image: mariadb:10.9 # Is it ok to upgrade Appwrite to >= 10.9 for JSON_OVERLAPS
```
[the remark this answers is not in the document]
fogelito: Done!

src/Database/Validator/Index.php
fogelito: Yes.. I added a few tests , we still have the limitation of max index length of 768, 
Now we have a new problem: Integers did not require length, but now Integer is a type Json or Longtext in maria , so how we limit or count the length. I added this check but still maybe we need to hardcode size values in some cases.. this errors will pop up

src/Database/Validator/Index.php
fogelito: Done

src/Database/Validator/Index.php
fogelito: Done!

src/Database/Validator/Query/Filter.php
fogelito: Done!
Do you think there is a better / nicer way to do this , not using convertQueries, because regular queries do not know if metadata is an array attribute.

src/Database/Adapter/MariaDB.php
```diff
@@ -1936,11 +1964,16 @@ protected function getSQLCondition(Query $query): string
...
      * @param bool $signed
+     * @param bool $array
      * @return string
-     * @throws Exception
+     * @throws DatabaseException
      */
-    protected function getSQLType(string $type, int $size, bool $signed = true): string
+    protected function getSQLType(string $type, int $size, bool $signed = true, bool $array = false): string
     {
+        if($array === true) {
```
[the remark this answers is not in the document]
fogelito: Done & added support for getSupportForQueryOverlaps

src/Database/Adapter/MariaDB.php
```diff
@@ -1912,19 +1932,27 @@ protected function getSQLCondition(Query $query): string
...
-                return "table_main.{$attribute} BETWEEN :{$placeholder}_0 AND :{$placeholder}_1";
+                return "`table_main`.{$attribute} BETWEEN :{$placeholder}_0 AND :{$placeholder}_1";
 
             case Query::TYPE_IS_NULL:
             case Query::TYPE_IS_NOT_NULL:
-                return "table_main.{$attribute} {$this->getSQLOperator($query->getMethod())}";
+                return "`table_main`.{$attribute} {$this->getSQLOperator($query->getMethod())}";
 
+            case Query::TYPE_CONTAINS:
+                // todo: change to JSON_OVERLAPS when using mariaDB 10.9
```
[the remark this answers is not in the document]
fogelito: Done!

src/Database/Adapter/MariaDB.php
```diff
@@ -614,43 +608,70 @@ public function renameIndex(string $collection, string $old, string $new): bool
...
+                '$updatedAt' => '_updatedAt',
+                default => $this->filter($attr),
+            };
 
-        foreach ($attributes as $key => $attribute) {
-            $length = $lengths[$key] ?? '';
-            $length = (empty($length)) ? '' : '(' . (int)$length . ')';
-            $order = $orders[$key] ?? '';
-            $attribute = $this->filter($attribute);
+            $attributes[$i] = "`{$attr}`{$length} {$order}";
```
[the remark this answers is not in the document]
fogelito: That would be better, but the problem is there are places createIndex is being called from the adapter itself, Skipping Database.php

src/Database/Adapter/MariaDB.php
```diff
@@ -614,43 +608,70 @@ public function renameIndex(string $collection, string $old, string $new): bool
...
 
-            if (Database::INDEX_FULLTEXT === $type) {
-                $order = '';
+            if(!empty($collectionAttribute['array']) && $collectionAttribute['array'] === true && $this->castIndexArray()) {
+                $attributes[$i] = '(CAST(' . $attr . ' AS char(255) ARRAY))';
```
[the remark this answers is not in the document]
fogelito: This is only for MySQL .. removed this check $collectionAttribute['array'] === true
it is covered by the first.

src/Database/Validator/Index.php
```diff
@@ -113,6 +112,42 @@ public function checkFulltextIndexNonString(Document $index): bool
...
+    {
+        $attributes = $index->getAttribute('attributes', []);
+        $orders = $index->getAttribute('orders', []);
+
+        $arrayAttributes = [];
+        foreach ($attributes as $key => $attribute) {
+            $attribute = $this->attributes[\strtolower($attribute)] ?? new Document();
+
+            if($attribute->getAttribute('array') === true) {
+                // Database::INDEX_UNIQUE Is not allowed! since mariaDB VS MySQL makes the unique Different on values
```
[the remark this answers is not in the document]
fogelito: Mysql checks each value in the array whether it is unique VS MariaDB which checks the whole JSON

src/Database/Validator/Index.php
fogelito: Today in Appwrite's console there is no way to add an Index on a varchar longer than 768.. an error message will pop up. 
Now since there is no way to pass a size limitation, there will be no way to create an index on arrays of integers for example

src/Database/Adapter/MariaDB.php
```diff
@@ -2042,6 +2044,16 @@ public function getSupportForFulltextWildcardIndex(): bool
         return true;
     }
 
+    /**
+     * Does the adapter handle Query Array Overlaps?
+     *
+     * @return bool
+     */
+    public function getSupportForQueryOverlaps(): bool
```
abnegate, line 2052:
```suggestion
    public function getSupportForJSONOverlaps(): bool
```

src/Database/Adapter/SQL.php
```diff
@@ -679,9 +680,16 @@ public function getSupportForCasting(): bool
...
-        return false;
+        return true;
     }
 
+    /**
+     * Does the adapter handle array Overlaps?
+     *
+     * @return bool
+     */
+    abstract public function getSupportForQueryOverlaps(): bool;
```
abnegate, line 691:
```suggestion
    abstract public function getSupportForJSONOverlaps(): bool;
```

src/Database/Adapter/SQL.php
```diff
@@ -14,6 +14,7 @@
 abstract class SQL extends Adapter
 {
     protected mixed $pdo;
+    protected string $like = 'LIKE';
```
abnegate, line 17: Let's add a method `getLikeOperator` for this

src/Database/Adapter/SQL.php
```diff
@@ -775,7 +791,8 @@ protected function getSQLOperator(string $method): string
                 return 'IS NOT NULL';
             case Query::TYPE_STARTS_WITH:
             case Query::TYPE_ENDS_WITH:
-                return 'LIKE';
+            case Query::TYPE_CONTAINS:
+                return $this->like;
```
abnegate, line 795: Do we need the array/string check here?

src/Database/Query.php
```diff
@@ -70,6 +70,7 @@ class Query
 
     protected string $method = '';
     protected string $attribute = '';
+    public bool $attributeArray = false;
```
abnegate, line 73: Let's add a getter method `onArray()` and renamed this `$onArray`

src/Database/Validator/Query/Filter.php
```diff
@@ -22,6 +22,7 @@ class Filter extends Base
     public function __construct(array $attributes = [], int $maxValuesCount = 100)
     {
         foreach ($attributes as $attribute) {
+            $attribute->setAttribute('array', $attribute->getAttribute('array', false));
```
abnegate, line 25: Why do we do this? In case it's null?

src/Database/Validator/Query/Filter.php
```diff
@@ -101,6 +102,16 @@ protected function isValidAttributeAndValues(string $attribute, array $values):
             }
         }
 
+        if($attributeSchema['array'] === false && $method === Query::TYPE_CONTAINS && $attributeSchema['type'] !==  Database::VAR_STRING) {
```
abnegate, line 105: Let's line break the conditions so it's easier to read

src/Database/Validator/Query/Filter.php
```diff
@@ -101,6 +102,16 @@ protected function isValidAttributeAndValues(string $attribute, array $values):
...
+            $this->message = 'Cannot query contains on attribute "' . $attribute . '" because it is not an array.';
```
abnegate, line 106:
```suggestion
            $this->message = 'Cannot query contains on attribute "' . $attribute . '" because it is not an array or string.';
```

src/Database/Validator/Query/Filter.php
```diff
@@ -101,6 +102,16 @@ protected function isValidAttributeAndValues(string $attribute, array $values):
...
+            return false;
+        }
+
+        if($attributeSchema['array'] === true && !in_array($method, [Query::TYPE_CONTAINS, Query::TYPE_IS_NULL, Query::TYPE_IS_NOT_NULL])) {
+            $this->message = 'Cannot query '. $method .' on attribute "' . $attribute . '" because it is an array. Please use ' . Query::TYPE_CONTAINS;
```
abnegate, line 111:
```suggestion
            $this->message = 'Cannot query '. $method .' on attribute "' . $attribute . '" because it is an array.';
```

tests/e2e/Adapter/Base.php
```diff
@@ -1591,23 +1591,230 @@ public function testDeleteDocument(Document $document): void
...
+            size: 0,
+            required: true,
+            array: true
+        ));
+
+        $this->assertEquals(true, static::getDatabase()->createAttribute(
+            $collection,
+            'names',
+            Database::VAR_STRING,
+            size: 255, // Does this mean each Element max is 255? We need to check this on Structure validation?
```
abnegate, line 1625: I don't think we actually validate this do we :thinking: do we just use it to determine the data type?

src/Database/Adapter/SQL.php
fogelito: Done!

src/Database/Query.php
fogelito: Done!

src/Database/Adapter/SQL.php
fogelito: We will hit this case only for String non-array attribute 👍

src/Database/Validator/Query/Filter.php
fogelito: Yes.. 
Since you commented I will move the logic locally inside the function :)

tests/e2e/Adapter/Base.php
fogelito: Hmmm ..
Metadata will save the size value for string / integer and validate the value in the structure validation.. 
Looks like the structure validator does validate according to the size value.. Will add a test for it.
Is this expected behavior for arrays?

abnegate: Should be okay, I would expect it to be the per-item size for arrays :+1:

src/Database/Adapter/Postgres.php
```diff
@@ -2141,4 +2132,12 @@ protected function processException(PDOException $e): void
 
         throw $e;
     }
+
+    /**
+     * @return string
+     */
+    public function getLikeOperator(): string
+    {
+        return 'LIKE';
```
abnegate, line 2141: Should this be `ILIKE` for Postgres?

abnegate (approved):

PR #383, opened 2024-01-24
Integer validator

fogelito:
Structure validator for checking value ranges Numeric attributes

### DIFF
--- a/src/Database/Validator/Structure.php
+++ b/src/Database/Validator/Structure.php
@@ -300,7 +300,7 @@ class Structure extends Validator
             if ($format) {
                 // Format encoded as json string containing format name and relevant format options
                 $format = self::getFormat($format, $type);
-                $validators = [$format['callback']($attribute)];
+                $validators[] = $format['callback']($attribute);
             }
 
             if ($array) { // Validate attribute type for arrays - format for arrays handled separately
--- a/src/Database/Validator/Structure.php
+++ b/src/Database/Validator/Structure.php
@@ -269,6 +269,7 @@ class Structure extends Validator
                     break;
 
                 case Database::VAR_INTEGER:
+                    // We need both Integer and Range because Range implicitly casts non-numeric values
                     $validators[] = new Integer();
                     $max = $size >= 8 ? Database::BIG_INT_MAX : Database::INT_MAX;
                     $min = $signed ? -$max : 0;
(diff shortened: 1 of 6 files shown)


src/Database/Validator/Structure.php
```diff
@@ -249,6 +249,8 @@ public function isValid($document): bool
             $array = $attribute['array'] ?? false;
             $format = $attribute['format'] ?? '';
             $required = $attribute['required'] ?? false;
+            $size = $attribute['size'] ?? 0;
+            $signed = $attribute['signed'] ?? false;
```
abnegate, line 253: Looks like from the API we default to true, with no way for users to change it, let's default to true here as well.

src/Database/Validator/Structure.php
```diff
@@ -260,12 +262,19 @@ public function isValid($document): bool
...
+
+                    if($signed === false) {
+                        //$max *= 2;
+                        $min = 0;
+                    }
+
+                    $validator = new Range($min, $max, Database::VAR_INTEGER);
                     break;
 
                 case Database::VAR_FLOAT:
```
abnegate, line 280: Do we need to do the same for floats?

src/Database/Validator/Structure.php
fogelito: Done!

src/Database/Validator/Structure.php
fogelito: Yep Done!

src/Database/Validator/Structure.php
```diff
@@ -291,7 +300,7 @@ public function isValid($document): bool
             if ($format) {
                 // Format encoded as json string containing format name and relevant format options
                 $format = self::getFormat($format, $type);
-                $validator = $format['callback']($attribute);
+                $validators = [$format['callback']($attribute)];
```
abnegate, line 303: Should this be `$validators[] = $format['callback']($attribute);`?

src/Database/Validator/Structure.php
```diff
@@ -258,26 +261,32 @@ public function isValid($document): bool
...
             switch ($type) {
                 case Database::VAR_STRING:
-                    $size = $attribute['size'] ?? 0;
-                    $validator = new Text($size, min: 0);
+                    $validators[] = new Text($size, min: 0);
                     break;
 
                 case Database::VAR_INTEGER:
-                    $validator = new Integer();
+                    $validators[] = new Integer();
```
abnegate, line 272: Let's add a comment here (and for floats) that we need both `Integer` and `Range` because `Range` implicitly casts non-numeric values

src/Database/Validator/Structure.php
fogelito: It does make sense to have boundaries for the format.
Changed 👍

src/Database/Validator/Structure.php
fogelito: Done!

PR #387, opened 2024-02-15
Remove Tenant for find and getDocument

fogelito:


src/Database/Adapter/SQLite.php
```diff
@@ -152,6 +156,7 @@ public function createCollection(string $name, array $attributes = [], array $in
         $this->createIndex($id, '_updated_at', Database::INDEX_KEY, [ '_updatedAt'], [], []);
 
         if ($this->shareTables) {
+            // todo are we missing here _tenant attribute?
```
abnegate, line 159: It's added in the `createIndex` function for SQLite I think

fogelito: Yep!

src/Database/Database.php
```diff
@@ -5347,4 +5347,20 @@ private function purgeRelatedDocuments(Document $collection, string $id): void
             $this->cache->purge($key);
         }
     }
+
+    /**
+     * @return  array<array<string, mixed>>
+     */
+    public function getInternalIds(): array
```
abnegate, line 5354: Let's rename this `getPublicInternalAttributes`

abnegate (approved):

PR #391, opened 2024-03-04
Feat 0.45 isolation modes

abnegate:
Backport isolation modes to 0.45.x


src/Database/Adapter.php
```diff
@@ -73,15 +71,11 @@ public function resetDebug(): self
      * @param string $namespace
      *
      * @return bool
-     * @throws Exception
+     * @throws DatabaseException
      *
      */
     public function setNamespace(string $namespace): bool
     {
-        if (empty($namespace)) {
```
eldadfux, line 81: Do we need to remove this check?

src/Database/Adapter.php
```diff
@@ -93,15 +87,10 @@ public function setNamespace(string $namespace): bool
      * Get namespace of current set scope
      *
      * @return string
-     * @throws DatabaseException
      *
      */
     public function getNamespace(): string
     {
-        if (empty($this->namespace)) {
```
eldadfux, line 101: Same, do we need to remove this check?

src/Database/Adapter.php
```diff
@@ -133,16 +117,72 @@ public function setDefaultDatabase(string $name, bool $reset = false): bool
...
+    /**
+     * Set Share Tables.
+     *
+     * Set whether to share tables between tenants
+     *
+     * @param bool $shareTables
+     *
+     * @return bool
+     */
+    public function setShareTables(bool $shareTables): bool
```
eldadfux, line 141: Share or Shared?

src/Database/Adapter/MariaDB.php
```diff
@@ -116,21 +114,35 @@ public function createCollection(string $name, array $attributes = [], array $in
...
-				PRIMARY KEY (`_id`),
 				" . \implode(' ', $indexStrings) . "
-				UNIQUE KEY `_uid` (`_uid`),
-				KEY `_created_at` (`_createdAt`),
-				KEY `_updated_at` (`_updatedAt`)
-			)
 		";
 
+        if ($this->shareTables) {
+            $sql .= "
```
eldadfux, line 129: Do we mind creating this anyway?

src/Database/Adapter/MariaDB.php
```diff
@@ -139,17 +151,29 @@ public function createCollection(string $name, array $attributes = [], array $in
...
-				CREATE TABLE IF NOT EXISTS `{$database}`.`{$namespace}_{$id}_perms` (
-					`_id` int(11) unsigned NOT NULL AUTO_INCREMENT,
-					`_type` VARCHAR(12) NOT NULL,
-					`_permission` VARCHAR(255) NOT NULL,
-					`_document` VARCHAR(255) NOT NULL,
-					PRIMARY KEY (`_id`),
-					UNIQUE INDEX `_index1` (`_document`,`_type`,`_permission`),
-					INDEX `_permission` (`_permission`,`_type`,`_document`)
-				)
+				CREATE TABLE IF NOT EXISTS {$this->getSQLTable($id . '_perms')} (
```
eldadfux, line 154: Do we mind creating this anyway?

src/Database/Database.php
```diff
@@ -165,6 +155,15 @@ class Database
             'array' => false,
             'filters' => [],
         ],
+        [
```
eldadfux, line 158: How come we missed this?

src/Database/Database.php
```diff
@@ -174,6 +173,16 @@ class Database
...
```
eldadfux, line 176: Thinking about it.... $tenant or $tenantId? Should the value for this be internal ID or UID? Might be safer with internalId in case of rapid deletion and creation.

src/Database/Database.php
```diff
@@ -218,7 +232,7 @@ class Database
      *
      * @var array<string, mixed>
      */
-    protected array $collection = [
+    public const COLLECTION = [
```
eldadfux, line 235: Why this need to be public?

src/Database/Database.php
```diff
@@ -643,6 +659,60 @@ public function clearTimeout(string $event = Database::EVENT_ALL): void
         $this->adapter->clearTimeout($event);
     }
 
+    /**
+     * Enable validation
+     *
+     * @return $this
+     */
+    public function enableValidation(): self
```
eldadfux, line 667: What are these two designed for?

- enableValidation
- disableValidation

src/Database/Database.php
```diff
@@ -3806,7 +4060,12 @@ public function decreaseDocumentAttribute(string $collection, string $id, string
      */
     public function deleteDocument(string $collection, string $id): bool
     {
-        $document = Authorization::skip(fn () => $this->silent(fn () => $this->getDocument($collection, $id))); // Skip ensures user does not need read permission for this
+        if ($this->adapter->getShareTables() && empty($this->adapter->getTenant())) {
```
eldadfux, line 4063: I remember we discussed this, but not the details. Please remind me why do we have to repeat this check in so many places?

src/Database/Adapter.php
abnegate: Yep since for shared tables there is no specific namespace

src/Database/Adapter/MariaDB.php
abnegate: We initially did this but ran into issues during migrations

fogelito (approved):

src/Database/Database.php
abnegate: We used to have this as a list of strings as opposed to a list of documents, it was present there. We maintained two lists of internal attributes but I removed the redundant one now

src/Database/Database.php
abnegate: We used `$tenant` to be agnostic, it can be whatever you want. From the Appwrite side, we use the project internal ID

src/Database/Database.php
abnegate: Can't see why 🤔 will revert to protected

src/Database/Database.php
abnegate: These are mostly used for migrations so we can disable ID length validations, since some of our internal IDs exceed the max (36) allowed by the validators

src/Database/Database.php
abnegate: If shared tables is enabled, there must be a tenant set, otherwise we have no way to tell if the the current request is allowed to access the documents

PR #393, opened 2024-03-05
Fix: Proxy tests

Meldiron:
Tests were not running properly. This addresses that, as well as any new errors that popped up with this fix

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,8 +87,6 @@ services:
     container_name: utopia-database-proxy
     networks:
       - database
-    ports:
-      - "8720:80"
     environment:
       - UTOPIA_DATA_API_SECRET=test-secret
       - UTOPIA_DATA_API_SECRET_CONNECTION=mariadb://root:password@database-proxy-mariadb:3306/utopiaTests
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     volumes:
       - ./bin:/usr/src/code/bin
       - ./src:/usr/src/code/src
-      - ./vendor/utopia-php/fetch:/usr/src/code/vendor/utopia-php/fetch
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
     volumes:
       - ./bin:/usr/src/code/bin
       - ./src:/usr/src/code/src
+      # - ./vendor/utopia-php/fetch:/usr/src/code/vendor/utopia-php/fetch
```
abnegate, line 15: Let's remove instead

```suggestion
```

src/Database/Adapter/Proxy.php
```diff
@@ -48,14 +48,15 @@ public function __construct(string $endpoint, string $secret, string $database)
...
+        $roles = Authorization::getRoles();
+
         $response = Client::fetch(
             url: $this->endpoint . '/queries',
             headers: [
                 'x-utopia-secret' => $this->secret,
                 'x-utopia-database' => $this->database,
                 'x-utopia-namespace' => $this->getNamespace(),
-                'x-utopia-auth-roles' => $roles,
+                'x-utopia-auth-roles' => \strval(\json_encode($roles) ?: ''),
```
abnegate, line 59: Do you need strval here? Won't both options always be string?

tests/e2e/Adapter/Base.php
```diff
@@ -6784,6 +6784,10 @@ public function testOneToManyOneWayRelationship(): void
         //Todo: This is failing
         $this->assertEquals($albumDocument->getAttribute('name'), $artist1->getAttribute('albums')[0]->getAttribute('name'));
 
+
+        \var_dump("Sleeping 1");
+        sleep(5);
+
```
abnegate, line 6790: Let's remove these

abnegate (approved):

PR #402, opened 2024-04-07
Invalid values relations update

fogelito:
Update relationships fixes and logic improvement & added more descriptive for error messages.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,7 +7,7 @@
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
     processIsolation="false"
-    stopOnFailure="true">
+    stopOnFailure="false">
     <testsuites>
         <testsuite name="unit">
             <directory>./tests/unit</directory>
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -7,7 +7,7 @@
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
     processIsolation="false"
-    stopOnFailure="false">
+    stopOnFailure="true">
     <testsuites>
         <testsuite name="unit">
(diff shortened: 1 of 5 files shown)


phpunit.xml
```diff
@@ -7,7 +7,7 @@
     convertNoticesToExceptions="true"
     convertWarningsToExceptions="true"
     processIsolation="false"
-    stopOnFailure="false">
+    stopOnFailure="true">
```
abnegate, line 10: Let's revert

src/Database/Exception/InvalidRelationshipValue.php
abnegate, line 1: Let's name this `InvalidRelationshipValueException`

src/Database/Database.php
```diff
@@ -5407,4 +5421,19 @@ public function getInternalAttributes(): array
...
+     * @return bool
+     */
+    protected function arrayIsList(mixed $param): bool
+    {
+        if(!is_array($param)) {
+            return false;
+        }
+
+        //todo in php 8.1 use array_is_list
+        return \array_keys($param) === \range(0, count($param) - 1);
```
abnegate, line 5436: Let's move this check inline instead of having a separate function

phpunit.xml
fogelito: Done!

src/Database/Exception/InvalidRelationshipValue.php
fogelito: Done!

src/Database/Database.php
fogelito: Done

abnegate (approved):

PR #406, opened 2024-04-14
Delete relation oneToMany & manyToMany

fogelito:


abnegate:
Rebased to 0.45 here: https://github.com/utopia-php/database/pull/407

PR #407, opened 2024-04-15
Fix relationship many-to-one and one-to-many deletes

abnegate:


src/Database/Adapter/MariaDB.php
```diff
@@ -510,12 +508,12 @@ public function deleteRelationship(
...
                 if ($side === Database::RELATION_SIDE_PARENT) {
                     $sql = "ALTER TABLE {$relatedTable} DROP COLUMN `{$twoWayKey}`;";
-                } elseif ($twoWay) {
+                } else {
                     $sql = "ALTER TABLE {$table} DROP COLUMN `{$key}`;";
                 }
                 break;
             case Database::RELATION_MANY_TO_ONE:
-                if ($twoWay && $side === Database::RELATION_SIDE_CHILD) {
+                if ($side === Database::RELATION_SIDE_CHILD) {
```
fogelito, line 516: I removed $twoWay please double check me please?

fogelito: Since we have relationship logic inside the adapter, we can create the Indexes as well , not having metadata control them 🤔

src/Database/Adapter/MariaDB.php
```diff
@@ -445,12 +445,14 @@ public function updateRelationship(
                 }
                 break;
             case Database::RELATION_ONE_TO_MANY:
+                // todo: do we care about $twoWay? or side like we do in deleteRelations
```
fogelito, line 448: Do we need to check side?

src/Database/Adapter/MariaDB.php
```diff
@@ -445,12 +445,14 @@ public function updateRelationship(
...
                 if ($twoWay && !\is_null($newTwoWayKey)) {
                     $sql = "ALTER TABLE {$relatedTable} RENAME COLUMN `{$twoWayKey}` TO `{$newTwoWayKey}`;";
                 }
                 break;
             case Database::RELATION_MANY_TO_ONE:
-                if (!\is_null($newKey)) {
+                // todo: do we care about $twoWay? or side like we do in deleteRelations
```
fogelito, line 454: Do we need to check side?

src/Database/Database.php
```diff
@@ -1913,6 +1913,7 @@ public function updateRelationship(
                     }
                     break;
                 case self::RELATION_ONE_TO_MANY:
+                    // todo: do we care about size updating?
```
fogelito, line 1916: Do we need to check side?

abnegate: I think so, will update and test

fogelito (approved):

PR #424, opened 2024-05-15
Throw original Error

fogelito:
- To be consistent with all methods and see the original traceroute


src/Database/Adapter/MariaDB.php
```diff
@@ -1645,12 +1645,12 @@ public function deleteDocument(string $collection, string $id): bool
...
                 throw new DatabaseException('Failed to commit transaction');
             }
-        } catch (\Throwable $th) {
+        } catch (\Throwable $e) {
             if($this->getPDO()->inTransaction()) {
                 $this->getPDO()->rollBack();
             }
 
-            throw new DatabaseException($th->getMessage());
+            throw $e;
```
abnegate, line 1653: I think we want to do `throw new DatabaseException($e->getMessage(), $e->getCode(), previous: $e);` so we can get the trace, but still catch and wrap it in Appwrite

fogelito: $e->getCode() Will be a problem since DatabaseException code is an Integer.
I added a **_previous_** param to the constructor.
why parent::__construct($message, $code, $previous); 
does not exist in the  DatabaseException constructor 🤔 ?

abnegate (approved):

PR #436, opened 2024-08-12
Implement Renaming Attributes

PineappleIOnic:
This PR implements the ability to update attribute's names by adding $newKey to `updateAttributes()`; It also includes `xdebug` for easier debugging.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -314,7 +314,13 @@ class MariaDB extends SQL
         $id = $this->filter($id);
         $type = $this->getSQLType($type, $size, $signed, $array);
 
-        $sql = "ALTER TABLE {$this->getSQLTable($name)} CHANGE COLUMN `{$id}` {$newKey} {$type};";
+        if (!empty($newKey)) {
+            $newKey = $this->filter($newKey);
+            $sql = "ALTER TABLE {$this->getSQLTable($name)} CHANGE COLUMN `{$id}` {$newKey} {$type};";
+        } else {
+            $sql = "ALTER TABLE {$this->getSQLTable($name)} MODIFY `{$id}` {$type};";
+        }
+
         $sql = $this->trigger(Database::EVENT_ATTRIBUTE_UPDATE, $sql);
 
         return $this->getPDO()
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -303,18 +303,18 @@ class MariaDB extends SQL
      * @param int $size
      * @param bool $signed
      * @param bool $array
+     * @param string $newKey
      * @return bool
      * @throws Exception
      * @throws PDOException
      */
-    public function updateAttribute(string $collection, string $id, string $type, int $size, bool $signed = true, bool $array = false): bool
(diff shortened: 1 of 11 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -303,18 +303,18 @@ public function createAttribute(string $collection, string $id, string $type, in
...
      * @return bool
      * @throws Exception
      * @throws PDOException
      */
-    public function updateAttribute(string $collection, string $id, string $type, int $size, bool $signed = true, bool $array = false): bool
+    public function updateAttribute(string $collection, string $id, string $type, int $size, bool $signed = true, bool $array = false, string $newKey = null): bool
     {
         $name = $this->filter($collection);
         $id = $this->filter($id);
         $type = $this->getSQLType($type, $size, $signed, $array);
```
abnegate, line 315: Let's `filter` the newKey as well

tests/e2e/Adapter/Base.php
```diff
@@ -5725,6 +5725,36 @@ public function testUpdateAttributeStructure(): void
...
+            collection: 'rename',
+            id: 'rename_me',
+            newKey: 'renamed',
+        );
+
+        $doc = static::getDatabase()->getDocument('rename', $doc->getId());
+
+        $this->assertEquals('rename me', $doc->getAttribute('renamed'));
+        $this->assertNull($doc->getAttribute('rename_me'));
+    }
```
abnegate, line 5756: Let's also add an assertion that it's renamed when calling `getCollection` and checking the `attributes` attribute

tests/e2e/Adapter/Base.php
```diff
@@ -5725,6 +5725,89 @@ public function testUpdateAttributeStructure(): void
...
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'renamed' => 'string'
+        ]));
+
+        $this->assertEquals('string', $doc->getAttribute('renamed'));
+
+        // Make sure we can't create a new attribute with the old key
+        $this->expectException(StructureException::class);
```
abnegate, line 5799: We should avoid expectException, it can lead to false positive test passes in the future. For example, if I was to add more checks to this test after the call that throws the exception, they would never run, but the test would pass, because an exception that was expected was thrown.

A better pattern here is to use a try/catch, and fail if no exception was thrown. E.g.

```php
try {
    $this->callThatShouldThrow();
    $this->fail('Failed to throw exception');
} catch(\Exception $e) {
    $this->assertInstanceOf(TheException::class, $e)
}
```

tests/e2e/Adapter/Base.php
```diff
@@ -5725,6 +5725,89 @@ public function testUpdateAttributeStructure(): void
...
+                Permission::read(Role::any()),
+                Permission::create(Role::any()),
+                Permission::update(Role::any()),
+                Permission::delete(Role::any()),
+            ],
+            'renamed' => 'string'
+        ]));
+
+        $this->assertEquals('string', $doc->getAttribute('renamed'));
+
```
abnegate, line 5797: Let's also add a check that updating a document with new key works as expected

abnegate (approved):

PR #437, opened 2024-08-14
Feat migrations

abnegate:
- Add Mirror (previously `ProxyDatabase`) implementation
- Support single metadata document per collection per project

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -336,10 +336,10 @@ class MariaDB extends SQL
     {
         $name = $this->filter($collection);
         $id = $this->filter($id);
+        $newKey = empty($newKey) ? null : $this->filter($newKey);
         $type = $this->getSQLType($type, $size, $signed, $array);
 
         if (!empty($newKey)) {
-            $newKey = $this->filter($newKey);
             $sql = "ALTER TABLE {$this->getSQLTable($name)} CHANGE COLUMN `{$id}` {$newKey} {$type};";
         } else {
             $sql = "ALTER TABLE {$this->getSQLTable($name)} MODIFY `{$id}` {$type};";
@@ -347,9 +347,13 @@ class MariaDB extends SQL
 
         $sql = $this->trigger(Database::EVENT_ATTRIBUTE_UPDATE, $sql);
 
-        return $this->getPDO()
+        try {
+            return $this->getPDO()
             ->prepare($sql)
             ->execute();
+        } catch (PDOException $e) {
+            throw $this->processException($e);
+        }
     }
 
     /**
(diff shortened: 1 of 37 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -92,6 +92,23 @@ public function createCollection(string $name, array $attributes = [], array $in
                 $attribute->getAttribute('array', false)
             );
 
+            // Ignore relationships with virtual attributes
```
fogelito, line 95: Is this if block permanent or only for migration?

src/Database/Adapter/MariaDB.php
```diff
@@ -1091,7 +1122,7 @@ public function updateDocument(string $collection, string $id, Document $documen
 			";
 
             if ($this->sharedTables) {
-                $sql .= ' AND _tenant = :_tenant';
+                $sql .= ' AND (_tenant = :_tenant OR _tenant IS NULL)';
```
fogelito, line 1125: If this change only for migrations for updating the _tenant attribute?

src/Database/Adapter/MariaDB.php
```diff
@@ -143,64 +160,67 @@ public function createCollection(string $name, array $attributes = [], array $in
 		";
 
         if ($this->sharedTables) {
-            $sql .= "
+            $collection .= "
             	_tenant INT(11) UNSIGNED DEFAULT NULL,
 				UNIQUE KEY _uid (_tenant, _uid),
```
fogelito, line 165: Old change but I think we want the UNIQUE KEY _uid (_uid, _tenant)
Since _uid has much less values in the top BTREE level?

fogelito (approved):

src/Database/Adapter/MariaDB.php
abnegate: This is to allow a single metadata document for all projects, instead of one for each project. Now when a collection is created, if the tenant is set to null in the metadata document, all users can access that collection

src/Database/Adapter/MariaDB.php
abnegate: Yep makes sense, will update

src/Database/Adapter/MariaDB.php
fogelito: ok

src/Database/Adapter/MariaDB.php
```diff
@@ -1874,7 +1881,7 @@ public function find(string $collection, array $queries = [], ?int $limit = 25,
         }
 
         if ($this->sharedTables) {
-            $where[] = "table_main._tenant = :_tenant";
+            $where[] = "(table_main._tenant = :_tenant OR table_main._tenant IS NULL)";
```
fogelito, line 1884: This may affect not using the index

abnegate: Ran some benchmarks, saw no difference in query time for collections with 1m docs. Since it's just a logic change if we need to update it in future, will leave as is for now

PR #438, opened 2024-08-16
Feat: Datetime validator precision

Meldiron:


src/Database/Validator/Datetime.php
```diff
@@ -29,7 +44,23 @@ public function __construct(?bool $requireDateInFuture = null)
      */
     public function getDescription(): string
     {
-        return $this->message;
+        $message = 'Value must be valid date';
+
+        if($this->offset < 0) {
+            if($this)
```
christyjacob4, line 50: `if this` ??

src/Database/Validator/Datetime.php
```diff
@@ -6,21 +6,36 @@
...
+
+    /**
+     * @var int minimum offset from now in seconds
+     */
+    protected int $offset = 0;
+
+    /**
+     * @param bool $offset minimum offset from now in seconds
+     */
+    public function __construct(bool $requireDateInFuture = false, string $precision = self::PRECISION_ANY, int $offset = 0)
```
christyjacob4, line 34: precision and offset are only relevant if requireDateInFuture = true. 
So they can be both optional params

src/Database/Validator/Datetime.php
```diff
@@ -49,11 +80,45 @@ public function isValid($value): bool
...
+                $diff = $date->getTimestamp() - $now->getTimestamp();
+                if($diff <= $this->offset) {
+                    return false;
+                }
+            } else if($this->offset < 0) {
+                $diff = $now->getTimestamp() - $date->getTimestamp();
+                if($diff <= \abs($this->offset)) {
+                    return false;
+                }
+            }
```
christyjacob4, line 96: offset should be +ve only so we can limit the amount of edge cases.

src/Database/Validator/Datetime.php
Meldiron: copy&paste mistake, thanks

src/Database/Validator/Datetime.php
Meldiron: They have default value, and that makes both optional params

src/Database/Validator/Datetime.php
Meldiron: Removed support for negative offset

christyjacob4 (approved):

PR #447, opened 2024-09-10
Batch Deletes

PineappleIOnic:


src/Database/Adapter/MariaDB.php
```diff
@@ -1635,6 +1635,61 @@ public function deleteDocument(string $collection, string $id): bool
...
+        $sqlWhere = !empty($where) ? 'WHERE ' . implode(' AND ', $where) : '';
+
+        $selections = $this->getAttributeSelections($queries);
+
+        $sql = "
+            USE {$this->database};
+            DELETE {$this->getAttributeProjection($selections, 'table_main')}
+            FROM {$this->getSQLTable($name)} as table_main
+            {$sqlWhere};
+        ";
```
abnegate, line 1671: Since we are always doing a single table delete here:

```suggestion
        $sql = "
            DELETE FROM {$this->getSQLTable($name)}
            {$sqlWhere};
        ";
```

src/Database/Database.php
```diff
@@ -4970,6 +4970,85 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
+
+        $deleted = $this->withTransaction(function () use ($collection, $queries, $batchSize) {
+            $lastDocument = null;
+            while (true) {
+                $affectedDocuments = $this->find($collection->getId(), array_merge(
+                    empty($lastDocument) ? [
+                        Query::limit($batchSize),
+                    ] : [
+                        Query::limit($batchSize),
+                        Query::cursorAfter($lastDocument),
```
fogelito, line 5005: Since we filtered out remaining only with 'filters' queries not sure we have cursor queries. 
We should add a test.

src/Database/Database.php
```diff
@@ -4970,6 +4970,85 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
+                    ],
+                    $queries,
+                ));
+
+                if (empty($affectedDocuments)) {
+                    break;
+                }
+
+                foreach ($affectedDocuments as $document) {
```
fogelito, line 5014: Since we read & loop through all documents, and we have the $internalId, We can pass an array of InternalIds to the adapter , instead of queries @abnegate WDU

src/Database/Database.php
```diff
@@ -4970,6 +4970,85 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
+            throw new DatabaseException('Missing tenant. Tenant must be set when table sharing is enabled.');
+        }
+
+        $queries = Query::groupByType($queries)['filters'];
+        $collection = $this->silent(fn () => $this->getCollection($collection));
+
+        $deleted = $this->withTransaction(function () use ($collection, $queries, $batchSize) {
+            $lastDocument = null;
+            while (true) {
+                $affectedDocuments = $this->find($collection->getId(), array_merge(
```
fogelito, line 5000: I think we can skipRelations for the find.

src/Database/Database.php
```diff
@@ -4970,6 +4970,85 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
+                    }
+
+                    // Fire events
+                    $this->trigger(self::EVENT_DOCUMENT_DELETE, $document);
+
+                    $this->purgeRelatedDocuments($collection, $document->getId());
+                    $this->purgeCachedDocument($collection->getId(), $document->getId());
+                }
+
+                if (count($affectedDocuments) < 100) {
```
fogelito, line 5038: Use $batchSize

src/Database/Database.php
abnegate: They will be lost, we should group the filters into $filters

src/Database/Adapter/MariaDB.php
```diff
@@ -1632,6 +1632,79 @@ public function deleteDocument(string $collection, string $id): bool
...
+
+        try {
+            $name = $this->filter($collection);
+            $where = [];
+
+            if ($this->sharedTables) {
+                $where[] = "_tenant = :_tenant";
+            }
+
+            $where[] = "_uid IN (" . \implode(', ', \array_map(fn ($id) => ":_id_{$id}", \array_keys($ids))) . ")";
```
abnegate, line 1657: Let's use index here 

```suggestion
            $where[] = "_uid IN (" . \implode(', ', \array_map(fn ($index) => ":_id_{$index}", \array_keys($ids))) . ")";
```

src/Database/Adapter/MariaDB.php
```diff
@@ -1632,6 +1632,79 @@ public function deleteDocument(string $collection, string $id): bool
...
+
+            $sql = "DELETE FROM {$this->getSQLTable($name)} WHERE " . \implode(' AND ', $where);
+
+            $sql = $this->trigger(Database::EVENT_DOCUMENT_DELETE, $sql);
```
abnegate, line 1661: Let's use the bulk delete event here

src/Database/Database.php
```diff
@@ -4970,6 +4971,88 @@ private function deleteCascade(Document $collection, Document $relatedCollection
...
+                foreach ($affectedDocuments as $document) {
+                    if ($collection->getId() !== self::METADATA) {
+                        $documentSecurity = $collection->getAttribute('documentSecurity', false);
+                        $isValid = $this->authorization->isValid(new Input(self::PERMISSION_DELETE, [
+                            ...$collection->getDelete(),
+                            ...($documentSecurity ? $document->getDelete() : [])
+                        ]));
+                        if (!$isValid) {
+                            throw new AuthorizationException($this->authorization->getDescription());
+                        }
```
abnegate, line 5028: If we pass `delete` to find then to getSQLPermissions like we do for update, we can remove this check for each document

abnegate (approved):

PR #449, opened 2024-09-17
Make findOne return empty document instead of false

PineappleIOnic:
This PR makes `findOne` return an empty document instead of a false when no document is found. Tests have been updated appropriately.


fogelito (commented):
Looks good 👍 
Will have to merge once all dependencies are updated

stnguyen90 (approved):

PR #453, opened 2024-10-07
Update FindOne empty return value

PineappleIOnic:
Make findOne return empty document instead of false when no matches are found

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -5188,7 +5188,7 @@ class Database
         if (!$found) {
             return new Document();
         }
-        
+
         return $found;
     }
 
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -5185,7 +5185,11 @@ public function findOne(string $collection, array $queries = []): false|Document
 
         $this->trigger(self::EVENT_DOCUMENT_FIND, $found);
 
-        return $found;
+        if (!$found) {
+            return new Document();
+        } else {
+            return $found;
+        }
```
abnegate, line 5192: Stylistically and for readability it's a bit better to avoid else if we can:

```suggestion
        if (!$found) {
            return new Document();
        }
        
        return $found;
```

abnegate (approved):

abnegate (approved):

PR #456, opened 2024-10-07
Fix: Shared table operations

Meldiron:
Methods throw exception on shared table.
This copies existing hack to more places, in other to fix upcoming migration.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1934,7 +1934,7 @@ class Database
         $deleted = false;
         try {
             $deleted = $this->adapter->deleteAttribute($collection->getId(), $id);
-        } catch (DuplicateException $e) {
+        } catch (\Throwable $e) {
             // HACK: Metadata should still be updated, can be removed when null tenant collections are supported.
             if (!$this->adapter->getSharedTables()) {
                 throw $e;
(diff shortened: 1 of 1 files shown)


src/Database/Database.php
```diff
@@ -1065,7 +1065,14 @@ public function createCollection(string $id, array $attributes = [], array $inde
             }
         }
 
-        $this->adapter->createCollection($id, $attributes, $indexes);
+        try {
+            $this->adapter->createCollection($id, $attributes, $indexes);
+        } catch (DuplicateException $e) {
```
abnegate, line 1070: This one might be problematic, I don't think it throws duplicate because we use `create if not exists`

src/Database/Database.php
```diff
@@ -1294,7 +1301,14 @@ public function deleteCollection(string $id): bool
             $this->deleteRelationship($collection->getId(), $relationship->getId());
         }
 
-        $this->adapter->deleteCollection($id);
+        try {
+            $this->adapter->deleteCollection($id);
+        } catch (DuplicateException $e) {
```
abnegate, line 1306: Deletes won't throw duplicates, just database exceptions with message "x not found"

src/Database/Database.php
```diff
@@ -1294,7 +1297,14 @@ public function deleteCollection(string $id): bool
             $this->deleteRelationship($collection->getId(), $relationship->getId());
         }
 
-        $this->adapter->deleteCollection($id);
+        try {
+            $this->adapter->deleteCollection($id);
+        } catch (\Throwable $e) {
```
abnegate, line 1302: This will mean shared tables will ignore _any_ exception thrown on delete. Let's go with one of two options:
- Catch and check error code, throw a `NotFound` exception from the adapters
- Not handle deletes here, and manually remove them on patch script failure

Meldiron:
Closed in favour of: https://github.com/utopia-php/database/compare/fix-shared-table-crud?expand=1

PR #459, opened 2024-10-12
Fix: Add safeuards for shared table

Meldiron:


Meldiron:
We are skipping shared tables support for current migraiton

PR #461, opened 2024-10-13
Fix junction strings ids

fogelito:


fogelito:
Closing this PR since we do not want to allow non related UID for relations

PR #462, opened 2024-10-14
Feat bulk update

PineappleIOnic:
This PR implements the ability to bulk update documents using a single partial document and queries.


src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,92 @@ public function updateDocument(string $collection, Document $document): Document
...
+        $conditions = $this->getSQLConditions($queries);
+        if (!empty($conditions)) {
+            $where[] = $conditions;
+        }
 
-                foreach ($bindValues as $key => $value) {
-                    $stmt->bindValue($key, $value, $this->getPDOType($value));
-                }
+        if ($this->authorization->getStatus()) {
+            $where[] = $this->getSQLPermissionsCondition($name, $roles, 'update');
```
abnegate, line 1338: Let's use permissions constants here Database::PERMISSION_UPDATE

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,92 @@ public function updateDocument(string $collection, Document $document): Document
...
+        }
 
-                                if ($this->sharedTables) {
-                                    $addQuery .= ", :_tenant)";
-                                } else {
-                                    $addQuery .= ")";
-                                }
+        $name = $this->filter($collection);
+        $roles = $this->authorization->getRoles();
+        $columns = '';
```
abnegate, line 1326: We need to keep the checks that ensure createdAt, updatedAt and tenant weren't changed

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,92 @@ public function updateDocument(string $collection, Document $document): Document
...
-                    // Get removed Permissions
-                    $removals = [];
-                    foreach (Database::PERMISSIONS as $type) {
-                        $diff = array_diff($permissions[$type], $document->getPermissionsByType($type));
-                        if (!empty($diff)) {
-                            $removals[$type] = $diff;
-                        }
-                    }
-
-                    // Build inner query to remove permissions
```
abnegate, line 1410: We need to make sure we still run the SQL to add/remove permissions documents if needed

src/Database/Validator/PartialStructure.php
```diff
@@ -0,0 +1,301 @@
...
+use Utopia\Database\Exception as DatabaseException;
+use Utopia\Database\Validator\Datetime as DatetimeValidator;
+use Utopia\Http\Validator;
+use Utopia\Http\Validator\Boolean;
+use Utopia\Http\Validator\FloatValidator;
+use Utopia\Http\Validator\Integer;
+use Utopia\Http\Validator\Range;
+use Utopia\Http\Validator\Text;
+
+class PartialStructure extends Validator
```
abnegate, line 18: This is almost the same as the full Structure validator, let's:

- Abstract the logic in the `Structure::isValid` method into smaller methods like `checkForAllRequiredValues`, `checkForUnknownValues` `isValidValue` etc.
- Extend `Structure` validator here
- Only override `isValid` and call the required new check methods (basically just skip `checkForAllRequiredValues`)

src/Database/Adapter.php
```diff
@@ -595,17 +595,17 @@ abstract public function createDocuments(string $collection, array $documents, i
...
      * @param array<Document> $documents
-     * @param int $batchSize
      *
-     * @return array<Document>
+     * @return int
      *
      * @throws DatabaseException
      */
-    abstract public function updateDocuments(string $collection, array $documents, int $batchSize): array;
+    abstract public function updateDocuments(string $collection, Document $update, array $documents): int;
```
abnegate, line 608: Let's change the name from `$update` to `$updates`

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
...
-
-                $old = $this->authorization->skip(fn () => $this->silent(
-                    fn () => $this->getDocument(
-                        $collection->getId(),
-                        $document->getId(),
-                        forUpdate: true
-                    )
-                ));
+        $totalAffectedDocuments = [];
+        $affectedDocumentIds = [];
```
abnegate, line 3924: Looks like these are only used inside the transaction, let's move them inside the callback

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+
+                if (empty($affectedDocuments)) {
+                    break;
                 }
 
-                $validator = new Structure($collection);
-                if (!$validator->isValid($document)) {
-                    throw new StructureException($validator->getDescription());
+                foreach ($affectedDocuments as $document) {
+                    // Update relationship, TODO: Figure out.
```
abnegate, line 3956: Let's remove this is the problem is resolved

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
...
             }
 
-            return $this->adapter->updateDocuments($collection->getId(), $documents, $batchSize);
-        });
-
-        foreach ($documents as $key => $document) {
-            if ($this->resolveRelationships) {
-                $document = $this->silent(fn () => $this->populateDocumentRelationships($collection, $document));
+            if (empty($totalAffectedDocuments)) {
+                return;
```
abnegate, line 3975: Should return 0 here

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
...
-            $this->purgeCachedDocument($collection->getId(), $document->getId());
-        }
+            $getResults = fn () => $this->adapter->updateDocuments(
+                $collection->getId(),
+                $update,
+                $totalAffectedDocuments
+            );
 
-        $this->trigger(self::EVENT_DOCUMENTS_UPDATE, $documents);
+            return $skipAuth ? $this->authorization->skip($getResults) : $getResults();
```
abnegate, line 3986: Let's move this inside the while loop so that memory constraints can be controlled with the batch size parameter

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+        });
 
-        return $documents;
+        $this->purgeCachedCollection($collection->getId());
```
abnegate, line 3989: We should only purge documents that were updated

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,257 @@ public function updateDocument(string $collection, Document $document): Document
...
-                    $attributes['_createdAt'] = $document->getCreatedAt();
-                    $attributes['_updatedAt'] = $document->getUpdatedAt();
-                    $attributes['_permissions'] = json_encode($document->getPermissions());
+        $where = [];
 
-                    if ($this->sharedTables) {
-                        $attributes['_tenant'] = $this->tenant;
-                    }
+        $ids = \array_map(fn ($document) => $document->getId(), $documents);
+        $where[] = "_uid IN (" . \implode(', ', \array_map(fn ($id) => ":_id_{$id}", \array_keys($ids))) . ")";
```
abnegate, line 1334: Let's give this a better name for clarity:

```suggestion
$where[] = "_uid IN (" . \implode(', ', \array_map(fn ($index) => ":_id_{$index}", \array_keys($ids))) . ")";
```

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,257 @@ public function updateDocument(string $collection, Document $document): Document
...
-                        $sql .= ' AND _tenant = :_tenant';
-                    }
+        $sql = "
+                UPDATE {$this->getSQLTable($name)} AS table_main
+                SET {$columns}
+                {$sqlWhere}
+            ";
 
-                    $sql = $this->trigger(Database::EVENT_PERMISSIONS_READ, $sql);
+        $sql = $this->trigger(Database::EVENT_DOCUMENT_UPDATE, $sql);
```
abnegate, line 1361: Should be the bulk update event

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,257 @@ public function updateDocument(string $collection, Document $document): Document
...
-                        WHERE _document = :_uid
-                    ";
+            $bindIndex++;
+        }
 
-                    if ($this->sharedTables) {
-                        $sql .= ' AND _tenant = :_tenant';
-                    }
+        $sql = "
+                UPDATE {$this->getSQLTable($name)} AS table_main
```
abnegate, line 1356: I think we can remove the alias to table_main here (and in the shared table where clause)

src/Database/Adapter/MariaDB.php
```diff
@@ -2242,6 +2240,16 @@ public function getSupportForTimeouts(): bool
...
+    /**
+     * Are batch operations supported?
+     *
+     * @return bool
+     */
+    public function getSupportForBatchOperations(): bool
+    {
+        return true;
+    }
+
```
abnegate, line 2252: We can add this in the SQL.php adapter so we don't need to implement for each

src/Database/Adapter/SQLite.php
```diff
@@ -1215,14 +963,14 @@ protected function getSQLIndex(string $collection, string $id, string $type, arr
      * @return string
      * @throws Exception
      */
-    protected function getSQLPermissionsCondition(string $collection, array $roles): string
+    protected function getSQLPermissionsCondition(string $collection, array $roles, string $type = 'read'): string
```
abnegate, line 966: Let's make sure to use the const for read permission everywhere

src/Database/Adapter/SQLite.php
```diff
@@ -815,259 +815,7 @@ public function updateDocument(string $collection, Document $document): Document
...
-     *
-     * @param string $collection
-     * @param array<Document> $documents
-     * @param int $batchSize
-     *
-     * @return array<Document>
-     *
-     * @throws Duplicate
-     */
-    public function updateDocuments(string $collection, array $documents, int $batchSize = Database::INSERT_BATCH_SIZE): array
```
abnegate, line 829: Looks like we need to add the SQLite implementation

tests/e2e/Adapter/Base.php
```diff
@@ -15578,6 +15532,348 @@ public function testTransformations(): void
...
+
+        $this->assertNotFalse($document);
+        $this->assertEquals('text📝 updated', $document->getAttribute('string'));
+
+        $sisterDocument = $this->getDatabase()->getDocument('testUpdateDocumentsRelationships2', 'doc2');
+        $this->assertNotNull($sisterDocument);
+
+        $relationalDocument = $sisterDocument->getAttribute('testUpdateDocumentsRelationships1');
+        $this->assertEquals('text📝 updated', $relationalDocument->getAttribute('string'));
+    }
```
abnegate, line 15875: We should add a check for updating the relationship values themselves:

- Create some more documents in both collections
- Bulk update `testUpdateDocumentsRelationships2` documents with `testUpdateDocumentsRelationships1 = doc1`
- Get all documents from `testUpdateDocumentsRelationships2`
- Assert that the `testUpdateDocumentsRelationships1` attribute has the actual document with `$id = doc1` and `text = text📝`

src/Database/Adapter/SQLite.php
abnegate: Never mind, inherits the MariaDB implementation

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+                    return $this->find($collection->getId(), array_merge(
+                        empty($lastDocument) ? [
+                            Query::limit($batchSize),
+                        ] : [
+                            Query::limit($batchSize),
+                            Query::cursorAfter($lastDocument),
+                        ],
+                        $queries,
+                    ), forPermission: Database::PERMISSION_UPDATE);
+                });
```
abnegate, line 3949: Let's also add a test that matches more documents the $batchSize

src/Database/Database.php
```diff
@@ -3884,83 +3885,110 @@ public function updateDocument(string $collection, string $id, Document $documen
     }
 
     /**
-     * Update Documents in a batch
+     * Batch update documents
```
abnegate, line 3888: Let's make this similar to the batch deletes doc block

src/Database/Database.php
abnegate: We will also need to purgeRelatedDocuments for each document

src/Database/Database.php
```diff
@@ -3884,83 +3885,117 @@ public function updateDocument(string $collection, string $id, Document $documen
...
+        unset($update['$tenant']);
 
-            foreach ($documents as $key => $document) {
-                if (!$document->getId()) {
-                    throw new DatabaseException('Must define $id attribute for each document');
-                }
+        if (!$this->preserveDates) {
+            unset($update['$updatedAt']);
+            $update['$updatedAt'] = DateTime::now();
+        }
```
abnegate, line 3922: We should unset $updatedAt before the preserveDates check

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,262 @@ public function updateDocument(string $collection, Document $document): Document
     }
 
     /**
-     * Update Documents in batches
+     * Batch update documents
      *
      * @param string $collection
+     * @param Document $updates
      * @param array<Document> $documents
```
abnegate, line 1310: Let's update the docblock here too

src/Database/Database.php
```diff
@@ -3884,83 +3885,117 @@ public function updateDocument(string $collection, string $id, Document $documen
...
-     * @return array<Document>
+     * @return int
      *
      * @throws AuthorizationException
-     * @throws Exception
-     * @throws StructureException
+     * @throws DatabaseException
      */
-    public function updateDocuments(string $collection, array $documents, int $batchSize = self::INSERT_BATCH_SIZE): array
+    public function updateDocuments(string $collection, Document $update, array $queries = [], int $batchSize = self::INSERT_BATCH_SIZE): int
```
abnegate, line 3902: Let's call the param $updates here too

src/Database/Database.php
```diff
@@ -3884,83 +3885,117 @@ public function updateDocument(string $collection, string $id, Document $documen
...
 
-            $this->purgeCachedDocument($collection->getId(), $document->getId());
-        }
+            foreach ($affectedDocumentIds as $id) {
+                $this->purgeRelatedDocuments($collection, $id);
+                $this->purgeCachedDocument($collection->getId(), $id);
+            }
 
-        $this->trigger(self::EVENT_DOCUMENTS_UPDATE, $documents);
+            return $result ?? 0;
```
abnegate, line 3995: Should this be `$totalModified`? Let's make sure we have a test that updates more than one page of documents. We can just pass $batchSize = 1 or something to make it easy

src/Database/Validator/PartialStructure.php
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        if (empty($this->collection->getId()) || Database::METADATA !== $this->collection->getCollection()) {
+            $this->message = 'Collection not found';
+            return false;
+        }
+
+        $keys = [];
+        $structure = $document->getArrayCopy();
+        $attributes = $this->collection->getAttribute('attributes', []);
```
abnegate, line 33: Looks redundant

src/Database/Adapter/MariaDB.php
```diff
@@ -1303,260 +1303,262 @@ public function updateDocument(string $collection, Document $document): Document
...
-            $batches = \array_chunk($documents, max(1, $batchSize));
+        if (!empty($updates->getPermissions())) {
+            $attributes['_permissions'] = json_encode($updates->getPermissions());
+        }
 
-            foreach ($batches as $batch) {
-                $bindIndex = 0;
-                $batchKeys = [];
-                $bindValues = [];
+        if (empty($attributes)) {
```
abnegate, line 1328: This will return true even if we pass permissions to update, let's make sure we have a test that covers that

src/Database/Adapter/SQL.php
```diff
@@ -981,7 +991,7 @@ protected function getSQLPermissionsCondition(string $collection, array $roles):
                     SELECT _document
                     FROM {$this->getSQLTable($collection . '_perms')}
                     WHERE _permission IN (" . implode(', ', $roles) . ")
-                      AND _type = 'read'
```
abnegate, line 984: Because this can come from outside the library and we're not binding it, let's throw an exception if $type is not in the `Database::PERMISSIONS` array

src/Database/Adapter/MariaDB.php
PineappleIOnic: We want it to be true if permissions gets updated so it updated _permissions within the row itself

abnegate (approved):

PR #469, opened 2024-11-04
Fix datetime encoding on array

fogelito:


src/Database/Database.php
```diff
@@ -406,13 +406,12 @@ function (mixed $value) {
...
-             * @return string|null
-             * @throws Exception
+             * @param mixed $value
+             * @return mixed
              */
-            function (?string $value) {
+            function (mixed $value) {
                 if (is_null($value)) {
-                    return null;
+                    return;
```
abnegate, line 414: We shouldn't change here, but inside the `encodeAttribute` function, we should check if the value is an array, and encode each element individually. Same for decode

abnegate: Wait I see we already do that, is the issue that we have a `DateTime` attribute where array = false, but the data actually is an array? How could that get past the validator?

abnegate (approved):

PR #470, opened 2024-11-05
Don't throw on deleteAttribute if column doesn't exist

PineappleIOnic:
This PR makes it so that when you deleteAttributes, if the column for the attribute does not exist, the adapter will still return true so the metadata can be cleaned up appropriately. Tests are not included for this as replicating the environment for this isn't easy with the tests architecture we have

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -435,7 +435,7 @@ class MariaDB extends SQL
     /**
      * Rename Attribute
      *
-     * @param string $collectionp
+     * @param string $collection
      * @param string $old
      * @param string $new
      * @return bool
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -421,8 +421,8 @@ class MariaDB extends SQL
 
         try {
             return $this->getPDO()
-            ->prepare($sql)
-            ->execute();
+                ->prepare($sql)
+                ->execute();
         } catch (PDOException $e) {
             if ($e->getCode() === "42000" && $e->errorInfo[1] === 1091) {
                 return true;
@@ -435,7 +435,7 @@ class MariaDB extends SQL
     /**
      * Rename Attribute
      *
-     * @param string $collection
+     * @param string $collectionp
      * @param string $old
      * @param string $new
      * @return bool
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -419,9 +419,17 @@ class MariaDB extends SQL
 
         $sql = $this->trigger(Database::EVENT_ATTRIBUTE_DELETE, $sql);
 
-        return $this->getPDO()
+        try {
+            return $this->getPDO()
             ->prepare($sql)
             ->execute();
+        } catch (PDOException $e) {
+            if ($e->getCode() === "42000" && $e->errorInfo[1] === 1091) {
+                return true;
+            }
+
+            throw $e;
+        }
     }
 
     /**
--- a/src/Database/Adapter/Postgres.php
+++ b/src/Database/Adapter/Postgres.php
@@ -348,10 +348,10 @@ class Postgres extends SQL
 
         try {
             return $this->getPDO()
-            ->prepare($sql)
-            ->execute();
+                ->prepare($sql)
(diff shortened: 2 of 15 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -419,9 +419,17 @@ public function deleteAttribute(string $collection, string $id, bool $array = fa
 
         $sql = $this->trigger(Database::EVENT_ATTRIBUTE_DELETE, $sql);
 
-        return $this->getPDO()
+        try {
+            return $this->getPDO()
             ->prepare($sql)
             ->execute();
+        } catch (PDOException $e) {
```
abnegate, line 426: Indentation here is off, same for Postgres

src/Database/Adapter/Postgres.php
```diff
@@ -346,9 +346,17 @@ public function deleteAttribute(string $collection, string $id, bool $array = fa
 
         $sql = $this->trigger(Database::EVENT_ATTRIBUTE_DELETE, $sql);
 
-        return $this->getPDO()
+        try {
+            return $this->getPDO()
             ->prepare($sql)
             ->execute();
+        } catch (PDOException $e) {
+            if ($e->getCode() === "4242703000" && $e->errorInfo[1] === 7) {
```
abnegate, line 354: Is this really the code 😅

abnegate (commented):
Let's add a test to make sure this works as expected

src/Database/Adapter/Postgres.php
PineappleIOnic: No, I don't know where I got that code. Maybe my keyboard messaged up 😅

tests/e2e/Adapter/MariaDBTest.php
```diff
@@ -56,6 +57,17 @@ public static function getDatabase(bool $fresh = false): Database
 
         $database->create();
 
+        self::$pdo = $pdo;
         return self::$database = $database;
     }
+
+    protected static function deleteColumn(string $collection, string $column): bool
+    {
+        $SQLTable = "`" . self::getDatabase()->getDatabase() . "`.`" . self::getDatabase()->getNamespace() . "_" . $collection . "`";
```
abnegate, line 66: Let's use lower for SQL here `$sqlTable`

tests/e2e/Adapter/Base.php
```diff
@@ -1295,6 +1303,11 @@ public function testCreateDeleteAttribute(): void
         } catch (Exception $e) {
             $this->assertInstanceOf(DuplicateException::class, $e);
         }
+
+        // Test delete attribute when column does not exist
+        $this->assertEquals(true, static::getDatabase()->createAttribute('attributes', 'string1', Database::VAR_STRING, 128, true));
+        $this->assertEquals(true, static::deleteColumn('attributes', 'string1'));
+        $this->assertEquals(true, static::getDatabase()->deleteAttribute('attributes', 'string1'));
```
abnegate, line 1310: Let's also assert that it's remove from the collections metadata after this

tests/e2e/Adapter/MirrorTest.php
```diff
@@ -22,6 +22,7 @@
 class MirrorTest extends Base
 {
     protected static ?Mirror $database = null;
+    protected static ?PDO $pdo = null;
```
abnegate, line 25: Mirror should cache both source and destination PDO and delete the column from both

tests/e2e/Adapter/MirrorTest.php
```diff
@@ -312,4 +314,14 @@ public function testDeleteMirroredDocument(): void
         $this->assertTrue($database->getSource()->getDocument('testDeleteMirroredDocument', $document->getId())->isEmpty());
         $this->assertTrue($database->getDestination()->getDocument('testDeleteMirroredDocument', $document->getId())->isEmpty());
     }
+
+    protected static function deleteColumn(string $collection, string $column): bool
+    {
+        $sqlTable = "`" . self::getDatabase()->getDatabase() . "`.`" . self::getDatabase()->getNamespace() . "_" . $collection . "`";
```
abnegate, line 320: getDatabase and getNamespace should be called on source and destination DBs, not the mirror

abnegate (approved):

PR #472, opened 2024-11-05
Get Connection Id

fogelito:
Which is the value in process list

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -834,7 +834,7 @@ abstract class Adapter
     abstract public function getSupportForAttributeResizing(): bool;
 
     /**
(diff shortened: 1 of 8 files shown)


src/Database/Adapter.php
```diff
@@ -833,6 +833,13 @@ abstract public function getSupportForBatchOperations(): bool;
      */
     abstract public function getSupportForAttributeResizing(): bool;
 
+    /**
+     * Are attributes supported?
```
abnegate, line 837: Let's update here

fogelito: Done!

abnegate (approved):

PR #481, opened 2024-11-13
0.53.x alter attribute indexes

fogelito:


src/Database/Adapter/SQL.php
```diff
@@ -1085,7 +1085,10 @@ public function getMaxVarcharLength(): int
      */
     public function getMaxIndexLength(): int
     {
-        return 768;
+        /**
+         * $tenant int = 1
+         */
+        return $this->sharedTables ? 767 : 768;
```
abnegate, line 1091: This could be more than 1 right?

abnegate (approved):

PR #482, opened 2024-11-14
Primary key bigint

fogelito:


src/Database/Adapter/MariaDB.php
```diff
@@ -215,6 +215,7 @@ public function createCollection(string $name, array $attributes = [], array $in
         } catch (PDOException $e) {
             $e = $this->processException($e);
 
+            // Can we remove drop perms table?
```
abnegate, line 218: Please open these questions as comments on the PR or linear so we don't need code changes to remove them if there's no change to resolve :+1:

For this case, why would we remove it? Wouldn't it result in leftover data?

fogelito: Sure will open comment comment next time 👍 
Not sure if this can happen but , Trying to creat an existing collection will trigger a duplication exeption?

abnegate: I think this case will never actually happen unless you call createCollection directly on the adapter instead of `Database`, since database already does a duplicate check and throws the exception there 🤔

abnegate (approved):

PR #489, opened 2024-12-01
Remove DROP TABLE IF EXISTS

fogelito:
Remove drop table to ensure necessary tables are not accidentally deleted.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -215,11 +215,11 @@ class MariaDB extends SQL
         } catch (PDOException $e) {
             $e = $this->processException($e);
 
-//            if (!($e instanceof DuplicateException)) {
-//                $this->getPDO()
-//                    ->prepare("DROP TABLE IF EXISTS {$this->getSQLTable($id)}, {$this->getSQLTable($id . '_perms')};")
-//                    ->execute();
-//            }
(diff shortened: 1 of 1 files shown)


stnguyen90 (changes requested):
Lint error

stnguyen90 (approved):

src/Database/Adapter/MariaDB.php
```diff
@@ -215,11 +215,11 @@ public function createCollection(string $name, array $attributes = [], array $in
...
-            if (!($e instanceof DuplicateException)) {
-                $this->getPDO()
-                    ->prepare("DROP TABLE IF EXISTS {$this->getSQLTable($id)}, {$this->getSQLTable($id . '_perms')};")
-                    ->execute();
-            }
+            //            if (!($e instanceof DuplicateException)) {
+            //                $this->getPDO()
+            //                    ->prepare("DROP TABLE IF EXISTS {$this->getSQLTable($id)}, {$this->getSQLTable($id . '_perms')};")
+            //                    ->execute();
+            //            }
```
abnegate, line 222:
```suggestion
```

abnegate (commented):
This should never happen unless `getCollection` incorrectly returns an empty document in the `Database` wrapping method :thinking: maybe because of replication lag?

PR #493, opened 2024-12-06
Respect timestamps in bulk operations and add tests

PineappleIOnic:
Bulk operations didn't previously respect timestamps, this fixes that and adds tests for that case.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -4084,6 +4084,12 @@ class Database
                         $this->silent(fn () => $this->updateDocumentRelationships($collection, $document, $newDocument));
                         $documents[] = $newDocument;
                     }
+
+                    // Check if document was updated after the request timestamp
+                    $oldUpdatedAt = new \DateTime($document->getUpdatedAt());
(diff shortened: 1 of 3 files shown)


src/Database/Database.php
```diff
@@ -4084,6 +4084,12 @@ public function updateDocuments(string $collection, Document $updates, array $qu
                         $this->silent(fn () => $this->updateDocumentRelationships($collection, $document, $newDocument));
                         $documents[] = $newDocument;
                     }
+
+                    // Check if document was updated after the request timestamp
+                    $oldUpdatedAt = new \DateTime($document->getUpdatedAt());
```
abnegate, line 4089: Let's try catch here like we do for deletes

tests/e2e/Adapter/Base.php
```diff
@@ -2110,6 +2110,8 @@ public function testCreateDocuments(): array
             $this->assertEquals(9223372036854775807, $document->getAttribute('bigint'));
         }
 
+        // Test (FAIL) create documents with invalid timestamps
+
```
abnegate, line 2114: Let's remove this comment if it's not relevant here

abnegate (approved):

PR #494, opened 2024-12-17
Add cache fallback into DB library

PineappleIOnic:
This PR introduces exception catches for when a DB cache fails and makes it so we automatically falls back to not using the cache, this PR currently uses `Console::warning` as a way to emit these failure mainly because getting rid of them entirely could result in performance degradation not being noticed because of caches being down.

Tests have been added to check these changes work and checks will also be added into Appwrite

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -901,6 +901,13 @@ abstract class Adapter
     abstract public function getSupportForUpserts(): bool;
 
     /**
+     * Is Cache Fallback supported?
+     *
+     * @return bool
+     */
+    abstract public function getSupportForCacheFallback(): bool;
+
+    /**
      * Get current attribute count from collection document
      *
      * @param Document $collection
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -869,6 +869,13 @@ abstract class Adapter
     abstract public function getSupportForGetConnectionId(): bool;
 
     /**
+     * Is cast index as array supported?
+     *
+     * @return bool
+     */
+    abstract public function getSupportForCastIndexArray(): bool;
+
+    /**
      * Get current attribute count from collection document
      *
      * @param Document $collection
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -446,7 +446,7 @@ abstract class Adapter
     abstract public function deleteCollection(string $id): bool;
 
     /**
-     * Analyze a collection updating it's metadata on the database engine
+     * Analyze a collection updating its metadata on the database engine
      *
      * @param string $collection
      * @return bool
@@ -638,6 +638,24 @@ abstract class Adapter
     abstract public function updateDocuments(string $collection, Document $updates, array $documents): int;
 
     /**
+     * Create documents if they do not exist, otherwise update them.
+     *
+     * If attribute is not empty, only the specified attribute will be increased, by the new value in each document.
+     *
+     * @param string $collection
+     * @param string $attribute
+     * @param array<Document> $documents
+     * @param int $batchSize
+     * @return array<Document>
+     */
+    abstract public function createOrUpdateDocuments(
+        string $collection,
+        string $attribute,
+        array $documents,
+        int $batchSize
+    ): array;
+
+    /**
      * Delete Document
      *
      * @param string $collection
@@ -700,7 +718,7 @@ abstract class Adapter
     abstract public function count(string $collection, array $queries = [], ?int $max = null): int;
 
     /**
-     * Get Collection Size
+     * Get Collection Size of the raw data
      *
      * @param string $collection
      * @return int
@@ -876,6 +894,13 @@ abstract class Adapter
     abstract public function getSupportForCastIndexArray(): bool;
 
     /**
+     * Is upserting supported?
+     *
+     * @return bool
+     */
+    abstract public function getSupportForUpserts(): bool;
+
+    /**
      * Get current attribute count from collection document
      *
      * @param Document $collection
@@ -972,7 +997,7 @@ abstract class Adapter
      */
     public function filter(string $value): string
     {
-        $value = \preg_replace("/[^A-Za-z0-9\_\-]/", '', $value);
+        $value = \preg_replace("/[^A-Za-z0-9_\-]/", '', $value);
 
         if (\is_null($value)) {
             throw new DatabaseException('Failed to filter key');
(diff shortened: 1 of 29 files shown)


src/Database/Database.php
```diff
@@ -2956,7 +2957,14 @@ public function getDocument(string $collection, string $id, array $queries = [],
             $documentCacheHash .= ':' . \md5(\implode($selections));
         }
 
-        if ($cache = $this->cache->load($documentCacheKey, self::TTL, $documentCacheHash)) {
+        try {
```
fogelito, line 2960: Should we do these checks in the Adapter?

PineappleIOnic: I'm unsure if we could move this over to the cache library. Still, I'm concerned about the far-reaching impact of implementing auto-failover for everything that touches the cache library.

Another thing I'm concerned about is cache desync. Let's say we delete a file, and the cache is down for that delete, then it comes back up. Would we return the deleted file? How long would it take for the cache to become invalidated 🤔

abnegate: We'll only skip the cache if it's down on reads :+1:

src/Database/Database.php
```diff
@@ -3951,6 +3973,9 @@ public function updateDocument(string $collection, string $id, Document $documen
 
             $this->adapter->updateDocument($collection->getId(), $id, $document);
 
+            $this->purgeRelatedDocuments($collection, $id);
+            $this->purgeCachedDocument($collection->getId(), $id);
```
abnegate, line 3977: Why do we move this?

tests/e2e/Adapter/MariaDBTest.php
```diff
@@ -41,10 +42,14 @@ public static function getDatabase(bool $fresh = false): Database
...
-        $redis->flushAll();
-        $cache = new Cache(new RedisAdapter($redis));
+        try {
+            $redis = new Redis();
+            $redis->connect('redis', 6379);
+            $redis->flushAll();
+            $cache = new Cache(new RedisAdapter($redis));
+        } catch (\Exception $e) {
+            $cache = new Cache(new None());
+        }
```
abnegate, line 52: Why do we do this?

tests/e2e/Adapter/Base.php
```diff
@@ -70,6 +71,10 @@ public function setUp(): void
     public function tearDown(): void
     {
         Authorization::setDefaultStatus(true);
+
+        $stdout = '';
+        $stderr = '';
+        Console::execute('docker ps -a --filter "name=utopia-redis" --format "{{.Names}}" | xargs -r docker start', "", $stdout, $stderr);
```
abnegate, line 77: Let's do this at the end of the single test, no need to do it after every test

src/Database/Database.php
PineappleIOnic: So if the cache fails it'll rollback the transaction

tests/e2e/Adapter/MariaDBTest.php
PineappleIOnic: It was a left over from when I didn't have automatic redis reinitialization

abnegate (commented):
We need to re-target to main and get the tests passing

src/Database/Adapter/SQLite.php
```diff
@@ -224,7 +224,7 @@ public function createCollection(string $name, array $attributes = [], array $in
 
             if (!($e instanceof Duplicate)) {
                 $this->getPDO()
-                    ->prepare("DROP TABLE IF EXISTS {$this->getSQLTable($id)}, {$this->getSQLTable($id . '_perms')};")
+                    ->prepare("DROP TABLE IF EXISTS {$this->getSQLTable($id)}; DROP TABLE IF EXISTS {$this->getSQLTable($id . '_perms')};")
```
abnegate, line 227: Why do we change this?

src/Database/Adapter.php
```diff
@@ -900,6 +900,13 @@ abstract public function getSupportForCastIndexArray(): bool;
      */
     abstract public function getSupportForUpserts(): bool;
 
+    /**
+     * Is Cache Fallback supported?
+     *
+     * @return bool
+     */
+    abstract public function getSupportForCacheFallback(): bool;
```
abnegate, line 908: Let's use a better name here, `getSupportForCacheSkipOnFailure`, it's verbose but definitely portrays what it means

src/Database/Database.php
```diff
@@ -457,6 +458,20 @@ public function on(string $event, string $name, callable $callback): static
...
+     */
+    public function flushListeners(): static
+    {
+        $this->listeners = [
+            '*' => [],
+        ];
+
+        return $this;
+    }
+
```
abnegate, line 474: Let's remove this, test code belongs in tests. Pretty sure we can do the same by setting * listener to null

src/Database/Database.php
```diff
@@ -2992,7 +3007,14 @@ public function getDocument(string $collection, string $id, array $queries = [],
             $documentCacheHash .= ':' . \md5(\implode($selections));
         }
 
-        if ($cache = $this->cache->load($documentCacheKey, self::TTL, $documentCacheHash)) {
+        try {
+            $cache = $this->cache->load($documentCacheKey, self::TTL, $documentCacheHash);
```
abnegate, line 3011: Let's call this `$cached` instead

src/Database/Database.php
```diff
@@ -3040,10 +3062,50 @@ public function getDocument(string $collection, string $id, array $queries = [],
...
+                Console::warning('Failed to load document from cache: ' . $e->getMessage());
+                $cache = [];
+            }
+            if (empty($cache)) {
+                $cache = [];
+            }
+            if (!\in_array($v, $cache)) {
+                $cache[] = $v;
+                try {
+                    $this->cache->save($ck, $cache, $ck);
```
abnegate, line 3093: This shouldn't be added back, sync with main or remove manually

src/Database/Adapter/SQLite.php
PineappleIOnic: I tested the code and found that commas are not supported in SQL, so I fixed it to use multiple statements in one.

abnegate: Commas are valid like that in drop table, but we actually need to remove this if check altogether, and just throw $this->processException($e)

src/Database/Database.php
```diff
@@ -3040,10 +3054,20 @@ public function getDocument(string $collection, string $id, array $queries = [],
                 $attribute['type'] === Database::VAR_RELATIONSHIP
         );
 
+        $relationships = \array_filter(
+            $collection->getAttribute('attributes', []),
+            fn ($attribute) =>
+                $attribute['type'] === Database::VAR_RELATIONSHIP
+        );
+
```
abnegate, line 3062: This is duplicate

tests/e2e/Adapter/Base.php
```diff
@@ -17683,6 +17684,76 @@ public function testEvents(): void
             $database->deleteAttribute($collectionId, 'attr1');
             $database->deleteCollection($collectionId);
             $database->delete('hellodb');
+
+            // Remove all listeners
+            $database->on(Database::EVENT_ALL, 'test', null);
```
abnegate, line 17689: I see some other names too like `should-not-execute`

abnegate (approved):

PR #512, opened 2025-01-22
Feat batch upsert

abnegate:
Add three new methods:

- createOrUpdateDocuments(collection, documents, [batchSize])
- createOrUpdateDocumentWithInplaceIncrease(collection, attribute, documents, [batchSize])
- createOrUpdateDocumentWithIncrease(collection, attribute, value, documents, [batchSize])

To decrement, you can pass a negative value.


fogelito (approved):

tests/e2e/Adapter/Base.php
```diff
@@ -2230,6 +2230,310 @@ public function testCreateDocuments(): array
...
+                    Permission::read(Role::any()),
+                    Permission::create(Role::any()),
+                    Permission::update(Role::any()),
+                    Permission::delete(Role::any()),
+                ],
+            ]),
+            new Document([
+                '$id' => 'second',
+                'string' => 'text📝',
+                'integer' => 5,
```
shimonewman, line 2432: would be great to change the 'integer' value to a different value than 5  in order to understand 
that we increment by the value that we pass and not by the value stored in 'integer'

abnegate: Done

PR #513, opened 2025-01-22
Fix Document getid type

fogelito:


src/Database/Document.php
```diff
@@ -26,8 +27,12 @@ class Document extends ArrayObject
      */
     public function __construct(array $input = [])
     {
+        if (isset($input['$id']) && !\is_string($input['$id'])) {
+            throw new StructureException('$id must be of type string');
+        }
+
         if (isset($input['$permissions']) && !is_array($input['$permissions'])) {
-            throw new DatabaseException('$permissions must be of type array');
+            throw new StructureException('$permissions must be of type array');
```
fogelito, line 35: I changed this so we do not get 500

abnegate (approved):

PR #515, opened 2025-01-31
Implement Count Queries

PineappleIOnic:
This PR implements count queries and generally cleans up the code of how aggregates work so it's more maintainable. This supports all DB adapters excluding MongoDB.

### DIFF
--- a/tests/e2e/Adapter/Base.php
+++ b/tests/e2e/Adapter/Base.php
@@ -17525,7 +17525,7 @@ abstract class Base extends TestCase
 
     public function testSumQueries(): void
     {
-        if (!static::getDatabase()->getAdapter()->getSupportForSum()) {
+        if (!static::getDatabase()->getAdapter()->getSupportForAggregateQueries()) {
             $this->expectNotToPerformAssertions();
             return;
         }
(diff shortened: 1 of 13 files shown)


tests/e2e/Adapter/Base.php
```diff
@@ -17679,6 +17679,134 @@ public function testSumQueries(): void
...
+            ]);
+            $this->fail('Failed to throw exception');
+        } catch (QueryException $e) {
+            $this->assertEquals('Invalid query: Attribute not found in schema: nonExistent', $e->getMessage());
+        }
+
+        // Test mixing count with other queries
+        $documents = $database->find('testCountQueries', [
+            Query::count('integer'),
+            Query::limit(1),
```
fogelito, line 17764: Perhaps should validate not using limits / offset queries Since they happen after the aggregations.
Most likely count/sum will return 1 row.

tests/e2e/Adapter/Base.php
```diff
@@ -17679,6 +17679,134 @@ public function testSumQueries(): void
...
+        ]);
+
+        $this->assertCount(1, $documents);
+        $this->assertEquals(1, $documents[0]->getAttribute('integer'));
+
+        $documents = $database->find('testCountQueries', [
+            Query::count('integer'),
```
fogelito, line 17771: - Do we allow select queries when using aggregation queries, which will lead to errors?
- Do we allow Query::count('integer'), Query::sum('integer') will both have the alias?

fogelito: - Do we validate aggregation queries in nested queries?

tests/e2e/Adapter/Base.php
PineappleIOnic: We run aggregate SQL after the main SQL using a subquery so this shouldn't be a problem

tests/e2e/Adapter/Base.php
PineappleIOnic: 1. No we do not allow select queries when using aggregation queries, we have a validator for that.
2. We only have the sum query, we could add an alias if we feel we need it
3. I haven't touched nested queries, so I don't think so

PR #521, opened 2025-02-06
Internal ids chunks

fogelito:


src/Database/Adapter/MariaDB.php
```diff
@@ -990,15 +990,19 @@ public function createDocuments(string $collection, array $documents, int $batch
...
+                    /**
+                     * @var Document $document
+                     */
                     $attributes = $document->getAttributes();
                     $attributes['_uid'] = $document->getId();
                     $attributes['_createdAt'] = $document->getCreatedAt();
                     $attributes['_updatedAt'] = $document->getUpdatedAt();
                     $attributes['_permissions'] = \json_encode($document->getPermissions());
 
                     if (!empty($document->getInternalId())) {
```
fogelito, line 1002: We can not insert internal ids since the attribute list is not dynamic, or all document must have internal id or all must be empty ,we will get invalid attribute count.

abnegate: I think we need this behavior for migrations, can we keep it, but throw an exception if only some documents $internalId?

fogelito: We need this check for all attributes not only internal id. 
Added logic for attributes count and  a test

src/Database/Adapter/MariaDB.php
```diff
@@ -965,129 +965,158 @@ public function createDocument(string $collection, Document $document): Document
...
+
+    /**
+     * Translate Uid's to internal id's
+     *
+     * @param string $collection
+     * @param array<string> $documentIds
+     * @return array<string>
+     * @throws DatabaseException
+     */
+    private function translateUidsToInternalIds(string $collection, array $documentIds): array
```
abnegate, line 1102:
```suggestion
    private function getInternalIds(string $collection, array $documentIds): array
```

src/Database/Adapter/MariaDB.php
```diff
@@ -965,129 +965,158 @@ public function createDocument(string $collection, Document $document): Document
...
-                        $stmtPermissions->bindValue(':_tenant', $this->tenant);
-                    }
+                if ($this->sharedTables) {
+                    $stmtPermissions->bindValue(':_tenant', $this->tenant);
+                }
+
+                $stmtPermissions?->execute();
+            }
+
+            $internalIds = $this->translateUidsToInternalIds($collection, $documentIds);
```
abnegate, line 1080:
```suggestion
            $internalIds = $this->getInternalIds($collection, $documentIds);
```

src/Database/Adapter/MariaDB.php
```diff
@@ -1612,218 +1632,195 @@ public function createOrUpdateDocuments(
...
+                    $stmtAddPermissions->bindValue(':_tenant', $this->tenant);
+                }
+
+                $stmtAddPermissions->execute();
             }
 
-            $stmt->execute();
-            $internalIds = $stmt->fetchAll(PDO::FETCH_KEY_PAIR); // Fetch as [documentId => internalId]
-            $stmt->closeCursor();
+            $internalIds = $this->translateUidsToInternalIds($collection, $documentIds);
```
abnegate, line 1823:
```suggestion
            $internalIds = $this->getInternalIds($collection, $documentIds);
```

src/Database/Database.php
```diff
@@ -3431,6 +3444,12 @@ public function createDocuments(
                 ->setAttribute('$createdAt', empty($createdAt) || !$this->preserveDates ? $time : $createdAt)
                 ->setAttribute('$updatedAt', empty($updatedAt) || !$this->preserveDates ? $time : $updatedAt);
 
+            if (is_null($columnCount)) {
+                $columnCount = count($document);
+            } elseif ($columnCount != count($document)) {
+                throw new StructureException('Insert value list does not match column list');
+            }
+
```
abnegate, line 3452: We are going to handle this with a different approach and allow count mismatches, but just defaulting to null for any values that are missing. This is what we do now, and the decode function takes care of filling in missing optional values with their defaults. See this PR: https://github.com/utopia-php/database/pull/524

```suggestion
```

src/Database/Database.php
```diff
@@ -3419,7 +3419,20 @@ public function createDocuments(
...
+         */
+        if ($collection->getId() !== self::METADATA) {
+            $authorization = new Authorization(self::PERMISSION_CREATE);
+            if (!$authorization->isValid($collection->getCreate())) {
+                throw new AuthorizationException($authorization->getDescription());
+            }
+        }
+
         $time = DateTime::now();
+        $columnCount = null;
```
abnegate, line 3435:
```suggestion
```

src/Database/Adapter/MariaDB.php
```diff
@@ -965,129 +965,158 @@ public function createDocument(string $collection, Document $document): Document
...
             }
+        } catch (PDOException $e) {
+            throw $this->processException($e);
+        }
 
+        return $documents;
+    }
+
+    /**
+     * Translate Uid's to internal id's
```
abnegate, line 1095:
```suggestion
     * Get internal IDs for the given documents
```

abnegate (approved):

PR #523, opened 2025-02-10
feat: add ScyllaDB adapter support

tinkerer-shubh:
# Add ScyllaDB Adapter Support

related issue : [#9295](https://github.com/appwrite/appwrite/issues/9295)

This PR adds initial support for ScyllaDB as a new database adapter in the Utopia Database library.

## Features
The ScyllaDB adapter implements core database operations:
- Database management (create, delete, list)
- Collection operations (create, delete, list)
- Document CRUD operations
- Attribute management
- Index support (key and unique)
- Query operations with support for all standard query types
- Schema validation

## Implementation Details
- Follows the same pattern as existing adapters
- Leverages ScyllaDB's native CQL capabilities
- Includes comprehensive test suite
- Adds ScyllaDB services to docker-compose.yml for testing
- Updates README with connection example and specifications

## Testing
- Full test suite in `tests/e2e/Adapter/ScyllaDBTest.php`
- All tests passing
- Can be tested using `docker-compose up scylladb`

## Documentation
Added to README:
- ScyllaDB in supported databases table
- Connection example
- Database specifications and limits

## Specifications
- ID max size: 255 bytes
- Document: Unrestricted size
- Collection: Unrestricted attributes
- Indexes: Unrestricted
- String max size: Unrestricted
- Integer max size: 2^63 - 1

### DIFF
--- a/src/Database/Adapter/ScyllaDB.php
+++ b/src/Database/Adapter/ScyllaDB.php
@@ -1,0 +1,971 @@
+<?php
+
+namespace Utopia\Database\Adapter;
+
+use Exception;
+use PDO;
+use PDOException;
+use Utopia\Database\Database;
+use Utopia\Database\Document;
+use Utopia\Database\Exception as DatabaseException;
+use Utopia\Database\Exception\Duplicate as DuplicateException;
+use Utopia\Database\Exception\NotFound as NotFoundException;
+use Utopia\Database\Exception\Timeout as TimeoutException;
+use Utopia\Database\Exception\Truncate as TruncateException;
+use Utopia\Database\Query;
+use Utopia\Database\Validator\Authorization;
+
+class ScyllaDB extends SQL
+{
+    /**
+     * Create Database
+     *
+     * @param string $name
+     * @return bool
+     * @throws Exception
+     * @throws PDOException
+     */
+    public function create(string $name): bool
+    {
+        $name = $this->filter($name);
+
+        if ($this->exists($name)) {
+            return true;
+        }
+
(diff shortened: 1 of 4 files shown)


src/Database/Adapter/ScyllaDB.php
```diff
@@ -0,0 +1,971 @@
...
+     */
+    public function create(string $name): bool
+    {
+        $name = $this->filter($name);
+
+        if ($this->exists($name)) {
+            return true;
+        }
+
+        $sql = "CREATE KEYSPACE {$name} WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3}";
```
DH-555, line 36: I think that we should set the replication factor to 1 since by default we will not have a cluster?

Or maybe set some kind of env variable to customise it and the strategy

tinkerer-shubh: Good point! The replication factor is set to 3 as a default for fault tolerance in case we scale to a cluster later. But you’re right if we’re not running a cluster yet, 1 makes more sense for now.

ChiragAgg5k:
amazing job @tinkerer-shubh! 🔥

tinkerer-shubh:
> amazing job @tinkerer-shubh! 🔥

Thank you! This is far from perfect though. The initial goal was to achieve an MVP of sorts and then iterate further based on the feedback. I see the checks failing too, will take a look. :)

I was also thinking that maybe it's best to mark this as a draft for now while it's still in its iteration phase?

PR #524, opened 2025-02-12
Fix different column for nulls bug in create documents

PineappleIOnic:
Create documents constructs the column list based on all documents however there is a chance that values don't exist on a document (because of it not being required) meaning it doesn't get set in the bindValues resulting in a diff between the number of values expected and the number of values given.

This PR fixes that by making a column list of all the documents and checking against that when binding values, if a value doesn't exist it will be set as null within the bindValues. Validators from the database adapter should catch if a value is required or doesn't exist at all so we don't need to check that.

Tests have been added to make sure that this fixes the bug explained.

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -988,6 +988,10 @@ class MariaDB extends SQL
             }
             $attributeKeys = array_unique($attributeKeys);
 
+            if ($this->sharedTables) {
+                $attributeKeys[] = '_tenant';
+            }
+
             $batches = \array_chunk($documents, \max(1, $batchSize));
             $documentIds = \array_map(fn ($document) => $document->getId(), $documents);
 
@@ -1013,6 +1017,7 @@ class MariaDB extends SQL
                     if (!empty($document->getInternalId())) {
                         $internalIds[$document->getId()] = true;
                         $attributes['_id'] = $document->getInternalId();
+                        $attributeKeys[] = '_id';
                     }
 
                     if ($this->sharedTables) {
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -982,9 +982,17 @@ class MariaDB extends SQL
             $name = $this->filter($collection);
 
             $attributeKeys = Database::INTERNAL_ATTRIBUTE_KEYS;
+
+            $hasInternalId = null;
             foreach ($documents as $document) {
                 $attributes = $document->getAttributes();
                 $attributeKeys = array_merge($attributeKeys, array_keys($attributes));
+
+                if ($hasInternalId === null) {
+                    $hasInternalId = !empty($document->getInternalId());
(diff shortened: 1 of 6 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -980,9 +980,28 @@ public function createDocuments(string $collection, array $documents, int $batch
 
         try {
             $name = $this->filter($collection);
+
+            $attributeKeys = [
+                '_uid',
+                '_createdAt',
+                '_updatedAt',
+                '_permissions',
+            ];
```
abnegate, line 989: There are some const we can use here, INTERNAL_ATTRIBUTES or something

src/Database/Adapter/MariaDB.php
```diff
@@ -1005,16 +1024,10 @@ public function createDocuments(string $collection, array $documents, int $batch
...
-                        $columns[$key] = "`{$this->filter($attribute)}`";
-                    }
-
-                    $columns = '(' . \implode(', ', $columns) . ')';
-
                     $bindKeys = [];
 
-                    foreach ($attributes as $value) {
+                    foreach ($attributeKeys as $key) {
+                        $value = $attributes[$key] ?? null;
```
abnegate, line 1030: Not sure if this should be null or the default value for the attribute, what's the behavior for a single document create if optional attributes are missing?

src/Database/Adapter/MariaDB.php
```diff
@@ -980,6 +980,31 @@ public function createDocuments(string $collection, array $documents): array
...
+            foreach ($documents as $document) {
+                $attributes = $document->getAttributes();
+                $attributeKeys = array_merge($attributeKeys, array_keys($attributes));
+
+                if ($hasInternalId === null) {
+                    $hasInternalId = !empty($document->getInternalId());
+                } elseif ($hasInternalId == empty($document->getInternalId())) {
+                    throw new DatabaseException('All documents must have an internalId if one is set');
+                }
+            }
```
abnegate, line 995:
```suggestion
            $expectInternalId = !empty($documents[0]->getInternalId());

            foreach ($documents as $document) {
                $attributeKeys = array_merge($attributeKeys, array_keys($document->getAttributes()));

                if ($expectInternalId !== !empty($document->getInternalId())) {
                    throw new DatabaseException('All documents must have an internalId if one is set');
                }
            }
```

tests/e2e/Adapter/Base.php
```diff
@@ -2238,6 +2238,72 @@ public function testCreateDocuments(): array
...
+            new Document([
+                '$id' => 'third',
+                '$internalId' => 'third',
+                'string' => 'text📝',
+            ]),
+            new Document([
+                '$id' => 'fourth',
+                'string' => 'text📝',
+            ]),
+        ];
```
abnegate, line 2289: Let's add another check for the other way: first document doesn't have internal ID, second does

abnegate (approved):

PR #526, opened 2025-02-12
Check last insert

fogelito:


src/Database/Adapter/MariaDB.php
```diff
@@ -950,6 +950,10 @@ public function createDocument(string $collection, Document $document): Document
 
             $document['$internalId'] = $this->getDocument($collection, $document->getId(), [Query::select(['$internalId'])])->getInternalId();
 
+            if (empty($document['$internalId'])) {
+                throw new DatabaseException('Error creating document empty "$internalId"');
```
stnguyen90, line 954: We need to make sure not to propagate this back to API callers

abnegate: We don't catch this by default, from the API it will be a 500 and users will see "Server error", which is correct for this case

PR #535, opened 2025-03-11
Move auth outside transaction

fogelito:
- Prevent starting a transaction if authorization is not permitted.
- Fix array merge , since we use first Cursor/Limit in array of queries

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -5395,7 +5395,7 @@ class Database
             throw new DatabaseException("Cursor document must be from the same Collection.");
         }
 
-        $documents = $this->withTransaction(function () use ($collection, $queries, $batchSize, $limit, $cursor) {
+        $documents = $this->withTransaction(function () use ($collection, $queries, $batchSize, $limit, $cursor, $skipAuth, $authorization) {
             $documents = [];
             $originalLimit = $limit;
             $lastDocument = $cursor;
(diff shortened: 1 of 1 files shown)


src/Database/Database.php
```diff
@@ -5454,7 +5461,7 @@ public function deleteDocuments(string $collection, array $queries = [], int $ba
             }
 
             foreach (\array_chunk($documents, $batchSize) as $chunk) {
-                $this->adapter->deleteDocuments(
+                $this->adapter->deleteDocuments( // Do we need to check $skipAuth like in updateDocuments?
```
fogelito, line 5464: Do we need to check $skipAuth like in updateDocuments?

src/Database/Database.php
```diff
@@ -4121,15 +4120,19 @@ public function updateDocuments(string $collection, Document $updates, array $qu
...
+
+                if (! empty($lastDocument)) {
+                    $new[] = Query::cursorAfter($lastDocument);
+                }
+
+                $affectedDocuments = $this->silent(fn () => $this->find(
+                    $collection->getId(),
+                    array_merge($new, $queries),
+                    forPermission: Database::PERMISSION_UPDATE
+                ));
```
abnegate, line 4135: Let's revert this since we already use the user limit

src/Database/Database.php
```diff
@@ -5404,15 +5407,19 @@ public function deleteDocuments(string $collection, array $queries = [], int $ba
...
+
+                if (! empty($lastDocument)) {
+                    $new[] = Query::cursorAfter($lastDocument);
+                }
+
+                $affectedDocuments = $this->silent(fn () => $this->find(
+                    $collection->getId(),
+                    array_merge($new, $queries),
+                    forPermission: Database::PERMISSION_DELETE
+                ));
```
abnegate, line 5422: Revert here too

src/Database/Database.php
abnegate: Yep we should

fogelito: Done

abnegate (approved):

PR #538, opened 2025-03-16
Remove Mongo Adapter

fogelito:
Since 

Query::orderDesc('')
Is the same as 
Query::orderDesc('$internalId')


src/Database/Query.php
```diff
@@ -86,6 +86,10 @@ class Query
      */
     public function __construct(string $method, string $attribute = '', array $values = [])
     {
+        if ($attribute === '' && \in_array($method, [Query::TYPE_ORDER_ASC, Query::TYPE_ORDER_DESC])) {
+            $attribute = '$internalId';
```
fogelito, line 90: Thinking of a way for dynamic fallback per adapter

fogelito: Is attribute === '' In Mongo means a different order by?

fogelito: Seems like a bug

abnegate, line 90: Instead of this, let's set it as the attribute parameter's default on the orderDesc and orderAsc methods

fogelito: We can't,  since Query::parse Will skip it by creating new Self(), it goes straight to Constructor.

abnegate (approved):

PR #541, opened 2025-03-24
fix: purgeCachedDocument after transaction

loks0n:
This is kind of a big conceptual change, previously we attempted to be strongly consistent with cache and DB (but imperfect due to scenario above), now we are eventually consistent. There is now a gap between purgeCachedDocument and the commit, where gets can be stale


abnegate:
Addressed elsewhere

PR #542, opened 2025-03-24
Feat multi tenant insert

abnegate:
Add function to allow inserting documents with the tenant set in each document instead of set on the database instance.


src/Database/Adapter/MariaDB.php
```diff
@@ -942,13 +941,13 @@ public function createDocument(string $collection, Document $document): Document
...
                 if ($this->sharedTables) {
-                    $stmtPermissions->bindValue(':_tenant', $this->tenant);
+                    $stmtPermissions->bindValue(':_tenant', $document->getTenant());
                 }
             }
 
             $stmt->execute();
 
-            $document['$internalId'] = $this->getDocument($collection, $document->getId(), [Query::select(['$internalId'])])->getInternalId();
+            $document['$internalId'] = $this->pdo->lastInsertId();
```
fogelito, line 951: Hmm not sure this worked in Appwrite, if does not work 
We can do this
$statment = $this->getPDO()->prepare("select last_insert_id() as id");
$statment->execute();
$last = $statment->fetch();
$document['$internalId'] = $last['id'];

abnegate: What do you mean? What causes the issue from Appwrite?

abnegate: Interesting, will pull this branch in and run tests

abnegate: Looks okay in Appwrite tests :+1:

tests/e2e/Adapter/Base.php
```diff
@@ -16438,6 +16438,95 @@ public function testSharedTablesDuplicates(): void
...
+
+        $database->createAttribute(__FUNCTION__, 'name', Database::VAR_STRING, 10, false);
+        $database->createIndex(__FUNCTION__, 'nameIndex', Database::INDEX_KEY, ['name']);
+
+        $doc1Id = ID::unique();
+
+        // Create doc for tenant 1
+        $database
+            ->setTenant(null)
+            ->setTenantPerDocument(true)
```
fogelito, line 16478: Cache-wise will it be ok , since _tenant is a part of the cache key 🤔

src/Database/Database.php
```diff
@@ -3372,7 +3403,16 @@ public function createDocument(string $collection, Document $document): Document
             ->setAttribute('$updatedAt', empty($updatedAt) || !$this->preserveDates ? $time : $updatedAt);
 
         if ($this->adapter->getSharedTables()) {
-            $document['$tenant'] = (string)$this->adapter->getTenant();
```
fogelito, line 3375: Do we need to check in case of getTenantPerDocument = true
if $document->getTenant has write permissions to that collection?

tests/e2e/Adapter/Base.php
abnegate: It's okay because we are only inserting, to read you still need the correct tenant set on the instance

src/Database/Database.php
abnegate: Tenant and permissions don't care about each other, if the _user_ has write permissions on the collection, they can write to it with any tenant

fogelito (approved):

PR #545, opened 2025-03-26
Delete documents & update documents chunks

fogelito:


src/Database/Adapter/MariaDB.php
```diff
@@ -1965,49 +1966,51 @@ public function deleteDocuments(string $collection, array $ids): int
...
+                WHERE _document IN (" . \implode(', ', \array_map(fn ($index) => ":_id_{$index}", \array_keys($permissionIds))) . ")
             ";
 
-            if ($this->sharedTables) {
-                $sql .= ' AND _tenant = :_tenant';
-            }
-
-            $sql = $this->trigger(Database::EVENT_PERMISSIONS_DELETE, $sql);
+                if ($this->sharedTables) {
+                    $sql .= ' AND _tenant = :_tenant';
```
abnegate, line 1996: Can we use the `getTenantQuery` method here?

src/Database/Database.php
```diff
@@ -5495,6 +5495,9 @@ public function deleteDocuments(string $collection, array $queries = [], int $ba
                 $new[] = Query::cursorAfter($lastDocument);
             }
 
+            /**
+             * @var array<Document> $affectedDocuments
+             */
```
abnegate, line 5500: We don't need this since find declares it returns array<Document>, and silent declares it returns whatever it's callback returns

src/Database/Database.php
```diff
@@ -5505,9 +5508,15 @@ public function deleteDocuments(string $collection, array $queries = [], int $ba
                 break;
             }
 
-            $documents = \array_merge($affectedDocuments, $documents);
-
+            $internalId = [];
+            $permIds = [];
```
abnegate, line 5512: Let's call these `$internalIds` and `$permissionIds`

src/Database/Adapter/MariaDB.php
fogelito: Done!

src/Database/Database.php
fogelito: Done!

src/Database/Database.php
abnegate: We can keep it if its helps 👌

abnegate (approved):

PR #546, opened 2025-03-26
Feat pool adapter

abnegate:


src/Database/Database.php
```diff
@@ -1033,48 +1033,8 @@ public function getAdapter(): Adapter
...
      *
-     * If a transaction is already active, this will only increment the transaction count and return true.
-     *
-     * @return bool
-     * @throws DatabaseException
-     */
-    public function startTransaction(): bool
-    {
-        return $this->adapter->startTransaction();
-    }
```
abnegate, line 1046: Methods removed because only `withTransaction` should be used from outside the library

fogelito (approved):

PR #551, opened 2025-04-07
Feat doc per tenant upsert

abnegate:
- Allow using `setTenantPerDocument` with `createOrUpdateDocument` functions.


src/Database/Database.php
```diff
@@ -4648,17 +4662,32 @@ public function createOrUpdateDocumentsWithIncrease(
...
-                $document->getId(),
-                [Query::select(['$internalId', '$permissions'])],
-                forUpdate: true
-            )));
+            if ($this->getSharedTables() && $this->getTenantPerDocument()) {
+                $old = Authorization::skip(fn () => $this->withTenant($document->getTenant(), fn () => $this->silent(fn () => $this->getDocument(
+                    $collection->getId(),
+                    $document->getId(),
+                    [Query::select($selects)],
+                    forUpdate: true
```
fogelito, line 4677: The forUpdate is outside the Transaction, not sure will do the job  here?

src/Database/Adapter/MariaDB.php
```diff
@@ -1466,18 +1446,20 @@ public function createOrUpdateDocuments(
             $sql = "
                 SELECT _document, _type, _permission
                 FROM {$this->getSQLTable($name . '_perms')}
-                WHERE _document IN (" . \implode(',', \array_map(fn ($index) => ":_key_{$index}", \array_keys($documentIds))) . ")
-                {$this->getTenantQuery($collection)}
+                WHERE _document IN (" . \implode(',', \array_map(fn ($index) => ":_key_{$index}", \array_keys($documents))) . ")
```
fogelito, line 1449: - Can we skip this query and take  the permissions data from $old Document , which we read in Databse.php?

- Will need to check this queries , since will have 2 in , might be good since _document is first part of the index?

abnegate: Let's address separately - will take a fair amount of time, since at the moment it's not aware of which documents already exist vs. which one are newly inserted

fogelito: ok ..
Can we hardcoded skip it if it is stats collection?

src/Database/Adapter/MariaDB.php
```diff
@@ -1009,11 +1010,12 @@ public function createDocuments(string $collection, array $documents): array
                 $attributes['_updatedAt'] = $document->getUpdatedAt();
                 $attributes['_permissions'] = \json_encode($document->getPermissions());
 
-                if (! empty($document->getInternalId())) {
+                if (!empty($document->getInternalId())) {
                     $attributes['_id'] = $document->getInternalId();
                     $attributeKeys[] = '_id';
                 } else {
                     $documentIds[] = $document->getId();
+                    $documentTenants[] = $document->getTenant();
```
fogelito, line 1018: Why $documentTenants[] is in the else block? 
Can non shared tables have internal ids?

src/Database/Adapter/MariaDB.php
```diff
@@ -1547,38 +1506,27 @@ public function createOrUpdateDocuments(
...
-                if ($this->sharedTables) {
-                    $stmtAddPermissions->bindValue(':_tenant', $this->tenant);
-                }
-
                 $stmtAddPermissions->execute();
             }
 
-            $internalIds = $this->getInternalIds($collection, $documentIds);
-
+            $internalIds = $this->getInternalIds($collection, $documentIds, $documentTenants);
```
fogelito, line 1526: getInternalIds should be called from outside the transaction, It should not hold the locks since it is a helper query.

src/Database/Adapter/MariaDB.php
abnegate: Skipped the query, passing through `old` for each doc so we can check what exists already

src/Database/Adapter/MariaDB.php
abnegate: It's inside transaction scope for multiple other functions too, let's address separately

src/Database/Database.php
```diff
@@ -4702,39 +4734,55 @@ public function createOrUpdateDocumentsWithIncrease(
...
+             * @var array<Change> $chunk
+             */
+            \array_push(
+                $stack,
+                ...$this->withTransaction(fn () => Authorization::skip(fn () => $this->adapter->createOrUpdateDocuments(
+                    $collection->getId(),
+                    $attribute,
+                    $chunk
+                )))
+            );
```
fogelito, line 4756: Would we  like us to clean cache for the $chunk documents? 
To be safe in case of failure?

abnegate: Yep makes sense

fogelito (approved):

PR #554, opened 2025-04-16
Adding new order exception and using it in find methods

ArnabChatterjee20k:
* Added a new Order Exception
* Replaced DatabaseException with Order exception in the db adapters when cursor pointing document is null

### DIFF
--- a/src/Database/Adapter/MariaDB.php
+++ b/src/Database/Adapter/MariaDB.php
@@ -11,6 +11,7 @@ use Utopia\Database\Document;
 use Utopia\Database\Exception as DatabaseException;
 use Utopia\Database\Exception\Duplicate as DuplicateException;
 use Utopia\Database\Exception\NotFound as NotFoundException;
+use Utopia\Database\Exception\Order;
 use Utopia\Database\Exception\Timeout as TimeoutException;
 use Utopia\Database\Exception\Truncate as TruncateException;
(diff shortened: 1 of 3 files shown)


src/Database/Adapter/MariaDB.php
```diff
@@ -11,6 +11,7 @@
 use Utopia\Database\Exception as DatabaseException;
 use Utopia\Database\Exception\Duplicate as DuplicateException;
 use Utopia\Database\Exception\NotFound as NotFoundException;
+use Utopia\Database\Exception\Order;
```
abnegate, line 14: Let's follow the pattern and alias this as `OrderException`

src/Database/Adapter/MariaDB.php
```diff
@@ -1716,7 +1717,7 @@ public function find(string $collection, array $queries = [], ?int $limit = 25,
                 }
 
                 if (\is_null($cursor[$originalAttribute] ?? null)) {
-                    throw new DatabaseException("Order attribute '{$originalAttribute}' is empty");
+                    throw new Order("Order attribute '{$originalAttribute}' is empty", 400, null, $originalAttribute);
```
abnegate, line 1720: From the database, we can leave the `code` empty, the database doesn't know about HTTP codes

abnegate (approved):

PR #557, opened 2025-04-21
Skip auth in bulks

fogelito:
- Nothing to skip in updateDocuments.
- @abnegate  Should we skip auth in the find method?

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -4212,7 +4212,7 @@ class Database
                 }
             }
 
-            $this->withTransaction(function () use ($collection, $updates, $authorization, $skipAuth, $batch) {
+            $this->withTransaction(function () use ($collection, $updates, $batch) {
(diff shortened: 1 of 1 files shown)


src/Database/Database.php
```diff
@@ -4179,11 +4179,15 @@ public function updateDocuments(
...
+            $getResults = fn () => $this->find(
                 $collection->getId(),
                 array_merge($new, $queries),
                 forPermission: Database::PERMISSION_UPDATE
-            ));
+            );
+
+            $batch = $this->silent(
+                fn () => $skipAuth ? $authorization->skip($getResults) : $getResults()
+            );
```
abnegate, line 4190: Do we need this? Won't `find` handle it already?

fogelito: Yes , I think it will be checked.. 
Shoud we remove the existing $authorization check and queries check if they will be checked anyhow?

PR #558, opened 2025-04-21
Update using internal Id

fogelito:


fogelito:
Do we call the adapter directly from somewhere like migrations? 
Perhaps we can send to delete the internalId as well?

abnegate:
> Do we call the adapter directly from somewhere like migrations? Perhaps we can send to delete the internalId as well?

Not adapter directly, we still use the `Database` wrapper

abnegate (approved):

PR #559, opened 2025-04-23
Auto-increment integer attribute

ArnabChatterjee20k:
### About
Add a new parameter to createAttribute in Database, bool autoIncrement = false. If set to true, and it's an integer type attribute, the added column should be an auto-increment column that is not required to be set manually

### Behaviour
* createAttribute -> add autoIncrement to the metadata. Basically a global autoincrement value

* createDocument -> add +1 to the autoIncrement value or if any value passed then update that document and have the max last increment and the current value

* updateDocument -> no modification in the global value

* deleteDocument -> no reduction in the global value

### Testing
docker compose exec tests vendor/bin/phpunit tests/e2e/Adapter/SharedTables/MariaDBTest.php --filter testCreateAutoIncrementAttributes

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1521,7 +1521,10 @@ class Database
             'filters' => $filters
         ];
         if ($type === Database::VAR_INTEGER && $autoIncrement === true) {
-            $attributeInput['autoIncrement'] = 0;
+            if ($required == true) {
+                throw new DatabaseException('Cannot set auto increment for a required attribute');
+            }
+            $attributeInput['autoIncrement'] = true;
         }
         $attribute = new Document($attributeInput);
 
@@ -1830,14 +1833,6 @@ class Database
      * @return Document
      * @throws Exception
      */
-    public function updateAttributeAutoIncrement(string $collectionId, string $attributeId, int $newValue): Document
-    {
-        return $this->updateAttributeMeta($collectionId, $attributeId, function (Document $attribute) use ($newValue) {
-            $currentValue = $attribute->getAttribute('autoIncrement');
-            $updatedValue = max($currentValue, $newValue);
-            $attribute->setAttribute('autoIncrement', $updatedValue);
-        });
-    }
 
     /**
      * Update Attribute. This method is for updating data that causes underlying structure to change. Check out other updateAttribute methods if you are looking for metadata adjustments.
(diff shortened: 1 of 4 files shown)


src/Database/Database.php
```diff
@@ -1572,7 +1577,10 @@ public function createAttribute(string $collection, string $id, string $type, in
         }
 
         if ($collection->getId() !== self::METADATA) {
+            // FIXME: remove the unused methods
             $this->silent(fn () => $this->updateDocument(self::METADATA, $collection->getId(), $collection));
+            $val = $this->silent(fn () => $this->getDocument(self::METADATA, $collection->getId()));
```
fogelito, line 1582: Reminder to remove..

ArnabChatterjee20k: Thanks for pointing it out. Will remove it

src/Database/Database.php
```diff
@@ -3429,8 +3456,24 @@ public function createDocument(string $collection, Document $document): Document
         if (!$structure->isValid($document)) {
             throw new StructureException($structure->getDescription());
         }
+        $document = $this->withTransaction(function () use ($collection, $collectionId, $document) {
+            $metadataDocument = $this->getDocument(Database::METADATA, $collectionId);
```
fogelito, line 3460: No need to getDocument here , you have $collection above..

src/Database/Database.php
```diff
@@ -1517,8 +1518,12 @@ public function createAttribute(string $collection, string $id, string $type, in
             'array' => $array,
             'format' => $format,
             'formatOptions' => $formatOptions,
-            'filters' => $filters,
-        ]);
+            'filters' => $filters
+        ];
+        if ($type === Database::VAR_INTEGER && $autoIncrement === true) {
+            $attributeInput['autoIncrement'] = 0;
```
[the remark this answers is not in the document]
fogelito: @abnegate, will we need to create a new attribute, autoIncrement, in the **Attributes** collection in Appwrite?

src/Database/Database.php
```diff
@@ -3429,8 +3453,24 @@ public function createDocument(string $collection, Document $document): Document
         if (!$structure->isValid($document)) {
             throw new StructureException($structure->getDescription());
         }
-
         $document = $this->withTransaction(function () use ($collection, $document) {
+            $collectionId = $collection->getId();
+            $attributes = $collection->getAttribute('attributes', []);
+
+            foreach ($attributes as $index => $attr) {
+                if ($attr instanceof Document && $attr->getAttribute('autoIncrement') !== null) {
```
abnegate, line 3461: Let's invert and early `continue` to reduce nesting

abnegate, line 3461: You shouldn't need `$attr instanceof Document`, should always be a document

src/Database/Database.php
```diff
@@ -1815,6 +1820,25 @@ public function updateAttributeDefault(string $collection, string $id, mixed $de
...
+     * Update autoIncrement attribute of metadata of a collection.
+     *
+     * @param string $collectionId
+     * @param string $attributeId
+     * @param int $newValue
+     *
+     * @return Document
+     * @throws Exception
+     */
+    public function updateAttributeAutoIncrement(string $collectionId, string $attributeId, int $newValue): Document
```
abnegate, line 1833: For consistency, second param should just be called `$id`

src/Database/Database.php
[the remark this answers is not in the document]
abnegate: @fogelito Yep we'll need to add to the Appwrite collection too, over there it can just be a bool, I don't think we'll need to expose the counter per collection

@ArnabChatterjee20k Agree with Shmuel, we should keep as before and just add `$attribute->setAttribute('autoIncrement', 1)` (making sure to start from 1)

abnegate (commented):
As is, we will have concurrency issues. Two concurrent requests can read `autoIncrement` as 5, both increment to 6, then the counter is out of sync.

To avoid this, we need the metadata update to be atomic. To achieve it we'll need some more changes:

- `getCollection` function should have a new `$forUpdate = false` parameter
- `$forUpdate` should be passed through to the `getDocument(METADATA, ...)` call inside getCollection.
- Pass `$forUpdate = true` when fetching the collection at the start of `updateAttributeMeta`

That will make the metadata read use `SELECT ... FOR UPDATE`, which will mean the metadata document is locked until each transaction is complete.

ArnabChatterjee20k:
closing this pr as the implementation plan is changed

PR #560, opened 2025-04-24
added block queries against encrypt fields

ArnabChatterjee20k:
For blocking queries against attributes against encrypt filters
https://linear.app/appwrite/issue/DAT-440/block-queries-against-encrypted-attributes-with-an-informative-error

### DIFF
--- a/src/Database/Validator/Query/Filter.php
+++ b/src/Database/Validator/Query/Filter.php
@@ -41,6 +41,13 @@ class Filter extends Base
      */
     protected function isValidAttribute(string $attribute): bool
     {
+        if (
+            \in_array('encrypt', $this->schema[$attribute]['filters'] ?? [])
+        ) {
+            $this->message = 'Cannot query encrypted attribute: ' . $attribute;
+            return false;
+        }
+
         if (\str_contains($attribute, '.')) {
             // Check for special symbol `.`
             if (isset($this->schema[$attribute])) {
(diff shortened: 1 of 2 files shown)


ItzNotABug (approved):

src/Database/Validator/Query/Filter.php
```diff
@@ -74,6 +74,13 @@ protected function isValidAttribute(string $attribute): bool
      */
     protected function isValidAttributeAndValues(string $attribute, array $values, string $method): bool
     {
+        if (
+            isset($this->schema[$attribute]['filters'])
+            && in_array('encrypt', $this->schema[$attribute]['filters'])
+        ) {
+            $this->message = 'Cannot select encrypted attribute: ' . $attribute;
+            return false;
+        }
```
abnegate, line 83: We should do this inside `isValidAttribute`

src/Database/Validator/Query/Filter.php
```diff
@@ -74,6 +74,13 @@ protected function isValidAttribute(string $attribute): bool
      */
     protected function isValidAttributeAndValues(string $attribute, array $values, string $method): bool
     {
+        if (
+            isset($this->schema[$attribute]['filters'])
+            && in_array('encrypt', $this->schema[$attribute]['filters'])
+        ) {
+            $this->message = 'Cannot select encrypted attribute: ' . $attribute;
```
abnegate, line 81:
```suggestion
            $this->message = 'Cannot query encrypted attribute: ' . $attribute;
```

abnegate (approved):

PR #565, opened 2025-05-02
Manage wildcards and remove extra filter loop

ItzNotABug:


ItzNotABug:
@abnegate the tests related to relationships are a bit broken now due to removing the filter check.

PR #571, opened 2025-05-05
Fix: collection deletion on V1

Meldiron:
Follows same approach as other places that need backwards support for shared tables V1.

CleanShot 2025-05-05 at 19 15 39@2x

CleanShot 2025-05-05 at 19 15 53@2x

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -3,7 +3,6 @@
 namespace Utopia\Database;
 
 use Exception;
-use PDOException;
 use Utopia\Cache\Cache;
 use Utopia\CLI\Console;
 use Utopia\Database\Exception as DatabaseException;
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -1434,11 +1435,22 @@ public function deleteCollection(string $id): bool
...
+        try {
+            foreach ($relationships as $relationship) {
+                $this->deleteRelationship($collection->getId(), $relationship->getId());
+            }
+
+            $this->adapter->deleteCollection($id);
+        } catch (PDOException $e) {
+            // HACK: Metadata should still be updated, can be removed when null tenant collections are supported.
+            // "42S02" is "Base table or view not found": https://learn.microsoft.com/en-us/sql/odbc/reference/appendixes/appendix-a-odbc-error-codes?view=sql-server-ver16
+            if ($e->getCode() === '42S02') {
```
abnegate, line 1448: We shouldn't check `PDOException` at the Database level, can't be sure we're using PDO here. The adapter method should have a try/catch around the PDO call, pass the exception to `processException` and rethrow that. Then here, you can just check `NotFoundException`. Check `createAttribute`, we use this pattern there.

abnegate, line 1448: There should two separate try/catches here: one inside the for loop, and one around the delete collection, otherwise failing to delete one relationship will still exit early

Meldiron: I actually dont need it, we dont have relations in appwrite tables

Meldiron: Updated

abnegate (approved):

PR #575, opened 2025-05-07
Skip documents without changes

abnegate:


src/Database/Database.php
```diff
@@ -4731,6 +4731,11 @@ public function createOrUpdateDocumentsWithIncrease(
                 )));
             }
 
+            if ($old->getAttributes() == $document->getAttributes()) {
```
abnegate, line 4734: Loose check because we don't care about the order

src/Database/Database.php
```diff
@@ -4731,6 +4731,11 @@ public function createOrUpdateDocumentsWithIncrease(
...
+                unset($documents[$key]);
```
fogelito, line 4735: Since unset($documents[$key]); 
Do we need to run array_values?

src/Database/Database.php
abnegate: Updated to only ignore $tenant and $internalId

PR #581, opened 2025-05-13
Add bulk create attribute function

abnegate:
- Add bulk create attributes function


phpunit.xml
```diff
@@ -16,14 +16,4 @@
...
-    <coverage processUncoveredFiles="true">
-        <include>
-            <directory suffix=".php">./src/</directory>
-            <directory suffix=".php">./tests/</directory>
-        </include>
-        <report>
-            <clover outputFile="tmp/clover.xml"/>
-            <html outputDirectory="tmp/coverage"/>
-        </report>
-    </coverage>
```
abnegate, line 28: We're not using this because of split CI tests, but the report generation still happens, which slows tests down.

composer.json
```diff
@@ -56,8 +56,8 @@
     },
     "config": {
         "allow-plugins": {
-            "php-http/discovery": true,
-            "tbachert/spi": true
+            "php-http/discovery": false,
+            "tbachert/spi": false
```
abnegate, line 60: Don't know why these were enabled here, they're a transitive dependency, we don't need them

src/Database/Database.php
```diff
@@ -1544,12 +1545,192 @@ public function createAttribute(string $collection, string $id, string $type, in
...
+                throw $e;
+            }
+        }
+
+        if ($collection->getId() !== self::METADATA) {
+            $this->silent(fn () => $this->updateDocument(self::METADATA, $collection->getId(), $collection));
+        }
+
+        $this->purgeCachedCollection($collection->getId());
+        $this->purgeCachedDocument(self::METADATA, $collection->getId());
```
fogelito, line 1586: Will 
$this->updateDocument(self::METADATA, $collection->getId(), $collection))
Do it ?

fogelito (approved):

src/Database/Database.php
abnegate: Seems it's not 🤔 will revert and investigate separately

PR #585, opened 2025-05-14
Re-force internal attribute selection

abnegate:
Because without them:

- It breaks select queries in all statically typed SDKs,
- There are many places we need them internally
- Select queries meant to be for choosing what of your data you want, regardless of internals
- They add <1KB of data, negligible in almost all scenarios

### DIFF
--- a/src/Database/Adapter/SQL.php
+++ b/src/Database/Adapter/SQL.php
@@ -1684,30 +1684,18 @@ abstract class SQL extends Adapter
             return '*';
         }
 
-        $selections = \array_diff($selections, ['$id', '$permissions', '$collection']);
+        $internalKeys = [
+            '$id',
+            '$internalId',
+            '$permissions',
+            '$createdAt',
+            '$updatedAt',
+        ];
 
-        $selections[] = $this->getInternalKeyForAttribute('$id');
-        $selections[] = $this->getInternalKeyForAttribute('$permissions');
+        $selections = \array_diff($selections, [...$internalKeys, '$collection']);
 
-        if (\in_array('$internalId', $selections)) {
-            $selections[] = $this->getInternalKeyForAttribute('$internalId');
-            $selections = \array_diff($selections, ['$internalId']);
-        }
-        if (\in_array('$createdAt', $selections)) {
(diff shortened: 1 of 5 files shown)


src/Database/Database.php
```diff
@@ -3169,6 +3169,7 @@ public function deleteIndex(string $collection, string $id): bool
      */
     public function getDocument(string $collection, string $id, array $queries = [], bool $forUpdate = false): Document
     {
+
```
abnegate, line 3172:
```suggestion
```

tests/e2e/Adapter/Scopes/DocumentTests.php
```diff
@@ -1038,33 +985,6 @@ public function testFind(): array
         ];
     }
 
-    /**
-    * @return void
-    * @throws \Utopia\Database\Exception
-    */
-    public function testSelectInternalID(): void
```
fogelito, line 1045: Why was this tests removed?

src/Database/Adapter/SQL.php
```diff
@@ -1684,30 +1684,18 @@ protected function getAttributeProjection(array $selections, string $prefix = ''
             return '*';
         }
 
-        $selections = \array_diff($selections, ['$id', '$permissions', '$collection']);
```
fogelito, line 1687: They were removed but manually added 
$permissions , $id 
Since they were used for getDocument , $permissions for sure?

tests/e2e/Adapter/Scopes/DocumentTests.php
abnegate: It's redundant since Internal ID is always selected

src/Database/Adapter/SQL.php
abnegate: Permissions get added in the adapters, they can't be part of the select query since they're in the `_perms` table

fogelito: getDocument do not reach _perms table that is why we always return it for getDocument, to avoid the subquery

abnegate: Got it, it's working because we are still adding it in the $internalKeys array below

fogelito (approved):

PR #588, opened 2025-05-19
Cursor POC

fogelito:
This PR will be a breaking change for Appwrite.


src/Database/Database.php
```diff
@@ -5775,17 +5788,21 @@ public function deleteDocuments(
...
 
             /**
              * @var array<Document> $batch
              */
-            $batch = $this->silent(fn () => $this->find(
-                $collection->getId(),
-                array_merge($new, $queries),
-                forPermission: Database::PERMISSION_DELETE
+
+            $batch = $this->silent(fn () => $this->withCursor(
```
fogelito, line 5799: Hmm, thinking perhaps add a second param to the Query::cursor , which will be the payload 
query::cursor(string $id, Document $payload)
So if we have a payload will use that and not the $id?

PR #589, opened 2025-05-20
added test for global collection test

ArnabChatterjee20k:
function setGlobalCollections, that allows you to specify collections that should not include $tenant in their cache key, returned by getCacheKeys
So added tests for this

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1046,6 +1046,11 @@ class Database
         return \array_keys($this->globalCollections);
     }
 
+    public function resetGlobalCollections(): void
+    {
+        $this->globalCollections = [];
+    }
+
     /**
      * Get list of keywords that cannot be used
      *
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -1046,6 +1046,11 @@ class Database
         return \array_keys($this->globalCollections);
     }
 
+    /**
+     * Clear global collections
+     *
+     * @return void
+     */
     public function resetGlobalCollections(): void
     {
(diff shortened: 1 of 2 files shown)


tests/e2e/Adapter/Scopes/CollectionTests.php
```diff
@@ -1439,4 +1439,55 @@ public function testTransformations(): void
...
     }
+
+    public function testSetGlobalCollection(): void
+    {
+        $db = static::getDatabase();
+
+        $collectionId = 'globalCollection';
+
+        // set collection as global
+        $db->setGlobalCollections([$collectionId]);
```
fogelito, line 1450: At the end of the tests, we should set the back to the initial state, so it will not effect other tests I think

ArnabChatterjee20k: Here I think it will not create an issue as here an actual collection in the db is not getting created and the ids are totally segregated

abnegate: Still good practice not to leak state between tests, please address

src/Database/Database.php
```diff
@@ -1046,6 +1046,11 @@ public function getGlobalCollections(): array
         return \array_keys($this->globalCollections);
     }
 
+    public function resetGlobalCollections(): void
+    {
```
abnegate, line 1050: Let's add a dockblock to explain what this does for consumers

abnegate (approved):

PR #596, opened 2025-05-27
Abstract execution method

fogelito:


src/Database/Adapter.php
```diff
@@ -1186,4 +1186,10 @@ abstract public function getSchemaAttributes(string $collection): array;
      * @return string
      */
     abstract public function getTenantQuery(string $collection, string $alias = ''): string;
+
+    /**
+     * @param mixed $stmt
+     * @return bool
+     */
+    abstract protected function execute(mixed $stmt): bool;
```
fogelito, line 1194: Should the name be executeStatement?

abnegate: This fine, make enough sense with the $stmt param

abnegate (approved):

PR #621, opened 2025-07-10
Add error handling for document updates with onError callback

ArnabChatterjee20k:
* Changes made to ensure an ongoing update doesn't stop due to exceptions in the `onNext` callable
* If exceptions coming in `onNext` passing them to the `onError` callback
* Relationship updates moved inside the transaction to ensure both relations and the original documents are deleted instead of relation deleted and the original not

## Summary by CodeRabbit

* **New Features**
  * Added support for an optional error handling callback during batch document updates to handle exceptions gracefully.
  * Introduced end-to-end tests verifying batch update behavior with callbacks and error handling.
  * Added tests for batch updates in various relationship types (one-to-one, one-to-many, many-to-one, many-to-many) ensuring data integrity and validation.

* **Refactor**
  * Enhanced batch document updates by processing all changes within a single transaction for improved consistency and reliability.

### DIFF
--- a/tests/e2e/Adapter/Scopes/Relationships/ManyToManyTests.php
+++ b/tests/e2e/Adapter/Scopes/Relationships/ManyToManyTests.php
@@ -6,6 +6,7 @@ use Exception;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
 use Utopia\Database\Exception\Restricted as RestrictedException;
+use Utopia\Database\Exception\Structure;
 use Utopia\Database\Helpers\ID;
 use Utopia\Database\Helpers\Permission;
 use Utopia\Database\Helpers\Role;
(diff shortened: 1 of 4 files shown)


src/Database/Database.php
```diff
@@ -4452,7 +4453,11 @@ public function updateDocuments(
             foreach ($batch as $doc) {
                 $this->purgeCachedDocument($collection->getId(), $doc->getId());
                 $doc = $this->decode($collection, $doc);
-                $onNext && $onNext($doc);
+                try {
+                    $onNext && $onNext($doc);
+                } catch (Exception $e) {
+                    $onError && $onError($e);
```
fogelito, line 4459: I think we need to throw the exception in case no onerror callback?

ArnabChatterjee20k: According to the linear ticket, the execution should go and if error comes it should get passed to the onError callback
If rethrown the execution will stop

PR #625, opened 2025-07-14
Mongo transactions

shimonewman:


src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,153 @@ public function clearTimeout(string $event): void
...
+                // This ensures the transaction is active even if validation fails later.
+                $this->client->query([
+                    'find' => 'system.version',
+                    'filter' => $this->client->toObject([]),
+                    'limit' => 1,
+                    'lsid' => ['id' => $this->sessionId],
+                    'txnNumber' => new \MongoDB\BSON\Int64($this->txnNumber), // Long type for txnNumber
+                    'autocommit' => false,
+                    'startTransaction' => true
+                ], 'admin');
```
abnegate, line 135: Why do we need the transaction active if validation fails? Doesn't `startSession` start the txn already?

src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,153 @@ public function clearTimeout(string $event): void
...
+
+                $this->firstOpInTransaction = false;
+            }
+            $this->inTransaction++;
+            return true;
+        } catch (\Throwable $e) {
+            throw new DatabaseException('Failed to start transaction: ' . $e->getMessage(), $e->getCode(), $e);
```
abnegate, line 142: For transactions errors, let's throw `TransactionException`

src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,153 @@ public function clearTimeout(string $event): void
...
+        }
     }
 
     public function commitTransaction(): bool
     {
-        return true;
+        try {
+            if ($this->inTransaction === 0) {
+                throw new DatabaseException('No transaction in progress');
+            }
```
abnegate, line 151: In this case, can we return false the same as SQL does?

src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,153 @@ public function clearTimeout(string $event): void
...
+                    throw new DatabaseException('No session in progress');
+                }
+                $result = $this->client->commitTransaction(
+                    ['id' => $this->sessionId], // Pass raw id object
+                    $this->txnNumber,
+                    false
+                );
+                if (($result->ok ?? 0) !== 1.0) {
+                    throw new DatabaseException('Failed to commit transaction');
+                }
```
abnegate, line 164: Throwing inside the try will be caught in the catch, and the exception message will be "Failed to commit transaction: Failed to commit transaction", let's narrow the scope of the try/catch and use multiple if needed

src/Database/Adapter/Mongo.php
shimonewman: startSession only opens the session against the server; transactions are incremented manually via startTransaction

src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,161 @@ public function clearTimeout(string $event): void
...
+     * @param callable(): T $callback
+     * @return T
+     * @throws \Throwable
+     */
+    public function withTransaction(callable $callback): mixed
+    {
+        // If the database is not a replica set, we can't use transactions
+        if(!$this->client->isReplicaSet()){
+            return true;
+        }
```
abnegate, line 95: In this scenario we should still run the callback and return it's result

src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,161 @@ public function clearTimeout(string $event): void
...
+                        ['id' => $this->sessionId], // Pass raw id object
+                        $this->txnNumber,
+                        false
+                    );
+                } catch (\Throwable $e) {
+                    throw new DatabaseException($e->getMessage(), $e->getCode(), $e);
+                }
+
+                // Session is now closed by the client using endSessions,  state is reseted
+                // TODO  do we want  session per transaction or to manage it on the connection level?
```
abnegate, line 174: Session per txn is fine :+1: in fact we will need it that way since Appwrite pools use a different connection for each DB call

src/Database/Adapter/Mongo.php
```diff
@@ -73,19 +81,161 @@ public function clearTimeout(string $event): void
...
+            throw new DatabaseException('Failed to rollback transaction: ' . $e->getMessage(), $e->getCode(), $e);
+        }
+    }
+
+    /**
+     * Helper to add transaction/session context to command options if in transaction
+     */
+    private function addTransactionContext(array $options = []): array
+    {
+
```
abnegate, line 226:
```suggestion
```

src/Database/Adapter/Mongo.php
```diff
@@ -838,25 +988,30 @@ public function createDocuments(string $collection, array $documents): array
...
                 $name,
                 $filters,
-                ['limit' => 1]
+                array_merge($options, ['limit' => 1])
             )->cursor->firstBatch[0];
 
+            /**
+             * TODO Do we even need this find?
+             * We can just return the result from the insertDocument.
+             */
```
abnegate, line 1013: We can just return from insert if that works

PR #639, opened 2025-07-27
Fix encoding decoding

fogelito:
Fix Null permissions issue added in decode

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of document permissions to ensure documents created without explicit permissions include a `$permissions` key set to an empty array.
  * Enhanced robustness in internal attribute processing by standardizing attribute handling and refining encoding and decoding of `$permissions`.
  * Strengthened validation of document attributes to correctly check presence and types of `$id` and `$permissions`.

* **Tests**
  * Added tests to verify that documents created without permissions contain a `$permissions` key with an empty array as its value.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -6293,9 +6293,7 @@ class Database
         $attributes = $collection->getAttribute('attributes', []);
 
         foreach ($this->getInternalAttributes() as $attribute) {
-            if ($attribute['$id'] !== '$permissions') { // Don't encode permissions into a JSON string
-                $attributes[] = new Document($attribute);
-            }
+            $attributes[] = $attribute;
         }
 
         foreach ($attributes as $attribute) {
@@ -6304,6 +6302,13 @@ class Database
             $default = $attribute['default'] ?? null;
             $filters = $attribute['filters'] ?? [];
             $value = $document->getAttribute($key);
+
+            if ($attribute['$id'] === '$permissions') {
+                if (empty($value)){
+                    $document->setAttribute('$permissions', []); // set default value
+                }
+                continue;
+            }
 
             // Continue on optional param with no default
             if (is_null($value) && is_null($default)) {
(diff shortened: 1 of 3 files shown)


src/Database/Database.php
```diff
@@ -6362,18 +6361,19 @@ public function decode(Document $collection, Document $document, array $selectio
         foreach ($relationships as $relationship) {
             $key = $relationship['$id'] ?? '';
 
-            if (
-                \array_key_exists($key, (array)$document)
-                || \array_key_exists($this->adapter->filter($key), (array)$document)
-            ) {
+            if ($document->offsetExists($key) || $document->offsetExists($this->adapter->filter($key))) {
```
fogelito, line 6364: We can use native function for arrayOffset

tests/e2e/Adapter/Scopes/PermissionTests.php
```diff
@@ -14,6 +14,40 @@
...
+        /** @var Database $database */
+        $database = static::getDatabase();
+
+        $database->createCollection(__FUNCTION__);
+
+        /**
+         * Validate the decode function does not add $permissions null entry when no permissions are provided
+         */
+
+        $document = $database->createDocument(__FUNCTION__, new Document());
```
fogelito, line 28: @abnegate If we create a Document with no $permissions entry , 
On The created Document we will not see an Entry, Is this expected, or we should see empty array ?

src/Database/Database.php
```diff
@@ -6392,10 +6392,11 @@ public function decode(Document $collection, Document $document, array $selectio
             $value = ($array) ? $value : [$value];
             $value = (is_null($value)) ? [] : $value;
 
-            foreach ($value as &$node) {
-                foreach (\array_reverse($filters) as $filter) {
+            foreach ($value as $k => $node) {
```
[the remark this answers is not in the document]
abnegate: Let's use a better name, $key, or something more descriptive if that collides

src/Database/Database.php
fogelito: offsetExists  returns true for nulls

fogelito: Reverted

abnegate (approved):

PR #642, opened 2025-07-30
Allow modifying createdAt and updatedAt

ArnabChatterjee20k:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation to prevent empty string values for date fields when creating and updating documents, ensuring more reliable data integrity.
  * Enforced mandatory presence of creation and update timestamps on documents.
  * Refined handling of creation and update timestamps to better preserve or auto-populate dates based on configuration.
  * Enhanced database update operations to include creation timestamps when present.

* **Tests**
  * Added and enhanced tests to verify rejection of empty string date values and correct handling of valid and null dates during document creation and updates.
  * Added tests to confirm proper preservation and auto-population of creation and update dates in single and bulk document operations.


src/Database/Validator/PartialStructure.php
```diff
@@ -36,7 +37,16 @@ public function isValid($document): bool
...
+        $requiredAttributesMap = [];
+        foreach ($this->attributes as $attribute) {
+            if ($attribute['required'] === true && in_array($attribute['$id'], $requiredAttributes)) {
+                $requiredAttributesMap[] = $attribute;
+            }
+        }
 
+        if (!$this->checkForAllRequiredValues($structure, $requiredAttributesMap, $keys)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
ArnabChatterjee20k: its fine here as they are internal attributes here

src/Database/Database.php
```diff
@@ -4119,10 +4119,11 @@ public function updateDocument(string $collection, string $id, Document $documen
 
                 $skipPermissionsUpdate = ($originalPermissions === $currentPermissions);
             }
+            $createdAt = $document->getCreatedAt();
 
             $document = \array_merge($old->getArrayCopy(), $document->getArrayCopy());
             $document['$collection'] = $old->getAttribute('$collection');   // Make sure user doesn't switch collection ID
-            $document['$createdAt'] = $old->getCreatedAt();                 // Make sure user doesn't switch createdAt
+            $document['$createdAt'] = ($createdAt === null || !$this->preserveDates) ? $old->getCreatedAt() : $createdAt;  // Make sure user doesn't switch createdAt
```
abnegate, line 4126: We can remove the comment now

src/Database/Validator/PartialStructure.php
```diff
@@ -36,7 +37,19 @@ public function isValid($document): bool
...
         }
+        /**
+         * @var array<mixed,mixed> $requiredAttributesMap
+         */
+        $requiredAttributesMap = [];
+        foreach ($this->attributes as $attribute) {
+            if ($attribute['required'] === true && in_array($attribute['$id'], $requiredAttributes)) {
+                $requiredAttributesMap[] = $attribute;
+            }
+        }
```
abnegate, line 48: I think we can do the same by just checking `$attribute['required'] === true && $document->offsetExists($attribute['$id'])`, then we don't need the new parameter

tests/e2e/Adapter/Scopes/GeneralTests.php
```diff
@@ -186,16 +254,33 @@ public function testPreserveDatesCreate(): void
...
+                '$permissions' => [],
+                'attr1' => 'value3',
             ]),
         ], batchSize: 2);
 
         $doc1 = $database->getDocument('preserve_create_dates', 'doc1');
         $doc2 = $database->getDocument('preserve_create_dates', 'doc2');
         $doc3 = $database->getDocument('preserve_create_dates', 'doc3');
+        $doc4 = $database->getDocument('preserve_create_dates', 'doc4');
+        $doc5 = $database->getDocument('preserve_create_dates', 'doc5');
```
abnegate, line 276: Shouldn't this fail because doc4 has null createdAt?

ArnabChatterjee20k: null basically means passing nothing as per the document structure => so will get replaced by default time => which is the default behaviour

abnegate (approved):

PR #653, opened 2025-08-07
Add date-based query helpers for createdAt and updatedAt

abnegate:


tests/e2e/Adapter/Scopes/DocumentTests.php
```diff
@@ -2644,6 +2644,110 @@ public function testFindOrderByUpdateDateAndCursor(): void
...
+        ]);
+
+        $this->assertGreaterThan(0, count($documents));
+
+        $documents = $database->find('movies', [
+            Query::createdBefore($pastDate),
+            Query::limit(1)
+        ]);
+
+        $this->assertEquals(0, count($documents));
```
fogelito, line 2670: Let's add tests that return results?

tests/e2e/Adapter/Scopes/DocumentTests.php
```diff
@@ -2644,6 +2644,110 @@ public function testFindOrderByUpdateDateAndCursor(): void
...
 
+    public function testFindCreatedBefore(): void
+    {
+        /** @var Database $database */
+        $database = static::getDatabase();
+
+        /**
+         * Test Query::createdBefore wrapper
+         */
+        $futureDate = '2050-01-01T00:00:00.000Z';
```
fogelito, line 2655: Let's test regular dates work as expected, since convert queries will convert them anyway

tests/e2e/Adapter/Scopes/DocumentTests.php
abnegate: We do, this test ensure we get 0 results, the previous check ensure we get more than 0 results

tests/e2e/Adapter/Scopes/DocumentTests.php
abnegate: That won't work (for any query), convertQueries expects to get a string, then converts it to a date

src/Database/Query.php
```diff
@@ -625,6 +625,50 @@ public static function notEndsWith(string $attribute, string $value): self
...
+        return self::lessThan('$createdAt', $isoDateString);
+    }
+
+    /**
+     * Helper method to create Query for documents created after a specific date
+     *
+     * @param string $isoDateString
+     * @return Query
+     */
+    public static function createdAfter(string $isoDateString): self
```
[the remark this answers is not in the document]
abnegate: True, from here, we actually accept anything that's a valid date for PHP. @cursoragent change the parameter name to `$value`, to be consistent with other methods as well.

PR #659, opened 2025-08-11
Fix cache clears in finally block

abnegate:


src/Database/Database.php
```diff
@@ -2219,11 +2239,7 @@ public function updateAttribute(string $collection, string $id, ?string $type =
                 if (!$updated) {
                     throw new DatabaseException('Failed to update attribute');
                 }
-
-                $this->purgeCachedCollection($collection);
             }
-
-            $this->purgeCachedDocument(self::METADATA, $collection);
```
abnegate, line 2226: @cursor don't remove these, even if they're redundant

src/Database/Database.php
```diff
@@ -2813,9 +2839,6 @@ function ($index) use ($newKey) {
                 throw new RelationshipException('Invalid relationship type.');
         }
 
-        $this->purgeCachedCollection($collection->getId());
-        $this->purgeCachedCollection($relatedCollection->getId());
-
```
abnegate, line 2818: @cursor Where is this handled instead? Seems required

PR #665, opened 2025-08-15
Add vector attribute support for PostgreSQL with pgvector extension

abnegate:


src/Database/Database.php
```diff
@@ -1834,8 +1835,21 @@ private function validateAttribute(
...
             case self::VAR_RELATIONSHIP:
                 break;
+            case self::VAR_VECTOR:
+                if (!($this->adapter instanceof \Utopia\Database\Adapter\Postgres)) {
+                    throw new DatabaseException('Vector type is only supported in PostgreSQL adapter');
+                }
+                if ($size <= 0) {
+                    throw new DatabaseException('Vector dimensions must be a positive integer');
+                }
+                if ($size > 16000) { // pgvector limit
```
abnegate, line 1845: @cursor add a const instead of hardcoding the vector limit

src/Database/Database.php
```diff
@@ -1904,8 +1918,18 @@ protected function validateDefaultTypes(string $type, mixed $default): void
                     throw new DatabaseException('Default value ' . $default . ' does not match given type ' . $type);
                 }
                 break;
+            case self::VAR_VECTOR:
+                if ($defaultType !== 'array') {
+                    throw new DatabaseException('Default value for vector must be an array');
+                }
```
abnegate, line 1924: @cursor does this work with the recursion further up the method? Maybe we need to skip the recursion for vector type?

tests/e2e/Adapter/PostgresTest.php
```diff
@@ -70,4 +73,161 @@ protected static function deleteIndex(string $collection, string $index): bool
...
+            Query::orderAsc('$id')
+        ]);
+
+        $this->assertCount(3, $results);
+
+        // Test vector euclidean distance query
+        $results = $database->find('vectorQueries', [
+            Query::vectorEuclidean('embedding', [1.0, 0.0, 0.0]),
+            Query::orderAsc('$id')
+        ]);
```
abnegate, line 214: @cursor Add some queries that only return some values

abnegate: @cursor I meant make some queries that do not match all vector results. E.g. no overlap that will be returned by and of cosine, euclidean, or dot. To ensure that not all results are always returned for vector queries.

PR #705, opened 2025-09-16
Enable reconnection + retry

abnegate:


src/Database/PDO.php
```diff
@@ -41,7 +42,26 @@ public function __construct(
...
+                Console::warning('[Database] Lost connection detected. Reconnecting...');
+
+                $inTransaction = $this->pdo->inTransaction();
+
+                // Attempt to reconnect
+                $this->reconnect();
+
+                // If we weren't in a transaction, also retry the query
+                // In a transaction we can't retry as the state is attached to the previous connection
+                if (!$inTransaction) {
```
fogelito, line 58: If transaction === true  Do we want to throw a specific error ? 
otherwise return $this->pdo->{$method}(...$args); ?

abnegate: If it was a connection error, but we were in a transaction, we'll throw the original exception

fogelito (approved):

PR #707, opened 2025-09-17
Handling not null spatial column in the existing data

ArnabChatterjee20k:


src/Database/Adapter/MariaDB.php
```diff
@@ -1917,4 +1917,14 @@ public function getSupportForSpatialAxisOrder(): bool
...
     }
+
+    /**
+     * Adapter supports error on not-null spatial attributes with existing rows.
+     *
+     * @return bool
+     */
+    public function getSupportForNotNullSpatialAttributeWithExistingRows(): bool
+    {
+        return false;
```
fogelito: I think we should not allow creating across all adapters so we have one common behaviour to all.
MariaDB can work as Mysql

PR #714, opened 2025-09-23
Add support for schemaless attributes in Database and Structure classes

ArnabChatterjee20k:
### Previous
Structure Validations were coming for collection attributes along with the internal attributes resulting to structure exceptions for free form document structure in the mongodb adapter

### Current
Structure validations happening for the internal attributes only and not for the free form document in the mongodb

### Previous
Previously due to attributes metadata storage , validations were able to able run and validations were getting done on the each attribute. 
Same for query

### Current
For mongodb, the attributes validation are getting skipped -> data directly getting dumped in the underlying db.
For query, we can query on both existing and non existing fields

## Index
Same for index

### DIFF
--- a/src/Database/Validator/Queries/Documents.php
+++ b/src/Database/Validator/Queries/Documents.php
@@ -30,6 +30,7 @@ class Documents extends IndexedQueries
         int $maxValuesCount = 100,
         \DateTime $minAllowedDate = new \DateTime('0000-01-01'),
         \DateTime $maxAllowedDate = new \DateTime('9999-12-31'),
+        private bool $supportForAttributes = true
     ) {
         $attributes[] = new Document([
             '$id' => '$id',
@@ -66,9 +67,10 @@ class Documents extends IndexedQueries
                 $maxValuesCount,
                 $minAllowedDate,
                 $maxAllowedDate,
+                $this->supportForAttributes
             ),
-            new Order($attributes),
-            new Select($attributes),
+            new Order($attributes, $this->supportForAttributes),
+            new Select($attributes, $this->supportForAttributes),
         ];
 
         parent::__construct($attributes, $indexes, $validators);
--- a/src/Database/Validator/Queries/Documents.php
+++ b/src/Database/Validator/Queries/Documents.php
@@ -30,7 +30,7 @@ class Documents extends IndexedQueries
         int $maxValuesCount = 100,
         \DateTime $minAllowedDate = new \DateTime('0000-01-01'),
         \DateTime $maxAllowedDate = new \DateTime('9999-12-31'),
-        private bool $supportForAttributes = true
+        bool $supportForAttributes = true
     ) {
         $attributes[] = new Document([
             '$id' => '$id',
@@ -67,10 +67,10 @@ class Documents extends IndexedQueries
                 $maxValuesCount,
                 $minAllowedDate,
                 $maxAllowedDate,
-                $this->supportForAttributes
+                $supportForAttributes
             ),
-            new Order($attributes, $this->supportForAttributes),
-            new Select($attributes, $this->supportForAttributes),
+            new Order($attributes, $supportForAttributes),
+            new Select($attributes, $supportForAttributes),
         ];
 
         parent::__construct($attributes, $indexes, $validators);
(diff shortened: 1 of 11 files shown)


src/Database/Validator/Queries/Documents.php
```diff
@@ -30,6 +30,7 @@ public function __construct(
         int $maxValuesCount = 100,
         \DateTime $minAllowedDate = new \DateTime('0000-01-01'),
         \DateTime $maxAllowedDate = new \DateTime('9999-12-31'),
+        private bool $supportForAttributes = true
```
abnegate, line 33: No need to make this a property if we only use it in the constructor

src/Database/Validator/Query/Order.php
```diff
@@ -11,12 +11,15 @@ class Order extends Base
...
      */
     protected array $schema = [];
+    protected bool $supportForAttributes;
 
     /**
      * @param array<Document> $attributes
+     * @param bool $supportForAttributes
      */
-    public function __construct(array $attributes = [])
+    public function __construct(array $attributes = [], bool $supportForAttributes = true)
```
abnegate, line 20: We can use a constructor promoted property here

src/Database/Validator/Structure.php
```diff
@@ -106,7 +110,12 @@ public function __construct(
         private readonly string $idAttributeType,
         private readonly \DateTime $minAllowedDate = new \DateTime('0000-01-01'),
         private readonly \DateTime $maxAllowedDate = new \DateTime('9999-12-31'),
+        private bool $supportForAttributes = true
     ) {
+        $this->internalAttributes = array_reduce($this->attributes, function ($carry, $attribute) {
+            $carry[$attribute['$id']] = 1;
+            return $carry;
+        }, []);
```
abnegate, line 118: I don't think we need this, we can use `$this->attributes` without merging with the collection attributes where we need it

src/Database/Validator/Structure.php
```diff
@@ -252,11 +261,14 @@ public function isValid($document): bool
     protected function checkForAllRequiredValues(array $structure, array $attributes, array &$keys): bool
     {
         foreach ($attributes as $key => $attribute) { // Check all required attributes are set
+            // schemaless adapter and not an internal attribute
+            if (!$this->supportForAttributes && !isset($this->internalAttributes[$key])) {
```
abnegate, line 265: Here, we can just use `$this->attributes`

tests/e2e/Adapter/Scopes/AttributeTests.php
```diff
@@ -392,6 +392,10 @@ public function testUpdateAttributeRequired(): void
         /** @var Database $database */
         $database = static::getDatabase();
 
+        if (!$database->getAdapter()->getSupportForAttributes()) {
+            $this->markTestSkipped('This test is only for schema-enforced adapters');
```
abnegate, line 396: Let's use `expectNotToPerformAssertions` instead of `markTestSkipped`, to stop PHPUnit from finishing with warnings

src/Database/Validator/Structure.php
ArnabChatterjee20k: Not using attributes as some point they are getting merged with collection level attributes as well
In schemaless, the validation for types should only be on internal level attributes

abnegate: `$attributes` is merged, but `$this->attributes` should still be only the internals as far as I can see

src/Database/Validator/Structure.php
```diff
@@ -251,12 +252,15 @@ public function isValid($document): bool
      */
     protected function checkForAllRequiredValues(array $structure, array $attributes, array &$keys): bool
     {
-        foreach ($attributes as $key => $attribute) { // Check all required attributes are set
+        foreach ($attributes as $attribute) { // Check all required attributes are set
+            $isInternalAttribute = in_array($attribute['$id'], array_column($this->attributes, '$id'));
+            if (!$this->supportForAttributes && $isInternalAttribute) {
```
abnegate, line 257: Should this be `!$this->supportForAttributes && !$isInternalAttribute`?

abnegate (approved):

PR #715, opened 2025-09-25
fix relationship top level filter

ArnabChatterjee20k:
* Previously, Selecting attributes in a relationship didn't set decoded values in the top-level collection of the the final document in getDocument and find

* Currently, its setting the filtered values to top-level collection if required

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved selective field decoding with nested relationship selections for more consistent results.
  - Ensured non-relationship fields remain accurate when requesting specific related fields.
  - Improved behavior when multiple value filters are applied during selection.

- **Tests**
  - Added end-to-end tests covering mixed selection types (including relationships and nested fields).
  - Added tests for decoding without relationships.
  - Added tests validating combined filters during field selection.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -6750,6 +6750,8 @@ class Database
             fn ($attribute) => $attribute['type'] === self::VAR_RELATIONSHIP
         );
 
+        $filteredValue = [];
+
         foreach ($relationships as $relationship) {
             $key = $relationship['$id'] ?? '';
 
@@ -6798,6 +6800,8 @@ class Database
                 $value[$index] = $node;
             }
 
+            $filteredValue[$key] = ($array) ? $value : $value[0];
+
             if (
                 empty($selections)
                 || \in_array($key, $selections)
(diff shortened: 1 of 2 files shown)


src/Database/Database.php
```diff
@@ -6807,6 +6811,29 @@ public function decode(Document $collection, Document $document, array $selectio
...
+        if (!empty($selections)) {
+            foreach ($selections as $selection) {
+                if (\str_contains($selection, '.')) {
+                    $hasRelationshipSelections = true;
+                    break;
+                }
+            }
+        }
+
+        if ($hasRelationshipSelections && !empty($selections) && !\in_array('*', $selections)) {
```
abnegate, line 6824: I think this also needs to check that there are no non-internal selections on top level? Lets cover it in tests to check

src/Database/Database.php
```diff
@@ -6807,6 +6811,29 @@ public function decode(Document $collection, Document $document, array $selectio
...
+            foreach ($collection->getAttribute('attributes', []) as $attribute) {
+                $key = $attribute['$id'] ?? '';
+
+                if ($attribute['type'] === self::VAR_RELATIONSHIP || $key === '$permissions') {
+                    continue;
+                }
```
abnegate, line 6830: We already have `$relationships` in scope, let's use that instead of iterating again

abnegate: Thought we were skipping non-relationships, but it's the opposite, all good :+1:

PR #852, opened 2026-04-08
Add skipDuplicates() scope guard to createDocuments

premtsd-code:
## Summary

Adds a `Database::skipDuplicates()` scope guard so callers can wrap a `createDocuments()` call and have duplicate-by-id rows silently skipped instead of throwing `DuplicateException`. Used by CSV import / re-import flows and database migration workers where re-running the same batch should be idempotent rather than fatal.

The shape follows the existing `silent()` / `skipRelationships()` pattern — instance-level scope guard, not a flag parameter — so callers compose naturally:

```php
$database->skipDuplicates(fn () => $database->createDocuments($collection, $batch));
```

## Design

Duplicate handling happens at the **adapter layer** via dialect-specific no-op inserts. `Database::createDocuments` sets a scope-guarded flag on the adapter before each chunk's INSERT; the adapter emits the right SQL (or Mongo equivalent) and returns the batch. No pre-filter, no intra-batch dedup — adapter-level row-level dedup is sufficient and avoids redundant queries.

## Per-adapter implementation

| Adapter | Strategy |
|---|---|
| MariaDB / MySQL | `INSERT IGNORE INTO` |
| Postgres | `INSERT INTO ... ON CONFLICT (...) DO NOTHING` |
| SQLite | `INSERT OR IGNORE INTO` |
| Mongo | `upsert + $setOnInsert`, bypassing the transaction wrap to avoid txn-abort-on-duplicate |

## Hot-path cost

One property read + one boolean branch per chunk. Callers that don't use `skipDuplicates` pay no closure allocation and no scope-guard setup — the feature cost is paid only when the feature is actually used. This addresses review comment #1 (don't penalize the common case with an `$enable` flag pattern).

## Mirror dual-write correctness

`Mirror::createDocuments` pre-filters against source to identify which input ids already exist, then after the source insert forwards only the genuinely-new docs to destination. Pre-existing ids were silently no-op'd by the adapter's `INSERT IGNORE` — they represent no user write, so nothing propagates to destination.

This replaces an earlier `captureOnNext`-based approach flagged by Greptile ([#3084293974](https://github.com/utopia-php/database/pull/852#discussion_r3084293974)): that approach wrapped `$onNext` to capture "what source inserted", but SQL adapters return the full input batch from `createDocuments` regardless of `rowCount()`, causing `$onNext` to fire for every doc — including skipped duplicates — so Mirror forwarded would-be values to destination, diverging from source. Pre-filter-then-insert matches the idempotent-write semantics: a skipped duplicate is not a user write, so it does not mirror.

Cost: one extra `SELECT` against source per batch when `skipDuplicates` is active and Mirror has an upgraded destination. Zero cost on the non-skip path. Skip and non-skip paths are unified into a single method body with branches only at the three points where behavior actually differs (source call, forward filter, destination call).

## `upsertDocumentsWithIncrease` tenant grouping (regression fix)

Restored per-tenant grouping for the existing-doc lookup in `sharedTables + tenantPerDocument` mode. An earlier review round (#2 — inline `find()` instead of a helper) was correct for single-tenant use cases but ran under the session tenant context (`null` for platform workers), silently missing rows from other tenants. This broke appwrite's StatsUsage worker, which accumulates stats across many projects and flushes via `$dbForLogs->setTenant(null)->setTenantPerDocument(true) + upsertDocumentsWithIncrease()` — every input doc looked "new" and went down the create path instead of the update path.

The fix: branch on `tenantPerDocument` mode and run one `find()` per unique tenant wrapped in `withTenant($tenant, ...)`. K queries total (K = unique tenants in the batch). Single-tenant callers (the common case, including CSV import and database migration) still hit the fast one-query batched path.

Helper: `tenantKey(Document)` — used only within `upsertDocumentsWithIncrease` to keep the in-memory lookup map and the intra-batch dedup check consistent. Returns `"<tenant>:<id>"` in `tenantPerDocument` mode, just `<id>` otherwise. Without it, a legitimate cross-tenant batch like `[{id: x, tenant: 1}, {id: x, tenant: 2}]` would falsely fail with `DuplicateException`.

## Tests

End-to-end coverage across all adapter variants (MariaDB / MySQL / Postgres / SQLite / MongoDB × Pool / Mirror / SharedTables):

- `testCreateDocumentsIgnoreDuplicates` — mix of existing + new docs, non-existing inserted, existing no-op'd
- `testCreateDocumentsIgnoreAllDuplicates` — every doc already exists, nothing newly inserted
- `testCreateDocumentsSkipDuplicatesEmptyBatch` — empty batch is a no-op
- `testCreateDocumentsSkipDuplicatesNestedScope` — scope guard saves & restores through nested calls
- `testCreateDocumentsSkipDuplicatesLargeBatch` — 300 docs / 50 pre-existing, exercises the chunk loop
- `testCreateDocumentsSkipDuplicatesSecondCallSkipsAll` — idempotent re-invocation preserves first-write values
- `testCreateDocumentsSkipDuplicatesRelationships` — relationship children aren't duplicated on re-run
- `testCreateDocumentsSkipDuplicatesDoesNotDivergeDestination` (Mirror) — source-seeded `dup=Original` with a `dup=WouldBe` caller input, asserts `destination.dup` stays empty and `destination.fresh` gets the newly-inserted row (Greptile's exact assertion from [#3084293974](https://github.com/utopia-php/database/pull/852#discussion_r3084293974))

## Test plan

- [x] skipDuplicates tests pass on every adapter variant
- [x] Full Mirror suite: 635 tests / 11,035 assertions green
- [x] Greptile divergence assertion holds
- [x] `composer analyze` (PHPStan) clean at the project's configured level
- [x] No regressions in existing `testCreateDocuments*` and `testUpsert*` suites

## Summary by CodeRabbit

* **New Features**
  * Scoped "skip duplicates" mode to ignore duplicate records during batch creates without overwriting existing data.

* **Refactor**
  * Duplicate-skipping integrated across adapters, transaction paths, pooling and mirror flows; batch lookups and tenant-aware identity handling improved for large batches and upserts.

* **Tests**
  * Added comprehensive end-to-end tests: nested scopes, large/empty batches, mirror backfill, relationship propagation, and multiple-scope behavior.

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -39,20 +39,13 @@ abstract class Adapter
      * Run a callback with skipDuplicates enabled.
      * Duplicate key errors during createDocuments() will be silently skipped
      * instead of thrown. Nestable — saves and restores previous state.
-     * Pass $enable = false to run the callback without toggling the flag
-     * (useful for conditional forwarding from wrappers like Pool/Mirror).
      *
      * @template T
      * @param callable(): T $callback
-     * @param bool $enable
      * @return T
      */
-    public function skipDuplicates(callable $callback, bool $enable = true): mixed
+    public function skipDuplicates(callable $callback): mixed
     {
-        if (!$enable) {
-            return $callback();
-        }
-
         $previous = $this->skipDuplicates;
         $this->skipDuplicates = true;
 
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -33,6 +33,29 @@ abstract class Adapter
 
     protected bool $alterLocks = false;
 
+    protected bool $skipDuplicates = false;
+
+    /**
+     * Run a callback with skipDuplicates enabled.
+     * Duplicate key errors during createDocuments() will be silently skipped
+     * instead of thrown. Nestable — saves and restores previous state.
+     *
+     * @template T
+     * @param callable(): T $callback
+     * @return T
+     */
+    public function skipDuplicates(callable $callback): mixed
+    {
+        $previous = $this->skipDuplicates;
+        $this->skipDuplicates = true;
+
+        try {
+            return $callback();
+        } finally {
+            $this->skipDuplicates = $previous;
+        }
+    }
+
     /**
      * @var array<string, mixed>
      */
@@ -729,13 +752,12 @@ abstract class Adapter
      *
      * @param Document $collection
      * @param array<Document> $documents
-     * @param bool $ignore If true, silently ignore duplicate documents instead of throwing
      *
      * @return array<Document>
      *
      * @throws DatabaseException
      */
-    abstract public function createDocuments(Document $collection, array $documents, bool $ignore = false): array;
+    abstract public function createDocuments(Document $collection, array $documents): array;
 
     /**
      * Update Document
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -729,12 +729,13 @@ abstract class Adapter
      *
      * @param Document $collection
      * @param array<Document> $documents
+     * @param bool $ignore If true, silently ignore duplicate documents instead of throwing
      *
      * @return array<Document>
      *
      * @throws DatabaseException
      */
-    abstract public function createDocuments(Document $collection, array $documents): array;
+    abstract public function createDocuments(Document $collection, array $documents, bool $ignore = false): array;
 
     /**
      * Update Document
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -36,27 +36,6 @@ abstract class Adapter
     protected bool $skipDuplicates = false;
 
     /**
-     * Run a callback with skipDuplicates enabled.
-     * Duplicate key errors during createDocuments() will be silently skipped
-     * instead of thrown. Nestable — saves and restores previous state.
-     *
-     * @template T
-     * @param callable(): T $callback
-     * @return T
-     */
-    public function skipDuplicates(callable $callback): mixed
-    {
-        $previous = $this->skipDuplicates;
-        $this->skipDuplicates = true;
-
-        try {
-            return $callback();
-        } finally {
-            $this->skipDuplicates = $previous;
-        }
-    }
-
-    /**
      * @var array<string, mixed>
      */
     protected array $debug = [];
@@ -413,6 +392,27 @@ abstract class Adapter
     public function inTransaction(): bool
     {
         return $this->inTransaction > 0;
+    }
+
+    /**
+     * Run a callback with skipDuplicates enabled.
+     * Duplicate key errors during createDocuments() will be silently skipped
+     * instead of thrown. Nestable — saves and restores previous state.
+     *
+     * @template T
+     * @param callable(): T $callback
+     * @return T
+     */
+    public function skipDuplicates(callable $callback): mixed
+    {
+        $previous = $this->skipDuplicates;
+        $this->skipDuplicates = true;
+
+        try {
+            return $callback();
+        } finally {
+            $this->skipDuplicates = $previous;
+        }
     }
 
     /**
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -39,13 +39,20 @@ abstract class Adapter
      * Run a callback with skipDuplicates enabled.
      * Duplicate key errors during createDocuments() will be silently skipped
      * instead of thrown. Nestable — saves and restores previous state.
+     * Pass $enable = false to run the callback without toggling the flag
+     * (useful for conditional forwarding from wrappers like Pool/Mirror).
      *
      * @template T
      * @param callable(): T $callback
+     * @param bool $enable
      * @return T
      */
-    public function skipDuplicates(callable $callback): mixed
+    public function skipDuplicates(callable $callback, bool $enable = true): mixed
     {
+        if (!$enable) {
+            return $callback();
+        }
+
         $previous = $this->skipDuplicates;
         $this->skipDuplicates = true;
 
--- a/src/Database/Adapter/Mongo.php
+++ b/src/Database/Adapter/Mongo.php
@@ -1511,8 +1511,7 @@ class Mongo extends Adapter
             }
 
             $operations = [];
-            $opIndexToDocIndex = [];
-            foreach ($records as $i => $record) {
+            foreach ($records as $record) {
                 $filter = ['_uid' => $record['_uid'] ?? ''];
                 if ($this->sharedTables) {
                     $filter['_tenant'] = $record['_tenant'] ?? $this->getTenant();
@@ -1527,7 +1526,6 @@ class Mongo extends Adapter
                     continue;
                 }
 
-                $opIndexToDocIndex[\count($operations)] = $i;
                 $operations[] = [
                     'filter' => $filter,
                     'update' => ['$setOnInsert' => $setOnInsert],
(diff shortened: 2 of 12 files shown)


src/Database/Adapter/SQL.php
```diff
@@ -2573,8 +2647,9 @@ public function createDocuments(Document $collection, array $documents): array
             $batchKeys = \implode(', ', $batchKeys);
 
             $stmt = $this->getPDO()->prepare("
-                INSERT INTO {$this->getSQLTable($name)} {$columns}
+                {$this->getInsertKeyword($ignore)} {$this->getSQLTable($name)} {$columns}
                 VALUES {$batchKeys}
+                {$this->getInsertSuffix($ignore, $name)}
```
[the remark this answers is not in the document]
premtsd-code: documents are validated by StructureValidator before reaching adapter

src/Database/Adapter/SQL.php
```diff
@@ -2583,13 +2660,111 @@ public function createDocuments(Document $collection, array $documents): array
 
             $this->execute($stmt);
```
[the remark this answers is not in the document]
premtsd-code: StructureValidator will handle the required NOT NULL columns

src/Database/Adapter/Mongo.php
```diff
@@ -1492,6 +1497,42 @@ public function createDocuments(Document $collection, array $documents): array
...
+                ];
+            }
+
+            try {
+                $this->client->upsert($name, $operations, $options);
+            } catch (MongoException $e) {
+                throw $this->processException($e);
+            }
+
+            return $documents;
```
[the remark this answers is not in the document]
premtsd-code: Client::upsertWithCounts()" line you're quoting is from the old PR description — that path was removed in a later commit and the body has now been updated.

Current behavior: upsert + $setOnInsert is atomic and doesn't abort the transaction the way insertMany would on a duplicate, so the race-correctness case is handled

src/Database/Adapter.php
```diff
@@ -33,6 +33,36 @@ abstract class Adapter
...
+     * instead of thrown. Nestable — saves and restores previous state.
+     * Pass $enable = false to run the callback without toggling the flag
+     * (useful for conditional forwarding from wrappers like Pool/Mirror).
+     *
+     * @template T
+     * @param callable(): T $callback
+     * @param bool $enable
+     * @return T
+     */
+    public function skipDuplicates(callable $callback, bool $enable = true): mixed
```
abnegate: Let's remove the $enable flag and check `if skipDuplicates` instead, otherwise we add the overhead of calling an extra callable when it's false

src/Database/Database.php
```diff
@@ -842,6 +844,99 @@ public function skipRelationshipsExistCheck(callable $callback): mixed
...
+    /**
+     * Batch-fetch documents by ID from a collection, keyed by tenantKey().
+     * Chunks by maxQueryValues and groups queries by tenant in tenant-per-document mode.
+     *
+     * @param Document $collection
+     * @param array<Document> $documents Source documents (only IDs are read)
+     * @param bool $idsOnly When true, uses Query::select(['$id']) for a lighter fetch
+     * @return array<string, Document>
+     */
+    private function fetchExistingByIds(Document $collection, array $documents, bool $idsOnly = false): array
```
abnegate: We can use `find`

src/Database/Database.php
```diff
@@ -5653,6 +5748,32 @@ public function createDocuments(
...
+
+            $preExistingIds = $this->fetchExistingByIds($collection, $documents, idsOnly: true);
+            if (!empty($preExistingIds)) {
+                $documents = \array_values(\array_filter(
+                    $documents,
+                    fn (Document $doc) => !isset($preExistingIds[$this->tenantKey($doc)])
+                ));
+            }
+        }
+
```
abnegate: I don't think we need to do this if using insert ignore

src/Database/Mirror.php
```diff
@@ -601,12 +601,26 @@ public function createDocuments(
...
+        // Forward only docs the source actually persisted, so skipDuplicates no-ops
+        // on the source don't inject would-be values into the destination.
+        /** @var array<Document> $insertedFromSource */
+        $insertedFromSource = [];
+        $captureOnNext = function (Document $doc) use (&$insertedFromSource, $onNext): void {
+            $insertedFromSource[] = $doc;
+            if ($onNext !== null) {
+                $onNext($doc);
+            }
+        };
```
abnegate: We can never capture like this internally, if there were many docs, this loop could easily OOM

src/Database/Adapter/SQL.php
```diff
@@ -2608,6 +2611,33 @@ public function createDocuments(Document $collection, array $documents): array
...
+
+    /**
+     * Returns a suffix for the permissions INSERT statement when ignoring duplicates.
+     * Override in adapter subclasses for DB-specific syntax.
+     */
+    protected function getInsertPermissionsSuffix(): string
+    {
+        return '';
+    }
+
```
abnegate: Let's group logically with other getter methods

src/Database/Database.php
premtsd-code: @abnegate The child collection is getting orphan rows. when multiple rows with same parent and different childs. where parent is ignored but child is not ignored

https://github.com/utopia-php/database/blob/41704eb1c90c9ce08c698e24e7774e4122c25e48/src/Database/Database.php#L5765-L5767

src/Database/Database.php
```diff
@@ -7116,18 +7143,57 @@ public function upsertDocumentsWithIncrease(
...
+                    $idsByTenant[$doc->getTenant()][] = $doc->getId();
+                }
+            }
+            foreach ($idsByTenant as $tenant => $tenantIds) {
+                $tenantIds = \array_values(\array_unique($tenantIds));
+                foreach (\array_chunk($tenantIds, \max(1, $this->maxQueryValues)) as $chunk) {
+                    $found = $this->authorization->skip(fn () => $this->withTenant($tenant, fn () => $this->silent(
+                        fn () => $this->find($collection->getId(), [
+                            Query::equal('$id', $chunk),
+                            Query::limit(PHP_INT_MAX),
```
abnegate: If the chunk size is always maxQueryValues, might as well use that here too

src/Database/Database.php
```diff
@@ -7116,18 +7143,57 @@ public function upsertDocumentsWithIncrease(
...
+                        ])
+                    )));
+                    foreach ($found as $doc) {
+                        $existingDocs[$tenant . ':' . $doc->getId()] = $doc;
```
abnegate: We can use the tenantKey function here

src/Database/Database.php
```diff
@@ -7116,18 +7143,57 @@ public function upsertDocumentsWithIncrease(
...
+                \array_map(fn (Document $doc) => $doc->getId(), $documents),
+                fn ($id) => $id !== ''
+            )));
+
+            if (!empty($docIds)) {
+                foreach (\array_chunk($docIds, \max(1, $this->maxQueryValues)) as $chunk) {
+                    $existing = $this->authorization->skip(fn () => $this->silent(
+                        fn () => $this->find($collection->getId(), [
+                            Query::equal('$id', $chunk),
+                            Query::limit(PHP_INT_MAX),
```
abnegate: Same here

src/Database/Mirror.php
```diff
@@ -601,13 +601,36 @@ public function createDocuments(
...
-        $modified = $this->source->createDocuments(
-            $collection,
-            $documents,
-            $batchSize,
-            $onNext,
-            $onError,
-        );
+        // In skipDuplicates mode, identify which input ids already exist on source.
+        // These will be silently no-oped by the adapter's INSERT IGNORE and must
+        // not propagate to destination — a skipped duplicate is not a user write.
```
abnegate: Same as earlier - we should still propagate duplicates to destination, they might not be duplicate there

src/Database/Database.php
```diff
@@ -7116,18 +7143,57 @@ public function upsertDocumentsWithIncrease(
...
+                            Query::equal('$id', $chunk),
+                            Query::limit($this->maxQueryValues),
+                        ])
+                    ));
+                    foreach ($existing as $doc) {
+                        $existingDocs[$this->tenantKey($doc)] = $doc;
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
premtsd-code: skipPreRead is not merged in main branch yet

abnegate (approved):

PR #859, opened 2026-04-20
feat: add Query::fingerprint() for shape-only query hashing

lohanidamodar:
## Summary
Adds `Query::fingerprint(array $queries): string` — a static helper that computes a deterministic hash of query structure (method + attribute) with values excluded.

## Why
Useful for grouping queries by pattern — for example, slow-query analytics where two queries with the same shape but different parameter values should count as the same pattern. Two queries like `equal("name", "Alice")` and `equal("name", "Bob")` produce the same fingerprint.

## Behavior
- Accepts raw query strings or already-parsed `Query` objects
- Order-independent (sorts shapes before hashing)
- Returns md5 hash of canonical `method:attribute|method:attribute|...` string
- Empty array returns a deterministic hash (`md5('')`)

## Test plan
Added `testFingerprint()` in `tests/unit/QueryTest.php` covering:
- Same shape different values → same hash
- Different attribute → different hash
- Different method → different hash
- Order-independence
- Accepts parsed `Query` objects
- Empty array behavior

All 5 tests pass locally (257 assertions).

## Summary by CodeRabbit

* **New Features**
  * Added query fingerprinting to generate consistent identifiers for queries with the same structure, independent of literal values and array order; accepts serialized or structured queries and returns a deterministic hash (empty input yields a stable result).
* **Tests**
  * Added unit tests verifying shape-only equality, order-independence, logical grouping behavior, input formats, and invalid-input error handling.

### DIFF
--- a/src/Database/Query.php
+++ b/src/Database/Query.php
@@ -420,6 +420,41 @@ class Query
     }
 
     /**
+     * Compute a shape-only fingerprint of an array of queries.
+     *
+     * The fingerprint captures the structure of the queries — method and
+     * attribute — without values. Two query sets with the same shape but
+     * different parameter values produce the same fingerprint, which is
+     * useful for pattern-based counting and slow-query grouping.
+     *
+     * Accepts either raw query strings or parsed Query objects.
+     *
+     * @param array<string|Query> $queries
+     * @return string md5 hash of the canonical shape
+     * @throws QueryException
+     */
+    public static function fingerprint(array $queries): string
+    {
+        $shapes = [];
+
+        foreach ($queries as $query) {
+            if (\is_string($query)) {
+                $query = self::parse($query);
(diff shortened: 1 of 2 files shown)


src/Database/Query.php
```diff
@@ -419,6 +419,70 @@ public static function parseQueries(array $queries): array
...
+        return \md5(\implode('|', $shapes));
+    }
+
+    /**
+     * Canonical shape string for a single Query — recursive for logical types.
+     *
+     * @param Query $query
+     * @return string
+     */
+    private static function queryShape(self $query): string
```
abnegate: If we need an instance anyway, lets make it an instance method `shape`, with a loop instead of recursion

abnegate (approved):

PR #878, opened 2026-05-15
test

premtsd-code:
## Summary

Adds a per-instance in-process cache for collection metadata in `Database::getCollection()`. Today, every authenticated read path calls `getCollection()` many times — directly and through internal helpers like `convertQuery`, validators, and relationship resolution — and each call goes through `getDocument(METADATA, $id)` which incurs a Redis round trip even when the metadata is unchanged.

For a relationship-heavy request that touches one collection 50 times via cascade resolution, that's currently 50 redundant Redis reads. With this change it becomes 1 read, with the remaining 49 served from in-process memory.

## How invalidation works

We don't add new invalidation call sites. The library already fires `EVENT_DOCUMENT_PURGE` with `$collection=METADATA` after every metadata mutation (createAttribute, deleteAttribute, createIndex, deleteRelationship, updateCollection, etc.). A new listener registered in the constructor catches those events and drops the corresponding cache entry.

For Appwrite Cloud, this also stays correct across regions for free: `cloud/app/controllers/general.php` already listens on the same event and propagates purges via `RegionManager`. The receiving region's `purgeCachedDocument(METADATA, $id)` call re-fires the event on its own Database instance, which invalidates the in-process cache there too.

Full cache wipes are added to the four context-changing setters (`setNamespace`, `setDatabase`, `setSharedTables`, `setTenant`) and `delete()`, since those alter the cache key shape or invalidate the underlying data wholesale.

## Memory safety

The cache stores the array form of each collection, not the `Document` instance. The recursive `arrayifyDocuments` helper unwraps nested attribute/index `Document` references too. On every cache hit, a fresh `Document` is rebuilt from the stored array — so callers can mutate the returned collection (e.g. `setAttribute('attributes', $new, SET_TYPE_APPEND)` as `createAttribute` does) without any risk of poisoning the cache.

## Cache key shape

`{database}:{namespace}:{tenant}:{collectionId}` — multi-tenant safe. `withTenant()` doesn't need a clear because the tenant is part of the key.

## Predicted impact

- 15-25% p99 latency reduction on relationship-heavy reads (listMemberships, listDocuments with selects, cascade fetches)
- 20-40% Dragonfly/Redis cache command rate reduction
- No MySQL impact (this is a cache-layer fix, not a query-layer one)
- Per-pod memory: ~5MB at steady state for a typical project (typically dozens of collections)

## Test plan

- [ ] Existing PHPUnit suite passes
- [ ] Manual: call `getCollection('foo')` → mutate the result → call again, verify second call returns the original (isolation test)
- [ ] Manual: call `getCollection('foo')` → `createAttribute('foo', ...)` → call `getCollection('foo')`, verify new attribute appears (invalidation test)
- [ ] Manual: under `withTenant($t1)` populate cache, switch to `withTenant($t2)`, verify isolation
- [ ] Staging: confirm Dragonfly command rate drops on canary region post-deploy

## Summary by CodeRabbit

* **Refactor**
  * Database operations involving relationships now execute more efficiently through optimized metadata handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/utopia-php/database/pull/878)

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -398,6 +398,18 @@ class Database
      */
     protected ?array $silentListeners = [];
 
+    /**
+     * In-process cache of collection metadata, keyed by
+     * "database:namespace:tenant:collectionId". Stores the raw array form so
+     * callers always get a fresh Document and cannot corrupt the cache.
+     * Invalidated via the EVENT_DOCUMENT_PURGE listener registered in the
+     * constructor, and fully cleared on context changes (setNamespace,
+     * setDatabase, setSharedTables, setTenant, delete).
+     *
+     * @var array<string, array<string, mixed>>
+     */
+    protected array $collectionCache = [];
+
     protected ?\DateTime $timestamp = null;
 
     protected bool $resolveRelationships = true;
@@ -701,6 +713,21 @@ class Database
                 return is_array($decoded) ? $decoded : $value;
             }
         );
+
+        // Keep the in-process collection cache in sync with the existing
+        // EVENT_DOCUMENT_PURGE plumbing. Every metadata mutation already
+        // fires this event with $collection=METADATA; cross-region purges
+        // (cloud) also flow through here, so this listener stays correct
+        // without changes to consumers.
+        $this->on(
+            self::EVENT_DOCUMENT_PURGE,
+            'collection-cache-invalidator',
+            function (string $event, Document $payload): void {
+                if ($payload->getCollection() === self::METADATA) {
+                    $this->invalidateCollectionCache($payload->getId());
+                }
+            }
+        );
     }
 
     /**
@@ -930,6 +957,7 @@ class Database
     public function setNamespace(string $namespace): static
     {
         $this->adapter->setNamespace($namespace);
+        $this->clearCollectionCache();
 
         return $this;
     }
@@ -969,6 +997,7 @@ class Database
     public function setDatabase(string $name): static
     {
         $this->adapter->setDatabase($name);
+        $this->clearCollectionCache();
 
         return $this;
     }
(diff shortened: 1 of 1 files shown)


premtsd-code:
Closing this in favor of keeping the existing selective caching pattern. Captured here so the trade-off is recorded for future reference.

## Why close

The existing utopia/database cache (Dragonfly via utopia/cache) is selective by design: it caches documents that are repeatedly read with the same key (`getDocument` hits), and explicitly does not cache things that vary per-query (`find`/`count`/`sum` results, `_perms` lookups). This PR violated that principle by caching **every** collection touched in a request, indiscriminately — including collections that are accessed once and never re-read.

Per-request lifetime bounds the memory growth, so it isn't catastrophic. But the principle "cache only what benefits from being cached" matters more than the few-ms p99 win this PR predicted.

## What we don't have evidence for

The PR's predicted impact (15-25% p99 reduction on relationship-heavy reads, 20-40% Dragonfly load reduction) was based on an estimate of ~30-50 `getCollection` calls per request via cascade resolution. That estimate was never measured. PR #877 (the EXISTS rewrite) has measured production evidence (~600 rows/SELECT baseline); this PR doesn't.

## When to revisit

If after PR #877 lands in staging and the rows-per-SELECT ratio drops as predicted, MySQL CPU should improve materially. If at that point Dragonfly cache load is still substantial or relationship-heavy endpoint p99 remains above target, revisit with a **selective** L1: only cache collections accessed ≥2 times in the current request, so we match the existing pattern of "cache only what's getting re-read."

For now, the EXISTS rewrite in #877 is the higher-confidence, higher-impact change to ship first.

PR #896, opened 2026-06-18
refactor(transaction): own connection recovery in the adapter, drop Swoole PDOProxy dependency

abnegate:
> Stacked on #895. Diff against `fix/start-transaction-desynced-rollback`.

## Problem

#895 stops the *symptom* of the nyc3 write outage, but the root cause is upstream of this library: consumers wrap connections in `Swoole\Database\PDOProxy`, which keeps **its own** `inTransaction` counter that is incremented on `beginTransaction`, decremented only on a *successful* commit/rollback, not reset by `reconnect()`, and not reset on pool checkin (`utopia-php/pools` never calls the proxy's `reset()` — that hook only fires under Swoole's own `PDOPool`).

A connection lost mid-transaction leaks the counter and **poisons the pooled connection**: every later `startTransaction` cleanup probe trusts the stale `getPDO()->inTransaction()`, issues a `rollBack` the real connection no longer holds, and fails with `There is no active transaction` — across all projects, until pods restart. (nyc3: ~110k errors over ~1.5h, writes only.)

This can't be fixed by patching the cleanup probe while a second, unreconciled source of truth for "am I in a transaction" lives in the proxy. The fix removes the *need* for the proxy and makes the adapter own connection-loss recovery, with the real `\PDO` as the single source of truth.

## What changed

- **`PDOStatement`** (new) — wraps prepared statements. On a lost connection at execution time, when **not** in a transaction, it reconnects the owning `PDO`, re-prepares against the fresh connection, replays bound params/attributes/options, and retries. This restores the only genuinely load-bearing behaviour the Swoole `PDOStatementProxy` provided — reconnect-on-execute — and it's what heals a stale pooled connection at the `prepare('ROLLBACK')->execute()` probe that fronts `startTransaction` (that probe runs at `inTransaction === 0`). Inside a transaction it rethrows; the connection state is read from the real `\PDO`, so there is no separate counter to desync.
- **`PDO::prepare()`** returns the wrapper; **`prepareNative()`** re-prepares a raw `\PDOStatement` for the wrapper. `ERRMODE_EXCEPTION` is defaulted in the constructor so the library no longer depends on the proxy/consumer to set it.
- Replaced the `Swoole\Database\PDOStatementProxy` type hints in `SQL`/`Postgres`/`SQLite` with the new `PDOStatement`.

Connection-level calls (`beginTransaction`/`rollBack`/`commit`/`exec`) are already covered by the pre-existing `Utopia\Database\PDO::__call` reconnect, and `Adapter::withTransaction`'s pre-existing retry loop replays the closure — so the mid-transaction-death path recovers by replay on the connection that `__call` reconnects. No change to `withTransaction` is needed.

## Reviewed

Two independent reviews of this PR led to: **removing a speculative `withTransaction` reconnect block** that an earlier revision added — it was unreachable on the real SQL path (a dead-connection `rollBack` throws and is caught earlier in the loop) and, when reached (e.g. a *nested* transaction), would have discarded the outer `BEGIN` and committed the inner write standalone (silent partial commit). Recovery is correctly carried by `PDOStatement` re-prepare + `PDO::__call`, both unit-covered. Also threaded `prepare()` `$options` through re-prepare and tightened a `@throws`.

**Known limitation (pre-existing, out of scope):** connection loss *inside a nested* transaction is not safely recoverable by the retry loop in any version; this PR does not make it worse (it no longer adds a reconnect there).

## Tests

`PDOStatementTest` (reconnect/re-prepare/replay outside a txn; rethrow inside; non-connection rethrow; forwarding), `PDOTest` (`PDO::__call` reconnect, wrapper return), `SQLTransactionTest` (desynced-cleanup recovery, MySQL + Postgres). `phpunit` (full unit suite), `phpstan --level 7`, `pint` green locally.

**Recommended follow-up:** an e2e test that severs a live MySQL/Postgres connection mid-transaction (`KILL` / `wait_timeout`) and asserts recovery — the recovery path is currently only unit-mocked.

## Consumer follow-ups

- cloud [appwrite-labs/cloud#4399](https://github.com/appwrite-labs/cloud/pull/4399), server-ce [appwrite/appwrite#12616](https://github.com/appwrite/appwrite/pull/12616) — stop wrapping in `PDOProxy`, construct `Utopia\Database\PDO` directly. Re-pin from the branch alias to the release tag once this merges + tags.
- [utopia-php/pools#33](https://github.com/utopia-php/pools/pull/33) backstop (recover/destroy faulted pooled connections).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced database connection resilience with automatic reconnection and statement re-preparation when connections are lost during execution
  * Improved error handling with exceptions enabled by default for database operations

* **Internal Improvements**
  * Updated database adapters for improved compatibility and stability

### DIFF
--- a/src/Database/Adapter/Postgres.php
+++ b/src/Database/Adapter/Postgres.php
@@ -5,7 +5,6 @@ namespace Utopia\Database\Adapter;
 use Exception;
 use PDO;
 use PDOException;
-use Swoole\Database\PDOStatementProxy;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
 use Utopia\Database\Exception as DatabaseException;
@@ -18,6 +17,7 @@ use Utopia\Database\Exception\Transaction as TransactionException;
 use Utopia\Database\Exception\Truncate as TruncateException;
 use Utopia\Database\Helpers\ID;
 use Utopia\Database\Operator;
+use Utopia\Database\PDOStatement;
 use Utopia\Database\Query;
 
 /**
@@ -2764,12 +2764,12 @@ class Postgres extends SQL
      * Bind operator parameters to statement
      * Override to handle PostgreSQL-specific JSON binding
      *
-     * @param \PDOStatement|PDOStatementProxy $stmt
+     * @param \PDOStatement|PDOStatement $stmt
      * @param Operator $operator
      * @param int &$bindIndex
      * @return void
      */
-    protected function bindOperatorParams(\PDOStatement|PDOStatementProxy $stmt, Operator $operator, int &$bindIndex): void
+    protected function bindOperatorParams(\PDOStatement|PDOStatement $stmt, Operator $operator, int &$bindIndex): void
     {
         $method = $operator->getMethod();
         $values = $operator->getValues();
--- a/src/Database/Adapter/SQL.php
+++ b/src/Database/Adapter/SQL.php
@@ -4,7 +4,6 @@ namespace Utopia\Database\Adapter;
 
 use Exception;
 use PDOException;
-use Swoole\Database\PDOStatementProxy;
 use Utopia\Database\Adapter;
 use Utopia\Database\Change;
 use Utopia\Database\Database;
@@ -16,6 +15,7 @@ use Utopia\Database\Exception\NotFound as NotFoundException;
 use Utopia\Database\Exception\Timeout as TimeoutException;
 use Utopia\Database\Exception\Transaction as TransactionException;
 use Utopia\Database\Operator;
+use Utopia\Database\PDOStatement;
 use Utopia\Database\Query;
 
 abstract class SQL extends Adapter
@@ -1978,12 +1978,12 @@ abstract class SQL extends Adapter
     /**
      * Bind operator parameters to prepared statement
      *
-     * @param \PDOStatement|PDOStatementProxy $stmt
+     * @param \PDOStatement|PDOStatement $stmt
      * @param \Utopia\Database\Operator $operator
      * @param int &$bindIndex
      * @return void
      */
-    protected function bindOperatorParams(\PDOStatement|PDOStatementProxy $stmt, Operator $operator, int &$bindIndex): void
+    protected function bindOperatorParams(\PDOStatement|PDOStatement $stmt, Operator $operator, int &$bindIndex): void
     {
         $method = $operator->getMethod();
         $values = $operator->getValues();
--- a/src/Database/Adapter/SQLite.php
+++ b/src/Database/Adapter/SQLite.php
@@ -5,7 +5,6 @@ namespace Utopia\Database\Adapter;
 use Exception;
 use PDO;
 use PDOException;
-use Swoole\Database\PDOStatementProxy;
 use Utopia\Database\Database;
 use Utopia\Database\Document;
 use Utopia\Database\Exception as DatabaseException;
@@ -18,6 +17,7 @@ use Utopia\Database\Exception\Transaction as TransactionException;
 use Utopia\Database\Exception\Truncate as TruncateException;
 use Utopia\Database\Helpers\ID;
 use Utopia\Database\Operator;
+use Utopia\Database\PDOStatement;
 use Utopia\Database\Query;
 
 /**
@@ -2002,12 +2002,12 @@ class SQLite extends MariaDB
      * Bind operator parameters to statement
      * Override to handle SQLite-specific operator bindings
      *
-     * @param \PDOStatement|PDOStatementProxy $stmt
+     * @param \PDOStatement|PDOStatement $stmt
      * @param Operator $operator
      * @param int &$bindIndex
      * @return void
      */
-    protected function bindOperatorParams(\PDOStatement|PDOStatementProxy $stmt, Operator $operator, int &$bindIndex): void
+    protected function bindOperatorParams(\PDOStatement|PDOStatement $stmt, Operator $operator, int &$bindIndex): void
     {
         $method = $operator->getMethod();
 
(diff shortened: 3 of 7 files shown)


abnegate:
Thanks for the thorough pass — addressed the P1s in `682c8638`:

- **Preserve statement iteration** (`PDOStatement` not iterable): fixed. The wrapper now `implements \IteratorAggregate` with `getIterator()` delegating to the underlying statement, so `foreach ($pdo->prepare(...) as $row)` and `getDriver()` consumers iterate rows again. Regression test added.
- **Limit retry methods**: fixed. `__call` now only reconnects + replays for `execute()`. Any other method (`fetch`/`fetchAll`/`rowCount`/`closeCursor`) rethrows the connection error instead of re-running against an unexecuted statement. Test added.
- **Recover prepare failures**: fixed. Restored reconnect-and-retry in `prepareNative()` for a lost connection outside a transaction, so native-prepares (where `prepare()` contacts the server) recover, matching `__call()`. No-op under emulated prepares. Test added.
- **Keep bound references**: fixed. `bindParam()` now stores the variable **by reference**, so a value changed between bind and `execute()` is what gets replayed after a reconnect.
- **Preserve omitted defaults** (`bindColumn`): fixed. We now track how many optional args were actually supplied (`func_num_args`) and forward/replay only those, so omitted ones keep PDO's real defaults instead of explicit nulls.
- **Preserve prepare options**: already addressed at HEAD — `prepare()` threads `$options` into the wrapper and `reprepare()` passes them back into `prepareNative()`. (The comment was against an earlier commit.)

**SQLite transactions reconnect** — I don't think this one is reachable, but want a second opinion before closing it: the reconnect branch requires **both** `!inTransaction()` **and** `Connection::hasError($e)`. `hasError` only matches `DetectsLostConnections`' network-DB strings (`server has gone away`, `Lost connection`, SSL/TCP/`getaddrinfo`, …) plus `Max connect timeout reached`. SQLite is file/in-memory and its errors (`database is locked`, `disk I/O error`, `database disk image is malformed`) match none of those, so `hasError` is false and the wrapper rethrows — it never reaches the reconnect path during a SQLite `BEGIN IMMEDIATE` transaction. With the "execute-only" change above it's further constrained. I deliberately avoided giving the wrapper its own transaction counter to track SQLite's raw-`BEGIN` state, since a second source of truth desyncing from the real one is exactly what caused the original outage. If you can construct a SQLite error that satisfies `hasError`, I'll add an adapter-driven transaction signal — otherwise I'll resolve this as not-reachable.

Full unit suite (365), PHPStan L7, and Pint are green.

abnegate:
On the remaining CodeRabbit items:

**Match `prepareNative()` expectation (test)** — fixed in `9bab43f0`; the mock now asserts `with('SELECT :id', [])`.

**Gate statement replay to idempotent / transaction-managed operations** — I've scoped this rather than coded it, and want to lay out why I think that's right for this PR:

- The wrapper now only retries **`execute()`** after a reconnect (the earlier "retries every method" issue is fixed), and only when **`!inTransaction()`**.
- All DML in the adapters runs inside `Database::withTransaction()` (14 call sites cover create/update/upsert/increment/decrement/delete). Inside a transaction the wrapper **rethrows** — it does not retry — so the transaction-level retry owns recovery there. That means the statement-level retry path is reached only by **reads** (idempotent) and **DDL** (which loud-fails on replay, e.g. "table already exists" — not a silent double-apply). There is no non-transactional DML that this path would silently re-apply.
- At-least-once-on-ack-loss (a disconnect after server apply but before the client ack) and loss of per-connection session state (`statement_timeout`, etc.) on reconnect are **inherent to any retry/reconnect-based recovery**. They already exist in the pre-existing `withTransaction()` retry loop and in `PDO::__call()`, and they existed in the `Swoole\Database\PDOStatementProxy` this replaces — which actually retried *every* method, so this change is strictly safer, not riskier.

So this PR doesn't introduce new exposure; it reduces it. True exactly-once writes (idempotency keys / dedup) and replaying session state across a reconnect are worthwhile hardening but are an architectural change well beyond removing the Swoole proxy. Happy to open a separate tracking issue for both rather than fold them in here.

The other two (prepare-time reconnect for native prepares, and by-reference `bindParam`/`bindColumn`) are already marked addressed in `682c863`.

abnegate:
Re-review on `682c863` surfaced one new one — **Preserve binding order** (`PDOStatement.php`): fixed in `12c3bf1d`. `bindValue()`/`bindParam()` now record call order in a single `bindOrder` list, and `reprepare()` replays value/param bindings in that exact order, so a placeholder rebound across methods replays with the binding the caller applied last (matching PDO). Regression test added (`bindValue(':id','old')` then `bindParam(':id', $new)` → replay is `value:old, param:new`).

For the record, on the **SQLite transactions reconnect** flag from the first pass (not re-raised since): I believe it's unreachable rather than fixed-in-code, and want to flag my reasoning in case you disagree. The reconnect path needs both `!inTransaction()` **and** `Connection::hasError($e)`. `hasError` only matches `DetectsLostConnections`' network-DB markers (`server has gone away`, `Lost connection`, SSL/TCP/`getaddrinfo`, `Max connect timeout reached`, …). SQLite is file/in-memory; its errors (`database is locked`, `disk I/O error`, `database disk image is malformed`) match none, so `hasError` is false and the wrapper rethrows — it never reconnects during a SQLite `BEGIN IMMEDIATE` transaction. I deliberately didn't give the wrapper its own transaction counter to track SQLite's raw-`BEGIN` state, since a second source of truth desyncing from the real one is exactly what caused the incident this PR fixes. If you can produce a SQLite error that satisfies `hasError`, I'll wire an adapter-driven transaction signal instead.

Full unit suite (366), PHPStan L7, Pint green at `12c3bf1d`.

PR #902, opened 2026-06-21
Reject stale cached snapshot after write

fogelito:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a cache coherency issue that could cause stale data to be served after document updates, resolving intermittent test failures.

* **Tests**
  * Added regression test to verify proper handling of stale cached snapshots.


fogelito:
Refactored in another PR

PR #903, opened 2026-06-22
Fence document cache fills

ChiragAgg5k:
## What

Uses the fenced cache lease API from `utopia-php/cache` to guard document cache fills in `Database::getDocument()`.

Depends on: https://github.com/utopia-php/cache/pull/74

## Why

A cache-miss read can race with a concurrent writer purge: the reader fetches an older DB row, the writer commits and purges, then the reader saves the older document back into cache. The database layer then serves that poisoned document to whole-document read-modify-write callers, which can permanently overwrite fields with stale values.

## Changes

- Acquire a cache lease before backend document reads for plain cacheable `getDocument()` calls.
- Save document cache entries through `saveLease()` so a purge between read and save makes the fill a no-op.
- Keep collection membership updates only when the fenced document save succeeds.
- Point `utopia-php/cache` at `dev-cache-lease-fence` from the cache PR branch and update the lockfile.

## Validation

- `php -l src/Database/Database.php`
- `composer validate --strict --no-check-publish`
- `php -d memory_limit=2G vendor/bin/pint --test composer.json src/Database/Database.php tests/unit/ForUpdateCacheTest.php`
- `vendor/bin/phpunit --configuration phpunit.xml tests/unit/ForUpdateCacheTest.php`
- `vendor/bin/phpstan analyse --level 7 src/Database/Database.php --memory-limit 2G`

Note: Composer update required `--ignore-platform-req=ext-pcov` locally because `ext-pcov` is not installed in this environment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,12 @@
         "check": "./vendor/bin/phpstan analyse --level 7 src tests --memory-limit 2G",
         "coverage": "./vendor/bin/coverage-check ./tmp/clover.xml 90"
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/utopia-php/cache"
+        }
+    ],
     "require": {
         "php": ">=8.4",
         "ext-pdo": "*",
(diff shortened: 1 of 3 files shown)


ChiragAgg5k:
Closing this draft because the fix now lives in utopia-php/cache as hidden cache fencing; this database branch only updates the dev dependency and no longer carries a meaningful library change.

PR #913, opened 2026-07-01
fix(mongo): honour query timeout on count() and sum() aggregate paths

abnegate:
## Problem

The Mongo adapter drops `maxTimeMS` on the `count()` and `sum()` aggregation paths, so those operations run **unbounded** even when a query timeout is set via `setTimeout()`:

- `count()` sets `$options['maxTimeMS']` early, but then **reassigns** `$options = $this->getTransactionOptions();` before building the aggregate pipeline, discarding the timeout.
- `sum()` never sets `maxTimeMS` at all.

`find()` is unaffected (it sets `maxTimeMS` on the options it actually passes), which is why this went unnoticed.

Impact: a `total=true` document list issues `count()`, so a large/slow Mongo read hangs until the client gives up instead of failing fast with a `Timeout`. Downstream (Appwrite Cloud DocumentsDB) this surfaced as a request stuck ~90s → 503, tying up a worker the whole time.

## Fix

Set `maxTimeMS` on the options actually passed to `aggregate()` in both `count()` and `sum()`, and remove the now-dead early assignment in `count()`.

## Test

`testCountTimeout` (shared adapter scope): builds a large collection, sets a 1ms timeout, and asserts `count()` throws `Timeout`. Fails without this change on Mongo (count completes unbounded); passes with it. `count()` is type-agnostic so it is safe across all timeout-supporting adapters (Mongo `maxTimeMS`, MariaDB `MAX_EXECUTION_TIME`, Postgres `statement_timeout`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling for aggregation-based count and sum operations, applying very short limits more consistently.
  * When clearing timeouts, the adapter now fully resets its internal timeout state to avoid carrying limits into later statements.
* **Tests**
  * Added an end-to-end test covering count behavior under extremely short timeouts, validating that timeout exceptions are raised when supported.

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -349,6 +349,11 @@ abstract class Adapter
     {
         // Clear existing callback
         $this->before($event, 'timeout');
+
+        // Adapters that apply the timeout from this property on every statement
+        // (e.g. Postgres SET statement_timeout) would otherwise keep enforcing a
+        // cleared timeout on all subsequent queries.
+        $this->timeout = 0;
     }
 
     /**
--- a/src/Database/Adapter/Mongo.php
+++ b/src/Database/Adapter/Mongo.php
@@ -2718,10 +2718,6 @@ class Mongo extends Adapter
             $options['limit'] = $max;
         }
 
-        if ($this->timeout) {
-            $options['maxTimeMS'] = $this->timeout;
-        }
-
         // Build filters from queries
         $filters = $this->buildFilters($queries);
 
@@ -2745,6 +2741,11 @@ class Mongo extends Adapter
          **/
 
         $options = $this->getTransactionOptions();
+
+        if ($this->timeout) {
+            $options['maxTimeMS'] = $this->timeout;
+        }
+
         $pipeline = [];
 
         // Add match stage if filters are provided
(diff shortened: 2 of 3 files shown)


abnegate:
All review comments addressed in `faeb5bf4`, plus the root causes of the adapter-test failures:

- **count() swallowing timeouts (Copilot)** — `catch (MongoException) { return 0; }` now routes through `processException()` and rethrows `Timeout`; other errors keep the lenient return 0.
- **sum() raw client exception (Copilot)** — wrapped in `processException()` so timeouts map to `Timeout`.
- **Dead `$options['limit']` in count() (Copilot)** — removed; the limit is applied via the `$limit` pipeline stage.
- **Cleanup only in catch (greptile P2)** — `clearTimeout`/`deleteCollection` moved to `finally`, and the assertion moved out of the catch (previously a swallowed timeout made `fail()` throw inside the try and get asserted as the exception).
- **`file_get_contents` per iteration (Copilot)** — fixture read once.

The CI failures had two distinct causes, both real bugs this PR now fixes:
1. **Mongo jobs**: the count timeout was swallowed (returned 0), so the test's `fail()` fired.
2. **Postgres jobs**: `Adapter::clearTimeout()` never reset the `timeout` property, and Postgres applies that property on every statement (`SET statement_timeout`) — so `testQueryTimeout`'s 1ms timeout leaked into every subsequent statement on the connection, and the new test's heavy inserts were the first thing slow enough to die. `clearTimeout()` now resets the property.

PR #914, opened 2026-07-01
fix: retry deadlocks with exponential backoff and jitter

claudear:
## Summary
- Deadlock errors (`SQLSTATE[40001]`) in `withTransaction` now get 5 retries (up from 2) with exponential backoff and random jitter
- Non-deadlock errors retain the existing 2-retry behavior
- Adds `isDeadlock()` helper to detect deadlock exceptions by SQLSTATE code and error message
- Adds `retrySleep()` helper for exponential backoff with jitter to avoid thundering-herd re-collisions

Fixes: https://appwrite.sentry.io/issues/7430260127/

## Context
Under high contention (e.g. concurrent stats workers updating the same project), MariaDB deadlocks (`SQLSTATE[40001]: Serialization failure: 1213`) were exhausting the 2-retry limit in `Adapter::withTransaction()`, causing unrecoverable failures. The linear backoff (50ms, 100ms) also made it likely for retries to re-collide.

The fix uses exponential backoff (50ms, 100ms, 200ms, 400ms, 800ms) with random jitter (0-100% of delay) to spread retries and increase the chance of resolution.

## Test plan
- [x] Added unit test verifying deadlocks get 6 total attempts (initial + 5 retries), more than generic errors (3 attempts)
- [x] Added unit test verifying transient deadlocks resolve successfully on retry
- [x] Existing unit tests pass with no regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction retry handling to better distinguish deadlocks from other transient database errors.
  * Added smarter retry timing with backoff and jitter for more reliable recovery under contention.
  * Made transaction startup more resilient when rollback cleanup is out of sync, while still surfacing real begin failures.
* **Tests**
  * Added coverage for deadlock retries, transient recovery, and rollback/startup edge cases across SQL and Postgres adapters.

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -402,8 +402,9 @@ abstract class Adapter
     {
         $sleep = 50_000; // 50 milliseconds
         $retries = 2;
+        $deadlockRetries = 5;
 
-        for ($attempts = 0; $attempts <= $retries; $attempts++) {
+        for ($attempts = 0; $attempts <= $deadlockRetries; $attempts++) {
             try {
                 $this->startTransaction();
                 $result = $callback();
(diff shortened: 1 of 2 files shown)


claudear:
## Fix Confidence: 82/100

The fix correctly identifies deadlock exceptions and provides significantly more retries with better backoff strategy. The exponential backoff with jitter is the standard approach for deadlock resolution. However, confidence is not higher because: (1) we can't fully test real concurrent deadlock scenarios in unit tests, (2) extremely high contention could still exhaust 5 retries, and (3) the CI checks haven't completed yet on the PR.

PR #920, opened 2026-07-16
fix(database): guard getDocument() against by-id cache/adapter identity mismatch

abnegate:
Found on Appwrite Cloud prod: `GET /v1/mysql/{id}` served a **different** database's document (wrong `$id`, hostname, port) for ~90s while a sibling row was being written concurrently, then self-healed. The list endpoint stayed correct the whole time — proving the underlying row was never wrong, only the by-id **cache entry**.

## Root cause
`Database::getDocument()` is cache-aside and saves the adapter's returned body under a key derived from the **requested** id, with no assertion that `body.$id === requestedId` (read at `Database.php` ~`:4916`, cached via `saveWithLease` ~`:4969`). If the adapter ever returns a foreign row (observed under a concurrent-write / pooled-connection result mixup), that body is cached under the wrong key and served sticky until the entry is invalidated.

## Fix
New `getDocument()` identity guard via `isForeignDocument($document, $id)` (`$actual !== '' && $actual !== $id`):
- **Cached** entry whose body is foreign → log + `purgeCachedDocument` + refetch (self-heal).
- **Adapter** read whose row is foreign → **throw** `DatabaseException` (fail-closed) — never cache or return it.
- A partial `select()` that omits `$id` is **not** treated as foreign (no false positives on projections).

### Why fail-closed (throw) rather than degrade
Cache keys are namespaced per project, so a poisoned *cache* entry is project-bounded. But the connection-level mixup that can seed it is **not** inherently bounded in a shared multi-tenant shard-pool deployment — a result mixup could copy another tenant's row into the requester's keyspace. Serving another tenant's hostname/credentials must never happen, so a foreign adapter row that survives a refetch raises loudly instead of being returned.

## Tests
`tests/unit/ForeignDocumentGuardTest.php` (3 tests; 2 fail without the guard) — covers the cached-foreign self-heal, the adapter-foreign throw, and the partial-`select()` no-false-positive case. Pint + PHPStan (L7) clean; neighbouring cache/document suites green.

## Consumer
Appwrite Cloud PR appwrite-labs/cloud#4827 adds a `Store`-level `databaseId === requestedId` back-reference guard on top of this, and pins this branch via a dev VCS ref pending a tagged release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved document retrieval safeguards to prevent returning a document under the wrong ID.
  * Automatically discards corrupted cached entries and refetches the correct document.
  * Raises an error when the data source returns a document with an unexpected ID.
  * Preserved normal behavior for valid cached documents.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -4883,23 +4883,33 @@ class Database
             // compare equal under strict equality (e.g. in updateDocument).
             $document = $this->casting($collection, $document);
 
-            if ($collection->getId() !== self::METADATA) {
+            // Defend the by-id read invariant: a cache entry whose body carries a
+            // different $id than requested is a poisoned key (a mismatched read
+            // was persisted under this key). Serving it returns the wrong document
+            // for every reader until the entry expires. Purge it and fall through
+            // to the source read instead of trusting the poisoned copy.
+            if ($this->isForeignDocument($document, $id)) {
+                Console::error("Cache for '{$collection->getId()}:{$id}' held foreign document '{$document->getId()}'; purging and refetching");
+                $this->purgeCachedDocument($collection->getId(), $id);
+            } else {
+                if ($collection->getId() !== self::METADATA) {
 
-                if (!$this->authorization->isValid(new Input(self::PERMISSION_READ, [
-                    ...$collection->getRead(),
-                    ...($documentSecurity ? $document->getRead() : [])
-                ]))) {
+                    if (!$this->authorization->isValid(new Input(self::PERMISSION_READ, [
+                        ...$collection->getRead(),
+                        ...($documentSecurity ? $document->getRead() : [])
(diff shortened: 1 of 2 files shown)


abnegate:
Closing — after review this guard is defense-in-depth on the `getDocument()` by-id path only, not a root-cause fix.

The wrong-document read originates upstream: a query for document A returned document B's row (a pooled-connection / result-set mixup, PDOProxy-class). That same mixup can corrupt `find()`/query results too, which a getDocument()-only identity check does **not** cover — so this PR is partial insurance, not a fix.

The sensitive path (dedicated-DB by-id reads that can expose a foreign hostname/credentials) is already guarded in a targeted way in the consumer (appwrite cloud Store layer). Pursuing the connection-layer root cause instead of a hot-path assertion in core utopia.

PR #921, opened 2026-07-16
fix(database): drop poisoned pooled connection on any lost-connection error (DAT-1904)

abnegate:
## Problem (DAT-1904)

A pooled DB connection is checked out per operation from a worker-shared, per-host pool. The reconnect wrapper in `PDOStatement::__call()` only self-healed a lost connection for the **`execute()`-outside-a-transaction** case. Every other lost-connection error was rethrown while the underlying PDO connection was left untouched and reclaimed to the pool as-is:

- `fetch()` / `fetchAll()` / `closeCursor()` (the fetch phase of every read, e.g. `SQL::getDocument()` at `execute()` → `fetchAll()` → `closeCursor()`, and `SQL::find()`)
- `execute()` **inside** a transaction

A connection lost mid-result is **wire-desynced**: a partial or stale result frame can remain buffered on the socket (classically a client-side read timeout while the server keeps streaming the result set). Returning that connection to the pool poisons it.

## Blast radius

The next coroutine to pop the poisoned connection runs *its own* query and reads the **leftover frame** first:

- **Reads**: returns a different document than requested — cross-document, and **cross-tenant on shared shards** (the leftover rows belong to whoever ran the failed query).
- **Writes**: the desynced protocol state produces the "no active transaction" class of failures (see utopia-php/database#895, #896) and can corrupt transactional writes.

Because `Adapter::withTransaction()` replays its callback up to 2× and calls `getPDO()` afresh on each attempt, a reclaimed poisoned connection can be handed straight back into the replay.

## Fix

Extend the recovery in `PDOStatement::__call()` so that on **any** detected lost-connection error (same detection as today, `Connection::hasError()` / `DetectsLostConnections`) that is *not* the `execute()`-outside-transaction fast path, we call `PDO::reconnect()` to **discard the wire-desynced socket before rethrowing**. This guarantees the connection reclaimed to the pool is always clean.

- The existing transparent `execute()`-outside-transaction retry (reprepare + replay bindings + re-execute) is **unchanged**.
- Non-connection errors (syntax errors, constraint violations, business logic) still rethrow **untouched** — a healthy connection is never thrashed on ordinary query errors.
- Callers / `withTransaction` still see the rethrow and roll back / replay on the now-fresh connection. In the server-genuinely-down case, `reconnect()` may itself throw `Connection refused`, which is already in the lost-connection needle set, so callers classify it identically.

### Why the fetch phase must *reconnect* rather than retry

A fetch cannot be retried in place: re-running `fetch()` on a freshly prepared statement would read no rows. The only safe recovery is to drop the desynced socket and let the caller replay the whole operation on a clean connection.

## Tests

`tests/unit/PDOStatementTest.php` (mock-based, no live DB):

- **`testFetchPhaseLostConnectionReconnectsBeforeRethrowing`** (new, key regression) — a lost connection during `fetchAll()` invokes `reconnect()` exactly once and does **not** re-prepare, then the original exception propagates. Fails without this change (reconnect called 0×).
- **`testExecuteInsideTransactionReconnectsThenRethrows`** (updated) — `execute()` losing the connection inside a transaction now drops the socket before rethrowing. Fails without this change.
- **`testDoesNotReconnectForNonExecuteNonConnectionError`** (new) — a non-connection error during `fetchAll()` must **not** reconnect (no pool thrashing).
- The `execute()`-outside-transaction reprepare/replay tests are unchanged and still pass.

```
OK (9 tests, 35 assertions)   # with the fix
Tests: 3, Assertions: 10, Failures: 2   # the two reconnect-asserting tests fail without the fix
```

`vendor/bin/pint --test` passes on the changed files. (The local full-suite run hits a pre-existing, unrelated `Utopia\Cache\Feature\Leasable` fatal in `WithCacheLeaseTest.php` from a stale local `utopia-php/cache` install — present on clean `main`; CI's fresh install is unaffected.)

## Relationship to other work

- **Complements** utopia-php/database#898 (CLOUD-3N2A): that hardens the *stale-transaction rollback* side in `startTransaction()`; this hardens the *poisoned-socket* side so the connection is never reclaimed desynced. Different files, no overlap.
- **Defense in depth**: utopia-php/pools#33 destroys/recreates unhealthy connections on reclaim at the pool layer. This DB-layer fix makes the reclaimed PDO connection fresh regardless.
- **Downstream containment**: appwrite-labs/cloud#4827.

Refs: utopia-php/database#895, utopia-php/database#896, appwrite-labs/cloud#4827.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of lost database connections during statement execution and result fetching.
  * Automatically retries eligible executions when no transaction is active.
  * Reconnects before reporting unrecoverable connection errors, helping prevent unstable connections from being reused.
  * Preserves immediate error reporting for non-connection-related failures.

* **Tests**
  * Expanded coverage for connection loss during transactions, fetching, and statement execution.

### DIFF
--- a/src/Database/PDOStatement.php
+++ b/src/Database/PDOStatement.php
@@ -10,11 +10,19 @@ use Utopia\Console;
  * against the fresh connection, previously bound parameters/columns/attributes
  * are replayed, and the failed execute() is retried.
  *
- * Recovery is attempted only for execute(), and only outside a transaction:
- * re-running any other method (fetch, rowCount, ...) without a fresh execute
- * would return data from an unexecuted statement, and a connection cannot be
- * healed in place mid-transaction (the uncommitted state is gone, so the call
- * is rethrown for Adapter::withTransaction to roll back and replay).
+ * Transparent retry is attempted only for execute(), and only outside a
+ * transaction: re-running any other method (fetch, rowCount, ...) without a
+ * fresh execute would return data from an unexecuted statement, and a
+ * connection cannot be healed in place mid-transaction (the uncommitted state
+ * is gone, so the call is rethrown for Adapter::withTransaction to roll back
+ * and replay).
+ *
+ * In every lost-connection case that cannot be retried, the poisoned connection
+ * is still discarded (reconnect) before the rethrow: the failed fetch/execute
+ * leaves the socket wire-desynced, and a pooled connection reclaimed with a
+ * partial result frame still buffered would bleed that frame into the next
+ * borrower's query. Reconnecting guarantees the connection returned to the pool
+ * is always clean.
  *
  * @mixin \PDOStatement
  * @implements \IteratorAggregate<int, mixed>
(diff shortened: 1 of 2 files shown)


abnegate:
Closing per course-correction. An adapter-level reconnect-on-fetch-error heals one *trigger* of a poisoned pooled connection, not the root cause. DAT-1904's root cause is at the connection-pool / lifecycle layer: a pooled connection can be handed to the next borrower carrying leftover state (an unconsumed result frame, or a leaked transaction counter). The correct invariant — a connection that experienced any error or still carries pending state is never reused; it is reset or destroyed before the next checkout — must be enforced there, independent of which trigger tainted it. Superseded by the pool/connection-lifecycle fix (see utopia-php/pools#33 lineage, utopia-php/database#895/#896).

abnegate:
Confirmed redundant after tracing the runtime path end-to-end — recording the evidence:

The production DB adapter `Utopia\Database\Adapter\Pool` routes **every** operation through `$this->pool->use(fn(Adapter $adapter) => $adapter->{$method}(...))`, and `withTransaction()` wraps the whole transaction in a single `use()`. When `fetchAll()` (or a mid-transaction write) raises a lost-connection error, it propagates out of the `use()` callback, so `Utopia\Pools\Pool::use()` sets `$failed = true` and calls `release($connection, failed: true)`. That path (pools#33) runs `recover()`, which finds `reconnect()` on the pooled `Adapter` and calls it — `SQL::reconnect()` opens a fresh PDO (discarding the wire-desynced socket) **and** resets `inTransaction = 0` (the leaked-counter vector), then reclaims the clean connection; if `reconnect()` throws it `destroy()`s instead. Either way the poisoned connection is never handed to the next borrower.

So the pool already reconnects the exact socket this PR reconnects, on the same error, one layer down — for both the result-frame and transaction-counter taint, and for every borrow-error path (not just fetch). An adapter-level reconnect here is redundant. Root fix: utopia-php/pools#33.

PR #934, opened 2026-08-11
fix(mongo): decode BSON Int64 objects in adapter read paths

bhardwajparth51:
## What does this PR do?
This PR fixes an issue where reading array-typed integer or bigint attributes stored in MongoDB causes a fatal TypeError. The MongoDB driver deserializes 64-bit integers as `MongoDB\BSON\Int64` objects inside PHP arrays. This change adds BSON `Int64` conversion to `convertStdClassToArray()`, applies normalization across `getDocument()`, `createDocument()`, and `createDocuments()`, and hardens `castingAfter()` for integer types. It also adds an `ArrayAccess` check in `Document::__construct()` while keeping the `!$child instanceof self` guard intact.

Related Issue: [#13175](https://github.com/appwrite/appwrite/issues/13175)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of nested document data when retrieving or creating records.
  * BSON values are now converted consistently into standard PHP arrays and integers.
  * Preserved non-document objects embedded within attribute arrays without altering their identity.

### DIFF
--- a/src/Database/Adapter/Mongo.php
+++ b/src/Database/Adapter/Mongo.php
@@ -1267,7 +1267,7 @@ class Mongo extends Adapter
 
         $resultArray = $this->client->toArray($result[0]);
         $result = $this->replaceChars('_', '$', $resultArray);
-        $document = new Document($result);
+        $document = new Document($this->convertStdClassToArray($result));
         $document = $this->castingAfter($collection, $document);
 
         // Ensure missing relationship attributes are set to null (MongoDB doesn't store null fields)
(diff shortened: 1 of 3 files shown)


bhardwajparth51:
@fogelito can u review

ArnabChatterjee20k:
@bhardwajparth51 can you please address the suggestions in the codeql test

abnegate:
Thanks for digging into this, the diagnosis is right.

We're fixing it a layer down instead, in utopia-php/mongo#49, so the client stops handing out `Int64` at the decode boundary.

The deciding case was `sum()`. It returns the aggregate total straight off the cursor with a `float|int` return type, so any total over int32 throws:

```
TypeError: Utopia\Database\Adapter\Mongo::sum(): Return value must be of type int|float, MongoDB\BSON\Int64 returned
```

Two rows of `2000000000` is enough. Normalising per call site misses that one, and leaves the next read path we add exposed too.

Worth knowing for next time:

- This branched off before #935, so `convertStdClassToArray` here predates the empty object guard. A rebase needs care or `{}` fidelity regresses.
- The `ArrayAccess` hydration in `Document::__construct` isn't related to this bug. Calling `iterator_to_array`/`getArrayCopy`/`toArray` on arbitrary objects adds a fatal surface we have no case for.

Closing in favour of utopia-php/mongo#49.

PR #945, opened 2026-08-25
fix: hold pooled timeouts on the pool, not on a connection

abnegate:
## What

`Pool::setTimeout()` delegated through `$pool->use()`, checking a connection out purely to carry the value.

No adapter needs a connection for this. `MariaDB`, `MySQL`, `Postgres` and `Mongo` all just record the timeout and apply it to the SQL they build afterwards — none contacts the server. The checkout existed only to write a number onto whichever connection answered, which the pool took back moments later.

## Why it matters

Two defects fell out of that:

1. **The timeout bound one connection and none of its siblings.** `Pool::setTimeout()` never recorded the value on the pool, so `getTimeout()` stayed `0` and the `if ($this->getTimeout() > 0)` replay in the checkout path never fired. A handle configured for a 300s timeout ran unbounded on every connection except the one that happened to answer the setter.

2. **Building a handle dialled the database.** A factory that only constructs a `Database` — sets database, namespace, timeout — failed outright while the backing was unreachable, reporting a refusal to a caller that had not yet issued a query. In Appwrite Cloud this is `CLOUD-3PKZ`: the billing project-aggregation worker raised 196,554 "database refused the request" errors in 25 days from `Factory->logs()`, which does nothing but build a handle. The same worker's genuine query-time refusals over the same window number 38.

## Change

The pool holds the timeout per event and applies it to each connection as it is checked out, replacing whatever the previous holder left there — connections outlive the handle that configured them and are shared across handles that want a different timeout or none. A cleared event is kept at `0` rather than dropped so the clear reaches that connection too.

`setTimeout()` keeps the `<= 0` rejection the concrete adapters raise, so a caller mistake stays an error instead of becoming a silently absent timeout.

## Tests

`tests/unit/PoolTimeoutTest.php`. Verified red before the fix — 3 failures + 1 error out of 5:

- `testSetTimeoutDoesNotOpenAConnection` — pool factory throws `Connection refused`; the setter must not reach it
- `testTimeoutIsAppliedToEveryCheckout` — a connection whose timeout was cleared by another holder gets it back on the next checkout
- `testClearedTimeoutIsClearedOnEveryCheckout`
- `testNonPositiveTimeoutIsRejected`
- `testPerEventTimeoutReachesTheConnection` — passes on both shapes; guards the fix against silently dropping timeouts

`composer check` (PHPStan level 7) and `composer lint` clean; unit suite 456 tests green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pool timeout settings are now retained consistently across connection checkouts.
  * Cleared timeouts remain cleared when connections are reused.
  * Per-event timeout settings are applied reliably to each checked-out connection.
  * Timeout configuration no longer requires opening a database connection immediately.
  * Non-positive timeout values are rejected with a clear database error.

### DIFF
--- a/src/Database/Adapter.php
+++ b/src/Database/Adapter.php
@@ -357,6 +357,24 @@ abstract class Adapter
     }
 
     /**
+     * Clears every timeout this adapter carries, for any event.
+     *
+     * A pooled connection outlives the handle that configured it, so the handle
+     * that takes it next has to be able to reset it without knowing which
+     * events the previous one set a timeout for.
+     *
+     * @return void
+     */
+    public function clearTimeouts(): void
+    {
+        foreach (\array_keys($this->transformations) as $event) {
+            $this->clearTimeout($event);
+        }
+
+        $this->timeout = 0;
+    }
+
+    /**
      * Start a new transaction.
      *
      * If a transaction is already active, this will only increment the transaction count and return true.
--- a/src/Database/Adapter/Pool.php
+++ b/src/Database/Adapter/Pool.php
@@ -34,9 +34,6 @@ class Pool extends Adapter
      * failed outright while the backing was unreachable, reporting a database
      * as down to a caller that had not yet issued a query.
      *
-     * A cleared event is kept with a timeout of zero instead of being dropped,
-     * so the connection it was applied to has it cleared again on checkout.
-     *
      * @var array<string, int>
      */
     private array $timeouts = [];
@@ -116,37 +113,56 @@ class Pool extends Adapter
 
     /**
      * Zero is the value a caller's own default carries when it wants no
-     * timeout, so it is recorded as a cleared event rather than refused. A
-     * connection is only ever asked for a timeout it can hold.
+     * timeout, so it clears the event rather than being refused. A connection
+     * is only ever asked for a timeout it can hold.
      */
     public function setTimeout(int $milliseconds, string $event = Database::EVENT_ALL): void
     {
-        $this->timeouts[$event] = \max($milliseconds, 0);
-        $this->timeout = \max($milliseconds, 0);
-    }
+        if ($milliseconds <= 0) {
+            $this->clearTimeout($event);
 
-    public function clearTimeout(string $event): void
-    {
-        $this->timeouts[$event] = 0;
-        $this->timeout = 0;
+            return;
+        }
+
+        $this->timeouts[$event] = $milliseconds;
+        $this->timeout = $this->timeouts[Database::EVENT_ALL] ?? 0;
     }
 
     /**
-     * Apply this pool's timeouts to a connection as it is checked out. The
-     * connection outlives the handle that configured it and is handed to
-     * handles that want a different timeout or none, so what it carries is
-     * replaced on every checkout rather than only when it is first set.
+     * Clearing one event leaves the others alone. The concrete adapters keep a
+     * single timeout scalar that Postgres and Mongo apply to every statement,
+     * so a clear forwarded verbatim would drop the timeout the caller still
+     * has configured for everything else.
+     */
+    public function clearTimeout(string $event): void
+    {
+        unset($this->timeouts[$event]);
+        $this->timeout = $this->timeouts[Database::EVENT_ALL] ?? 0;
+    }
+
+    /**
+     * Put a connection into the timeout state this pool holds, as it is checked
+     * out. The connection outlives the handle that configured it and is handed
+     * on to handles that want a different timeout or none at all, so it is
+     * reset first: a handle carrying no timeout must not inherit one, and a
+     * handle carrying its own must not be left with an event the last holder
+     * set. The global timeout is applied last so the scalar the connection ends
+     * on is the one {@see self::getTimeout()} reports.
      */
     protected function syncTimeouts(Adapter $adapter): void
     {
-        foreach ($this->timeouts as $event => $milliseconds) {
-            if ($milliseconds === 0) {
-                $adapter->clearTimeout($event);
+        $adapter->clearTimeouts();
 
+        foreach ($this->timeouts as $event => $milliseconds) {
+            if ($event === Database::EVENT_ALL) {
                 continue;
             }
 
             $adapter->setTimeout($milliseconds, $event);
+        }
+
+        if (isset($this->timeouts[Database::EVENT_ALL])) {
+            $adapter->setTimeout($this->timeouts[Database::EVENT_ALL]);
         }
     }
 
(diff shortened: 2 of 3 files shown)


abnegate:
Amended: `setTimeout(0)` is now held as "no timeout" rather than refused.

Callers pass zero from their own defaults — `Appwrite\Cloud\Database\Factory::logs()`/`tenant()` both default `$timeout` to `0` meaning none — and under the old delegation that landed on an adapter which either ignored it (no timeout support) or threw. Rejecting it in the pool turned working calls into errors, which four cloud `FactoryTest` cases caught. Zero is now recorded as a cleared event, so it reaches a connection as `clearTimeout()` and never as a value the connection would refuse. Covered by `testZeroIsHeldAsNoTimeout`.

abnegate:
Both Greptile findings were real and are fixed in 2a985011. I confirmed each against the concrete adapters before changing anything.

**Event clear disables global timeout** — confirmed. `Adapter::clearTimeout()` zeroes the single `$this->timeout` scalar whatever event it is handed, and that scalar is what `Postgres::getTimeoutStatement()` writes into `SET statement_timeout` (`Postgres.php:71`) and what Mongo writes into `maxTimeMS` (`Mongo.php:2524`). Replaying a per-event clear verbatim therefore left the whole handle unbounded. The pool now drops the event from its own set rather than forwarding the clear, and applies the global timeout last so the scalar a connection ends on is the one `getTimeout()` reports.

**Empty timeout map performs no reset on checkout** — also confirmed, and worse than the summary suggests: it is not only a cross-handle leak, it is directional. A migration handle built with no timeout (`Appwrite\Cloud\Platform\Migration` calls `clearTimeout()` before a long run) inherited a request handle's 5s bound off the shared pool and would be killed mid-run. `Adapter::clearTimeouts()` now resets every event a connection carries — iterating its own transformation keys, since the next holder cannot know which events the last one used — and `syncTimeouts()` leads with it.

Both have regression tests, both seen red on `aa26f15e`:

- `testClearingOneEventLeavesTheGlobalTimeoutInPlace` — `Failed asserting that 0 is identical to 300000`
- `testHandleWithNoTimeoutResetsTheConnectionItIsGiven` — `Failed asserting that two arrays are identical`

Unit suite 458 green, `composer lint` and `composer check` (PHPStan level 7) clean. Also ran the downstream consumer against the patched adapter: appwrite/cloud's full unit suite, 3125 tests, green.

Points worth a look:

1. `Adapter::clearTimeouts()` is new public API on the base adapter. It iterates `$this->transformations` keys, which is the only record of which events carry a hook.
2. Ordering in `syncTimeouts()` is load-bearing — per-event first, `EVENT_ALL` last — because the concrete adapters share one timeout scalar across events.
3. A per-event-only timeout (no `EVENT_ALL`) still leaves that value in the shared scalar, so on Postgres/Mongo it bounds everything. That matches what an unpooled adapter does with the same calls, so I left it rather than diverging pooled from direct behaviour. No caller in cloud or CE sets a per-event timeout.

abnegate:
@coderabbitai's finding was real and is fixed in bdc0ae1b — thanks, this was a regression I introduced.

Under the old code the setter delegated, and `delegate()` routes to `pinnedAdapter` when one is held, so a timeout changed inside `withTransaction()` reached the connection running it. Holding the value on the pool alone silently dropped that: a transaction pins one connection for its whole body and does not check out again before the commit, so the rest of the body kept running under the timeout the caller had just replaced.

I took the suggested shape with one change. Rather than reading `$this->pinnedAdapter` directly, the lookup goes through a `pin()` seam. appwrite/cloud subclasses this adapter and keys the pin by coroutine — a task resolves one handle for the whole process and hands it to every coroutine it spawns, so a pin on the object is read by whichever coroutine queries next and one coroutine's `find` executes on another's open transaction (MySQL SQLSTATE 2014). That subclass sets nothing on the object, so a direct read would have compiled, passed, and still missed every update in the consumer this PR exists for. It overrides `pin()` in one line and is covered on its own side.

`testTimeoutChangedInsideATransactionReachesThePinnedConnection`, seen red on 2a985011: `Failed asserting that two arrays are identical`. The cloud-side equivalent was also seen red: `Failed asserting that 5000 is identical to 300000`.

Unit suite 459 green, lint and PHPStan level 7 clean; cloud's full unit suite 3126 green against the patched adapter.

abnegate:
Round 2. One of the two findings was real and is fixed in a782c70d; the other I am refuting, with reasoning.

**Clear-all retains event timeouts — real, fixed.** `Pool` inherited `clearTimeouts()` from `Adapter`, which walks the events it finds in `$transformations`. This adapter delegates `before()`, so its own array never holds more than `EVENT_ALL` no matter how many events a caller has set a timeout for. The per-event entry survived in `$this->timeouts` — the map a checkout replays — and came back on the next checkout after the caller had cleared it. `Pool` now overrides `clearTimeouts()` to empty its own map and sync the pin.

`testClearingEveryTimeoutDropsPerEventEntriesToo`, seen red on bdc0ae1b:

```
--- Expected
+++ Actual
-Array &0 ()
+Array &0 (
+    'document_read' => 5000
+)
```

**Global timeout overrides event timeout — I do not think this should change.** The finding is an accurate description of the behaviour, but the requested ordering makes things worse on the two engines it names.

`Postgres::setTimeout()` (`Postgres.php:96-106`) and `Mongo::setTimeout()` (`Mongo.php:101-108`) both accept `$event` and discard it — neither registers a hook, both assign the single `$this->timeout` scalar. Per-event timeouts have never been expressible on those engines, pooled or not. So on a **direct, unpooled** Postgres adapter, `setTimeout(300000)` followed by `setTimeout(5000, EVENT_DOCUMENT_READ)` leaves the scalar at 5000 and bounds *every* statement at 5s, writes included. Replaying in that order is what "event-specific precedence" would mean here.

Applying the global last instead means a per-event refinement those engines cannot express is ignored on them. MariaDB and MySQL are unaffected either way — they hang the hook on the event, and it works. Between silently ignoring a refinement and silently applying a 5s read deadline to every write on the handle, I would rather ignore the refinement, so I have kept the ordering and documented the reasoning at `syncTimeouts()` rather than leaving the next reader to rediscover it.

The real defect is in the concrete adapters: `Postgres` and `Mongo` accept an `$event` argument they cannot honour and say nothing. Worth fixing, but not in a PR about pool checkout — happy to raise it separately.

Unit suite 460 green, lint and PHPStan level 7 clean. Cloud's full unit suite 3126 green against the patched adapter. Previous head bdc0ae1b was fully green on CI, including the whole adapter matrix (MySQL, Postgres, MongoDB, SQLite, Redis, Memory, Mirror, Pool and the SharedTables/Schemaless variants).

PR #946, opened 2026-08-26
feat: drop unknown attributes on schema-managed writes instead of failing

abnegate:
## What

Adds `Database::setDropUnknownAttributes(bool)`. When enabled, an attribute the collection schema does not declare is removed from the document before the write and logged as a warning, instead of failing the write with `Invalid document structure: Unknown attribute: "x"`.

Off by default, so nothing changes for any existing caller.

## Why

A deploy that starts writing an attribute before the migration that creates its column has run makes **every** write to that collection throw. On Appwrite Cloud this took OAuth2 login down for every provider: the `identities` row carried `photoUrl`, the deployed schema did not, and `/v1/account/sessions/oauth2/:provider/redirect` 500'd on each login until the feature was reverted (appwrite/appwrite#13326, reverted by appwrite/appwrite#13355).

The rejection is right when the caller owns the schema, and wrong when the application owns it. A column that has not been created yet is a deploy-ordering fact, not bad input, and refusing the write turns it into an outage. This lets the schema owner opt into the degradation: the request completes with the columns that exist, the missing value is lost until the migration runs and the writer sends it again, and the warning names what was dropped.

It stays off by default because for a caller writing into their own schema, silently discarding data they sent is worse than refusing it.

## Scope

The drop removes **exactly** the set `Structure` would have rejected: keys that are neither `$`-prefixed nor declared in `collection.attributes`. So any write that validates today is unchanged, and the only writes whose behaviour changes are the ones that throw today.

Adapters without attribute support are skipped entirely, so schemaless collections are untouched.

Applied on all five write paths that validate structure: `createDocument`, `createDocuments`, `updateDocument`, `updateDocuments`, `upsertDocumentsWithIncrease`.

On `updateDocument` and `upsertDocumentsWithIncrease` the removal runs **before** the change-detection diff, not after `encode()`. A dropped attribute is never in the stored document, so diffing against it first made a write carrying one look like a real change: `$updatedAt` bumped and `EVENT_DOCUMENT_UPDATE` fired on an identical row. That is the exact shape of the deploy window this flag is for, so it would have meant a spurious update event per request to every webhook, function and realtime subscriber on the collection.

`Mirror` forwards every write to its source and destination, so it delegates the setter the same way it already delegates `preserveDates` and `preserveSequence`. Without that the flag decided nothing on a mirrored deployment, which the new test caught under the Mirror adapter.

## Tests

`testDropUnknownAttributes` in `DocumentTests` drives the public entry points and asserts what the database is holding afterwards, not just that no exception was raised:

- with the flag off, `createDocument` still throws `Invalid document structure: Unknown attribute: "unknown"`
- with the flag on, create and update both succeed, the declared attribute persists, and the undeclared one is absent from a cache-purged read on both paths
- a further update whose only new value is the dropped attribute leaves `$updatedAt` untouched

Both halves were seen red before being fixed:

- with `removeUnknownAttributes` reverted to a pass-through, the test fails with `Utopia\Database\Exception\Structure: Invalid document structure: Unknown attribute: "unknown"`, the production error verbatim
- with the removal moved back after `encode()`, the `$updatedAt` assertion fails on a 10ms difference

The Mirror delegation was found by the test, not by inspection: it failed under `MirrorTest` before the override existed.

## Follow-up

Consumers opt in separately: appwrite/appwrite#13357 enables it on the platform, project and logs handles and leaves tenant handles strict, and appwrite-labs/cloud#5463 does the same for cloud's own factory.

## Summary by CodeRabbit

- **New Features**
  - Added support for removing attributes that are not defined in the document schema when configured.
  - Added configuration support for applying this behavior consistently across mirrored databases.

- **Bug Fixes**
  - Unknown attributes are now removed during encoding and persistence.
  - Updates containing only unknown attributes no longer register as document changes.

### DIFF
--- a/src/Database/Database.php
+++ b/src/Database/Database.php
@@ -5771,7 +5771,6 @@ class Database
         }
 
         $document = $this->encode($collection, $document);
-        $document = $this->removeUnknownAttributes($collection, $document);
 
         if ($this->validate) {
             $validator = new Permissions();
@@ -5893,7 +5892,6 @@ class Database
             }
 
             $document = $this->encode($collection, $document);
-            $document = $this->removeUnknownAttributes($collection, $document);
 
             if ($this->validate) {
                 $validator = new Structure(
@@ -6661,7 +6659,6 @@ class Database
             $updates,
             applyDefaults: false
         );
-        $updates = $this->removeUnknownAttributes($collection, $updates);
 
         if ($this->validate) {
             $validator = new PartialStructure(
@@ -9263,22 +9260,28 @@ class Database
     }
 
     /**
-     * Remove attributes the collection schema does not declare
+     * Remove attributes the collection schema does not declare.
+     *
+     * Used ahead of change detection on update/upsert so a dropped key is not
+     * counted as a write. Encode also calls this after iterating attributes.
      *
      * @param Document $collection
      * @param Document $document
+     * @param array<string, true>|null $known Attribute ids already collected (e.g. during encode)
      *
      * @return Document
      */
-    protected function removeUnknownAttributes(Document $collection, Document $document): Document
+    protected function removeUnknownAttributes(Document $collection, Document $document, ?array $known = null): Document
     {
         if (!$this->dropUnknownAttributes || !$this->adapter->getSupportForAttributes()) {
             return $document;
         }
 
-        $known = [];
-        foreach ($collection->getAttribute('attributes', []) as $attribute) {
-            $known[$attribute['$id'] ?? ''] = true;
+        if ($known === null) {
+            $known = [];
+            foreach ($collection->getAttribute('attributes', []) as $attribute) {
+                $known[$attribute['$id'] ?? ''] = true;
+            }
         }
 
         $dropped = [];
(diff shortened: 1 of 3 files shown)


abnegate:
Addressed the change-detection ordering: `removeUnknownAttributes()` now runs before the `$shouldUpdate` / `$hasChanges` diff in `updateDocument` and `upsertDocumentsWithIncrease`, rather than after `encode()`.

Verified the defect was real before fixing it: with the drop running after the diff, a write whose only new value is the dropped attribute bumped `$updatedAt` by 10ms and would have fired `EVENT_DOCUMENT_UPDATE` on an unchanged row. That is the exact shape of the deploy window this flag exists for, so a lagging column would have meant a spurious update event per request to every webhook, function and realtime subscriber on the collection.

`testDropUnknownAttributes` now asserts `$updatedAt` is unchanged across that write. Seen red against the previous ordering.

src/Database/Database.php
```diff
@@ -5750,6 +5771,7 @@ public function createDocument(string $collection, Document $document): Document
         }
 
         $document = $this->encode($collection, $document);
+        $document = $this->removeUnknownAttributes($collection, $document);
```
abnegate: @cursoragent Just drop them in encode while we're already iterating attributes

PR #975, opened 2026-09-17
fix: clarify invalid reverse one-way relationship values

HarshMN2345:
Creating a document with an array on the child side of a one-way one-to-one relationship failed with `Must be either a document ID or a document, array given.`, although a document or an ID is rejected from that side too. The array branch of `createDocumentRelationships()` now reports the direction error first, as its document and ID branches and the update path already do.

`testCreateInvalidOneWayChildArrayValueRelationship` sits with the other `testCreateInvalid*ValueRelationship` tests and uses the object payload from the issue. It fails on `main` with the old message.

Related to appwrite/appwrite#8345. Appwrite gets the fix once it requires a release that contains this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - One-to-one relationships configured as one-way now consistently reject values assigned from the child side.
  - Invalid child documents are not created, and the parent relationship remains unchanged.
  - Valid creation from the parent side continues to work as expected.


tests/e2e/Adapter/Scopes/Relationships/OneToOneTests.php
```diff
@@ -17,6 +18,48 @@
...
+        $database->createDocument($parents, new Document(['$id' => 'parent']));
+
+        try {
+            foreach ([['name' => 'parent'], ['parent'], [], 'parent', new Document(['$id' => 'parent'])] as $value) {
+                $id = ID::unique();
+                try {
+                    $database->createDocument($children, new Document(['$id' => $id, 'parent' => $value]));
+                    $this->fail('Expected the one-way reverse relationship to be rejected.');
+                } catch (RelationshipException $error) {
+                    $this->assertSame('Invalid relationship value. Cannot set a value from the child side of a oneToOne relationship when twoWay is false.', $error->getMessage());
```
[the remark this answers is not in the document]
HarshMN2345: The message is what this PR changes. `RelationshipException` was already thrown before the fix, just with "array given", so without the message assertion the test passes on `main`. Appwrite returns this text verbatim as the 400 error, and appwrite/appwrite#8345 is about that wording. The test now follows `testCreateInvalid*ValueRelationship`, which pins messages with `expectExceptionMessage` the same way.

src/Database/Database.php
```diff
@@ -5998,6 +5998,10 @@ private function createDocumentRelationships(Document $collection, Document $doc
             try {
                 switch (\gettype($value)) {
                     case 'array':
```
fogelito, line 6000: Should this throw in case of array? Maybe should be Object (Document)?

HarshMN2345: Yeah, it should. The child side is always invalid for one-way relationships, so arrays should hit the same error too. We only see an array here because that’s what reaches this code path.

HarshMN2345: To be precise, this only applies to one-way one-to-one. The Document case already throws the same error. The reason its an array here is that the child key isnt treated as an attribute, so it never gets converted to a Document. 🤔

fogelito (approved):

PR #977, opened 2026-09-21
fix: query relations in ManyToOne deleteSetNull and guard empty collections

PINYOPATTANAWASANPORN:
## What

1. Remove the `if (!$twoWay)` check in `deleteSetNull()` for `RELATION_MANY_TO_ONE`, ensuring that related parent documents referencing the child document being deleted are always queried and their foreign keys set to `null`.
2. Add defensive `empty($value)` early-return checks in `deleteSetNull()` for both `RELATION_ONE_TO_MANY` and `RELATION_MANY_TO_ONE`.
3. Add a test in `ManyToOneTests::testManyToOneTwoWayRelationship` verifying that deleting a child document with `onDelete: setNull` and 0 related parent documents succeeds without error.

## Why

In `Database::deleteSetNull()`, `case Database::RELATION_MANY_TO_ONE:` previously had:
```php
if (!$twoWay) {
    $value = $this->find($relatedCollection->getId(), [ ... ]);
}
foreach ($value as $relation) { ... }
```
When deleting a child document in a two-way Many-to-One relationship (such as deleting a user when files has a `manyToOne` relationship pointing to user):
- Since `$twoWay` is `true`, `$this->find()` was skipped.
- `$document` fetched during deletion does not have virtual relationship collections eagerly loaded, leaving `$value` as `null`.
- When deleting a child document that has 0 related rows (or any two-way Many-to-One child document), `foreach ($value as $relation)` triggered a fatal error: `TypeError: foreach() argument must be of type array|object, null given`.
- This causes API calls (like `databases.deleteDocument` or `tablesDB.deleteRow`) to fail with HTTP 500 `general_unknown`.

In comparison, `deleteCascade()` and `deleteRestrict()` unconditionally query the referencing collection via `$this->find()` / `$this->findOne()` regardless of `$twoWay`.

## How

- Removed `if (!$twoWay)` so `$this->find()` is always executed on the child side of `RELATION_MANY_TO_ONE` to look up referencing parent documents.
- Added `if (empty($value)) { break; }` so that if no related documents exist, the function exits early without looping.
- Added identical `if (empty($value)) { break; }` guard for `RELATION_ONE_TO_MANY`.

## Test Plan

- Added test case in `ManyToOneTests::testManyToOneTwoWayRelationship`:
  - Created a child document (`store8`) with 0 related parent products.
  - Executed `$database->deleteDocument('store', 'store8')` under `onDelete: setNull`.
  - Asserted `$deleted === true` and `$database->getDocument('store', 'store8')->isEmpty() === true`.

Related to appwrite/appwrite#13766.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deletion behavior for documents involved in one-to-many and many-to-one relationships.
  * Deletions now complete successfully when no related records exist.
  * Related references are cleared consistently for two-way relationships when a document is deleted.
  * Prevented unnecessary processing when relationship results are empty, improving reliability during set-null deletions.


PINYOPATTANAWASANPORN:
Closing as this was incorporated into main via merged PR #979. Thank you @HarshMN2345!

